PR #28, opened 2019-04-22
Use Promise.all for simultaneously load project data

ironsoul0:
This commit could be considered as a possible optimization for loading project data components.


tommai78101:
The change will make it so it'll be nearly identical in performance to chaining a `.then()` off the `Promise.all()`. Did you do any benchmark?

Issue #19, opened 2019-04-21
[Request] Create in the menu dropdown HELP>ABOUT section


JasonB32:
It would be great to be uniform with most software by having a HELP>ABOUT window dropdown so users would know what version of GB Studio their working with.

chrismaltby:
Good point! This is implemented now in the develop branch 94560f6d373b643a8074d0557a1f1ab3a9609ef4

JasonB32:
Thank you!

PR #32, opened 2019-04-23
#19 Adding a Help -> About message box

mike-franz:
#19 Adding a Help -> About message box so users can know what version they are using. Last tag in repo was 1.0.0.0 so I used that as the current version.

### DIFF
--- a/src/consts.js
+++ b/src/consts.js
@@ -1,5 +1,7 @@
 import path from "path";
 
+const gbStudioVersion = '1.0.0.0';
+
 const isDist =
   __dirname.endsWith("/dist") ||
   __dirname.endsWith("/dist/windows") ||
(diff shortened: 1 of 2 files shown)


chrismaltby:
Thanks for doing this but I'd actually been working on this already and now have a solution in 94560f6d373b643a8074d0557a1f1ab3a9609ef4.

Here's some constructive criticism of your pull request though if you don't mind:

- You added gbStudioVersion to the consts.js file, but actually you could have just imported the version directly from the package.json file instead that way it would always be up to date.
- Where you added the menu would have been good for Windows and Linux, but on a Mac the convention is to add the "About" menu item under the app name menu (first one in the list), actually the Mac version already had an "About" window so this commit would have added it twice (but slightly differently each time) on a Mac.
- Inconsistent quotes, you've got some single and some double, I'm using https://prettier.io/ as a code formatter so they're all doubles in mine (though I haven't documented this anywhere yet)
- Using a var to store a const object on the options
- I'm not a fan of the "OK" button, I prefer to use verbs on button labels so it's clear what's going to happen when you click it (See http://usabilitypost.com/2008/08/30/usability-tip-use-verbs-as-labels-on-buttons/) Although for this case I don't think it needs a button anyway.

I'm not trying to be picky though!

PR #52, opened 2019-04-24
add css theme selection

### LINKED ISSUE #16: [Request] Dark Theme for Windows
That would be awesome. I seen someone paste a photo of their Mac version and it looked like they already had a dark theme on it, please don't forget us Windows users :D

pay64k:
This closes #16.

This introduces platform independent theme selection. Feedback will be much appreciated :)

### DIFF
--- a/src/components/library/Button.css
+++ b/src/components/library/Button.css
@@ -45,10 +45,6 @@
   fill: var(--button-light-icon-color);
 }
 
-.Button:hover {
-  background-color: var(--toolbar-button-active-bottom-color);
-}
-
 .ButtonToolbar {
   display: flex;
   flex-direction: row;
@@ -86,17 +82,6 @@
   transform: rotate(180deg);
   width: 8px;
   height: 8px;
-}
-
-.Platform__win32 .Button {
-  box-shadow: none;
-  border-radius: 2px;
-  background: #fff;
-  border-color: #ddd;
-}
-
-.Platform__win32 .Button:hover {
-  background: #fafafa;
 }
 
 .Platform__win32 .Button--Transparent {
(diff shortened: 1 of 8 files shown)


chrismaltby:
Nice one, this is close to what I was going to implement. There's a few things though:

- I've been looking at other applications that do this (e.g. Sourcetree) and I'd like to match them by having this under the View menu rather than Window (I think below "Build and Run" and above "Show Collisions" but with a separator on either side .
- This bit https://github.com/chrismaltby/gb-studio/blob/dab277f2e4a8bf4003add8aae3ab8bb7ae369dc3/src/windows/projectRoot.js#L70-L80 is currently duplicating the functionality of this bit https://github.com/chrismaltby/gb-studio/blob/dab277f2e4a8bf4003add8aae3ab8bb7ae369dc3/src/windows/projectRoot.js#L130-L134
(oops I should remove that console.log, how embarrassing). You should just be able to call `updateMyAppTheme(theme==="dark")` in that ipcRenderer.on block.
- Your implementation doesn't currently store the user's preference, so if you open a new window you'll be back in the light mode again, there's no user settings currently in the app though. Show collisions and Show connections are the closest there is and they're stored in the project itself right now, think this should be application wide not project wide so I'm thinking we should add a module like https://github.com/nathanbuchar/electron-settings I can't vouch for this one specifically, maybe need to do some research on the best method.
- It would be nice if it used the checkbox menu to highlight the currently selected option like this
Screenshot 2019-04-24 at 22 11 28

Hope there's nothing too crazy in there, user preferences is something I'd been looking at too. You might want to hold off until I've found a method I want to go with because I'll probably be really picky about how that's set up given that it will effect how preferences for potentially lots of other things (language, default paths) get stored in the future. But if you have any good ideas on that I'm open to them.

pay64k:
Good points. It was supposed to be a quick think so my eyes stop hurting from light theme. With those requirements and insights above, it should be more clear what the end goal is. I am closing this for now and try to come back with better way :)

Issue #59, opened 2019-04-25
[Request] Assign random number to variable


Shokupanda:
Hey, I hope it's okay to have another request for a future version of GB Studio.

Would it be possible to assign random numbers to a variable? Like "Set variable to random number between x and y". That'd be great!

Cheers!

thomas-alrek:
@chrismaltby I think this issue can be closed

PR #61, opened 2019-04-25
Implement event for setting variable random value

thomas-alrek:
This feature implements a event for setting a variable to a random value between 0 and maxValue

Fixes issue #59


chrismaltby:
Hi Thomas, looks good. Just wondering which version you're running this on Windows, Mac or Linux? I've only had time to test on Mac just now (it works great!) and want to test to confirm its working on the others.

Windows is normally the problem as it's running a slightly older version of GBDK that tends to be a little buggy. I'm sure by now you'll have seen some of the weird bits of code that seemingly would do nothing but if altered will break things. I also had an issue the other day where (A==B) was not equal to (B==A) only on Windows...

I don't mind testing Windows and Linux (won't be until tomorrow now) but just want to check what you're running.

Thanks

thomas-alrek:
Hi! I work on a Mac as well, so I have only personally tested it on macOS 10.14.4

chrismaltby:
Hi Thomas, I tested on Windows and it (as is normally expected) didn't work, I checked out your branch and worked out a fix though so I've merged this and fixed the issue on Windows with 73401495a9a8016961dca1db9f86fd8905352c9e

thomas-alrek:
Great. It's weird that the windows version of GBDK acts so strange. Would it be possible to cross compile for Windows on macOS/Linux instead?

PR #64, opened 2019-04-25
Feature/i10n Norwegian

thomas-alrek:


chrismaltby:
Hi Thomas, thanks for this. I'm not sure I can accept the change to l10n.js though as what you have there will prevent being able to do a British English override, or Portuguese vs Brazilian Portuguese etc. What value are you getting back from app.getLocale()? Can you change the pull request to name the json as what you get for that value and revert the l10n.js changes?

thomas-alrek:
My bad. I was having some issues on my local device. But it seems to be working now. I have reverted the changes to l10n.js

chrismaltby:
Thanks!

chrismaltby:
By the way, I added a few more keys to en.json as part of e1b22b4c84932dfcf3b1c1d09d566890c97fa98e but you might want to hold off on doing those as I still need to go through the rest of the text and localise it, probably better if you're doing it in one go :-)

PR #81, opened 2019-04-27
Improve Actor scripting support

thomas-alrek:
# Improve Actor scripting support

This pull request implements a couple of new events to allow for more dynamic Actor scripting.

## Compiler
Adds support for compiling a pair of variables (vector pair) in a script.

### Actor Get Position
This event stores the given actors current position into a pair of variables.

### Actor Set Position To Value
This event sets the given actors position from a pair of variables.
Works like **Actor Set Position**, but the X and Y coordinates can now be dynamic.

### Actor Move To Value
This event moves the given actors position to the coordinate defined in a pair of variables.
Works like **Actor Move To**, but the X and Y coordinates can now be dynamic.

### Actor Set Relative Position
This events sets the given actors position relative to its current position.

### Actor Move Relative
This events moves the given actor relative to its current position.


chrismaltby:
Awesome, looks great. Just need some time to test this but from a quick scan of the code looks good. 👍

thomas-alrek:
I think there might be a bug in the Move Relative and Set Position Relative functions.
Could you explain how the different coordinate systems work? While looking at the earlier code, it seems to me like you have one coordinate for "world space", and one for the actual screen space?

RichardULZ:
Made a project with a numbered grid and text boxes with printed variables for testing the Relative position functions.
Prints last pos, the used movement vector, and the new position.
Seems to always do an extra 1 on any movement, +3 becomes +4, -2 becomes -3.
Y will use it's +- polarity, but always take X's value. eg:
X+1, Y-5, will produce a move of X+2, Y-2, (therefore X+0, Y+1 does nothing).
Also demonstrates going to the edge of screen reading actor position as 31, odd quirk.

testPos

Hope this helps debugging!
[Test_RelativePos.zip](https://github.com/chrismaltby/gb-studio/files/3131192/Test_RelativePos.zip)
Includes .gbsproject and .gb rom.

chrismaltby:
That’s really useful, thanks! I’ve started to build up a little folder of test projects like this in ‘test/projects’ will definitely add this one too.

PR #92, opened 2019-04-30
Patch ROM checksums

thomas-alrek:
Patches the ROMS checksums after modifying the headers.
If not done, the produced ROM will not boot on real hardware.


chrismaltby:
Ooops! Thanks!

PR #97, opened 2019-04-30
Fix Relative Actor positioning

thomas-alrek:
There was a bug in the implementation of the relative positioning.
the y axis was using the value for the x axis, and the offset was always too big.


chrismaltby:
Was going through this earlier, almost made the exact same commit :-D The +8 you had on the x and y values was an offset for the coordinate system and wasn't needed when moving relative.

PR #108, opened 2019-05-01
Improve Math argument labels and fix broken Event

thomas-alrek:
Fixes Copy Value event.
Improves argument labels of the Math events.


thomas-alrek:
Woops, closed this by accident

PR #117, opened 2019-05-04
Implement BG Script (Experimental) Run a script during player control

RichardULZ:
### A major feature that could run any script in the 'Background'
**Solution to eventually allow for use cases:**
+ Silently Wait for a button press like b, select, start, and do something or go to a menu scene when pressed. (Works great, but minor lag)
+ Have an actor follow the player while they walk. (Want actor collision off, like Hide actor)
+ A timer that counts down, like for a running race between actor and player. (can't count and run move to)

**The beginning, Investigation**
Originally formed from a crude "Allow Actor Movement" hack, which uncovers a bug btw, **"Push Actor"** will leave collisions on, even when running a new script. With this hack and bug, I could move around while text was playing, wait for button, everything. You could start to see **performance** drops when Writing text, playing music, and walking + scrolling with collision, all at the same time, but you can sometimes see that already.

This hack had issues, you could interact with the same actor while they were talking, which restarted their script from the beginning. And it could not return to a looped script. Yet...

After talking this through on discord, i was about to leave this as an issue, but then decided to bite the bullet and start coding it myself, I've never coded c before, i'm a visual script baby.

### The Solution
**A second script pointer!** The BG ptr. The main components would be:
- Internal variable for **BG_ptr, BG_ptr_bank, BG_start_ptr,** We store these on SetBGscript, and load them on script end.
- A **BGscript_Active** flag, To ignore "Is script ptr !=0" on collision and actor input.
- A script event, **Set BG Script** and **Clear BG Script**, copy's or reset's variables.
- Code in the **ScriptStart** event, to say BGscript not active, and _try_ and halt actor movement.
- Code in the **Script_End_b**, To check if there is a BGscript, and copy it over/ set active true. 
- Not sure when **if (!script_cmd_index)** (ScriptRunner.c line 126) is called true, but end check is also there, for now.

**Current issues**
+ **Actor moving one step every second on new script start,** This happens on 1.0 too. easier to pull off here, just run move actor to position as BG script, then talk to someone before they arrive. Possibly set actors[script_actor].moving = FALSE; on script start, but for all actors? because you never know which.
+ **Fixed!** bg actor move would start moving any other actor you talk to, **patched in ScriptStart**.
+ Issues with **IsLastFnComplete**, demo on UI text boxes auto completing, Try with looped Wait on a BG script.
+ General **performance**, even *Wait for Joypad* can cause *some* lag while collisions+scrolling+music. (Possible idea, option to halve the Script Update rate when BG script?)

I'm opening a pull so this is easy to find, and to get suggestions like on IsLastFnComplete.

**TL:DR**
Script that runs in Background, Enables Player movement and collision while on, Stops and restarts after other scripts.


thomas-alrek:
This is awesome. I am testing it now for handling:

```
Loop Forever
    Await Joypad Input
        Switch Scene To Menu
```

Seems to be working great. I think if we also adds a method to Get Current Scene, and Switch Scene from a Variable, we have a great solution for handling custom menus

RichardULZ:
Added a testing room with actors that initiate various BG script use cases.
A Start button return, A Start Button return with confirm, An actor that Follows you, A Cat that resets your BG script, A rock that waits (important for debug), Music player for speed testing, A dog that takes you to the outside world, A DUCK.
Outside is a person that want's to race, and default place to test pushing rocks.

thomas-alrek:
@RichardULZ @chrismaltby 

There was a lot of merge conflicts, so I have taken all the pending PRs and collected them into a new PR where all the new features are merged.

This PR should be directly mergable to develop.

#123

Zzausch:
Awesome, can't wait!

PR #123, opened 2019-05-05
Merge all new features

thomas-alrek:
This PR merges all the pending pull requests into one single PR.

* #104 Add support for notes field
* #114 Set project author from username
* #117 Implement BG Script (Experimental) Run a script during player control
* #119 Add support for pushing and popping scene state
* #120 Make Editor Sidebar resizeable
* #122 Feature/collisions disable
* Feature/Invoke Actor


thomas-alrek:
@chrismaltby Do you have any idea as to why this doesn't work on Windows?

Scene_b.c:256

```c
collisions_disabled = ReadBankedUBYTE(bank_ptr.bank, scene_load_col_ptr + 3);
```

chrismaltby:
@thomas-alrek quite a few merge conflicts on this now as I had to rewrite parts of the run loop to optimise enough to add animated actors with configurable animation/move speed. I ended up adding a few more events to control that too. Not too sure on merging the experimental parts yet as I want to release again soon, possibly needs to be a different request, how experimental is it?

RichardULZ:
@chrismaltby The main things that are experimental are 
+ #117 BG Script (Experimental) Run a script during player control
And possibly
+ #122 Feature/collisions disable

Not sure if collisions disable is working as intended, 
Currently functions as a noclip for actor (they dont test for collisions when moving, but are still collidable), might be better based on splitting actor.enabled into something like actor.collide & actor.visible.

BG script is inherently experimental, since the script runner was never expected to receive a new ScriptStart (by interacting with an actor) while running a script.
BUT It allows for a work around everyone seems to need, checking buttons while being able to walk, but also a lot of other possibilities.

The main experimental part is that "Script Complete = False" needs to be resolved for, in case you interact with someone, their script starts running, and an old function like "wait" completes.

You can definitely bring in
+ #104 Add support for notes field
+ #114 Set project author from username
+ #120 Make Editor Sidebar resizeable

But if you're looking for a hard to break build, these fancy features might need more discussion.

chrismaltby:
Thanks @RichardULZ I've merged the three you suggested separately

chrismaltby:
The other features sound great but think I need a bit of time to look over them and discuss, especially if they're not fully working yet

thomas-alrek:
Hi Chris. The experimental parts is really the BG Scripts. The rest of the features have been tested a lot, except for the Actor Invoke. That one has only been tested on Mac. And we should consider if the Actor Disable Collisions also should allow the Player to walk through the Actor

chrismaltby:
Hi Thomas, 

Maybe if I could get a separate pull request with just these two which I think will be pretty useful and shouldn't cause too much trouble with the updated engine.
> * #119 Add support for pushing and popping scene state
> * Feature/Invoke Actor

While I like the functionality that this #117 will enable, I definitely need more time to evaluate wether this is the right way to have background scripts and if theres any issues that this might cause.

> * #117 Implement BG Script (Experimental) Run a script during player control

There's the separate pull request for #117 already which I will try to review soon but I don't think this feature will make the next release.

> * #122 Feature/collisions disable

For #122, from what I can tell from looking at the code (might be wrong) seems like if you disabled collisions for an actor it doesn't allow the player to pass through them, the only way to allow that is to turn off all collisions for the player in that scene. I would assume the use case for this is more to have actors that the player can pass through. If that's the case I can probably have a look at this one myself, it should probably just be adding another check in `SceneNpcAt_b` but just need to be careful changing that fn as small changes can make big performance impacts so I'll need to do a lot of testing

The others have been merged now, made a few small changes but other than that all good

> * #104 Add support for notes field
> * #114 Set project author from username
> * #120 Make Editor Sidebar resizeable

thomas-alrek:
Alright, great. I will create a separate PR for the Actor Invoke stuff

PR #135, opened 2019-05-07
[Feature] Actor Invoke

thomas-alrek:
This PR implements a new event, **Actor Invoke**.

This event lets you invoke another Actor from inside your script, like a method call.
When that invocation is completed, it returns to where it was invoked, and continues as normal.
This also allows for recursion, which might be used for a lot of interesting purposes.

As of now, I have hardcoded the call stack depth to 8. I don't know how much would be viable to have, as I was concerned about memory usage. With a depth of 8, it should take about 16 bytes of memory.


chrismaltby:
Thanks Thomas, I merged the Scene State pull request but this one has conflicts now, it seems to partially include the Scene State code again but with different function names.

On a side note I'm thinking of changing `compileEntityEvents.js` a little so instead of hard coding the command code as a hex number, to instead just use the index in that array. That way we should have less merge conflicts as you wouldn't need to explicitly say `0x47` is `StackPop` you'd just need to make sure that the order of the commands in that array matches the order in the corresponding array in `ScriptRunner.c`

thomas-alrek:
@chrismaltby I have now fixed the conflicts

PR #137, opened 2019-05-08
Change Render order to Fix Vsync on game boy

RichardULZ:
Have a play with this, Reorders Update camera, shake, and scene render to the top, and fixes camera shake where the characters stay still by drawing actors after shake, unless this was intended.
Technically might add half a frame of extra input latency, but gains a rock solid feel and not buggy look.
Fades still seem to have vsync tearing.


chrismaltby:
Perfect! I guess you need to handle changing the SCX_REG and SCX_REG values as close to a vsync as possible to prevent the tearing. I've just made pretty much the same change for fades moving the call closer to the vsync and it seems to have fixed the issues there too! Thanks!

PR #139, opened 2019-05-09
BG Script, Run a script during player control

RichardULZ:
### Less Experimental BG Script!
The main components are:
- Internal variable for **BG_ptr, BG_ptr_bank, BG_start_ptr,** We store these on SetBGscript, and load them on script end.
- A **BGscript_Active** flag, To ignore "Is script ptr !=0" on collision and actor input.
- A script event, **Set BG Script** and **Clear BG Script**, copy's or reset's variables.
- Code in the **ScriptStart** event, to say BGscript not active, and _try_ and halt actor movement.
- Code in the **Script_End_b**, To check if there is a BGscript, and copy it over/ set active true. 
- Not sure when **if (!script_cmd_index)** (ScriptRunner.c line 126) is called true, but end check is also there, for now.
- **NEW** BGscript_actor, Copy the script actor across to stop it moving when it is halted.
- **NEW** Reset Time = 0 to solve wait's bleeding over to another script.

**Current issues**
+ [x] **Actor keeps moving if new script starts** Solved by moving BGscript_actor to nearest tile.
+ [x] **Fixed!** bg actor move would start moving any other actor you talk to, **patched in ScriptStart**.
+ [x] Issues with **IsLastFnComplete**, demo on UI text boxes auto completing, Try with looped Wait on a BG script.
+ [x] General performance, even *Wait for Joypad* can cause *some* lag while collisions+scrolling+music.   **Performance update** massively improves... performance... wait for actor input is **no longer slow**.
+ [x] Actor idle movement is allowed during BG script but **only** when using **Await Input**, Idle movement could be affected if all cases are allowed.

**TL:DR**
Script that runs in Background, Enables Player movement and collision while on, Stops and restarts after other scripts. 

Currently only way for start button menus, actors that do something while the player can move, or timers while the player moves for a score based or timed game.

## ALSO FIXES TRIGGERS
The performance update broke trigger movement, fixed with a last_joy = 0; on trigger start.
## PS
Windows builds since the performance update have this warning
`src/game.c(113):error *** struct/union/array '' :initialization needs curly braces`


RichardULZ:
Old pull request #117 if you want to know the thought process

thomas-alrek:
What is the reason that Actor Idle movement doesn't work while BG Script is active?

RichardULZ:
@thomas-alrek It's still running as a script, and there are a few more checks for idle movement, i'll try changing a few and hope for the best, but I expect some may be different with the new update.

RichardULZ:
@thomas-alrek Would need a special exception check for actors that are moving to/following, or else they start moving while idle... Animated frame cycles are unaffected and work in any script.

I understand it would be nice for actors to move idly if it's just for input checking, other cases may have issues, and idle movement is not always necessary, only other case i can think of is a timer.

Pushing a feature that allows idle movement when not script OR when bgscript active && await_input is anything. Start script now resets Await_input to 0.

RichardULZ:
@chrismaltby I think this is stable enough to integrate now.

gregtour:
@RichardULZ I'm interested in your thoughts on the BG script feature now. 

When you opened the pull request this was true:
> Currently only way for start button menus, actors that do something while the player can move, or timers while the player moves for a score based or timed game.

But now that there are input button scripts, it's possible to do start button menus. 

I have a pull request for a timer script that will run after a countdown interval here: #279 

Just comparing the code, it seems like it requires fewer changes now. Would you say there are benefits for background scripts over timed scripts?

PR #144, opened 2019-05-12
Adds Portuguese localisation

thomas-alrek:


chrismaltby:
Thanks!

Looks like I've added a few more keys since you did this, to make it easier going forwards for translations I've added a script you can run
`npm run missing-translations pt-PT`
which is giving the following output.

```
Missing translations for:

  "FIELD_VALUE": "Value",
  "FIELD_MIN_VALUE": "Min value",
  "FIELD_ANIMATION_FRAME": "Animation Frame",
  "FIELD_INITIAL_FRAME": "Initial Frame",
  "FIELD_MATH_NOTE": "Note: Variables have max values of 255 and will wrap if increased above 255 or below 0.",
  "FIELD_TRUE": "True",
  "FIELD_FALSE": "False",
  "FIELD_VARIABLE": "Variable",
  "FIELD_RANDOM": "Random",
  "FIELD_SET_TO": "Set To",
  "EVENT_ACTOR_SET_FRAME": "Actor: Set Animation Frame",
  "EVENT_VARIABLE_MATH": "Variable: Math Functions",
```

Going to try keeping that list fixed now as I'm on release candidate 2 (looking to launch officially next week) and will be spending the next few days looking at site documentation now.

toxworks:
chrismaltby sorry If this sounds a bit dumb, but what is the procedure to run that .js you created for missing translations?

PR #224, opened 2019-06-18
New feature: Set a custom replacement palette via Game Boy Color

fydo:
Hello!

Thanks for considering my pull request. I have done this because it seemed like a popular feature request in the community Discord. 

This change allows users to set their own 4 custom 15-bit colors to replace the standard Game Boy colors in both ROM and web deployments. In the World settings panel, users can select the color in the palette, then either type in the 0-31 RGB values or paste in an HTML Hex color code to see the color update in real-time.

The ROMs generated by this change have been tested on many different types of hardware by community member Veund. It was found that the custom colors worked on everything that supports Gameboy Color (including the GBA Player for the Gamecube).

I have noticed that the colors output by the emulators are a little washed out compared to the preview colors I generate in GB Studio. It could be a colorspace issue (I'm currently just multiplying by 8 to expand from 0-31 to 0-255) but I'll need to do more testing to see if there's a problem there.

In the future, it would be nice to also add support for Super Game Boy custom palettes. As it stands, GBDK doesn't provide the necessary functions for sending the palettes and I currently don't have enough experience with assembly to add the needed functions.

**Does this PR introduce a breaking change?**
Using an existing project file or creating a new one should both function. I have only tested using the English localization file so it's possible that using a different language may break things.

**Other information**
I was hoping to also add to the documentation but it appears that it does not reside on this repository.
Additionally, no tests have been added either.

Special thanks to community Discord members:
* **Dramakins** for initial inspiration
* **rulz** for assistance in developing the Game Boy ROM changes
* **Veund** for feedback and extensive hardware testing
* **personitis, Darkhog, Kenney, Kazy, and TOXWORKS** for feedback during UI development


chrismaltby:
Hi @fydo Thanks for this pull request, sorry it took a while to get back to you on this one, I've been busy refactoring almost the entire frontend and event system which I've just merged into develop.
 I'm going to clone your repo and have a full look through your implementation, and do some testing before I can merge this.

Luckily you completely avoided merge conflicts with my changes so that makes it a bit easier!

I'm thinking of putting features like this and probably things like the Cartridge Type under an "Advanced Settings" section where I've got a few other ideas (custom keyboard controls, inject CSS/JS into the head of the web player) eventually but I think leaving it where you have it is fine for now. I'll try to get back to you soon.

fydo:
Hey there! Thanks for the response. No worries on taking your time, I'm certainly not in any kind of hurry. :) I think I can speak for the community when I say that we really appreciate the time and care that you put into GB Studio. Please let me know if you require me to make any changes to my commits thus far.

Also, sorry about the Softagram spam. I was playing around with it and I didn't expect it to post a comment in here.

chrismaltby:
Hi @fydo Thanks for your work on this, I really appreciate you helping out on this highly requested feature.

I've had a little time to do some testing, I think there might be a few issues on your implementation though. Some of colours aren't coming through correctly when the ROM is being built.

e.g. See this one where blue has become a darker green and orange has become red in the build.
Screenshot 2019-06-29 at 20 25 46
Also I noticed when reloading that file the blue had changed to the green (but strangely the orange wasn't red)
Screenshot 2019-06-29 at 20 26 09
I've attached the project so you can see
[UntitledDX.zip](https://github.com/chrismaltby/gb-studio/files/3342262/UntitledDX.zip)

I think the blue become green is because of a bug around here
https://github.com/fydo/gb-studio/blob/e62466458a06b2f3a2fce145b1a6b6848f2cd776/src/components/forms/CustomPalettePicker.js#L128-L155
You store changes with `this.setState()` then call `this.saveToSettings();` straight afterwards that reads the state using `this.state.` which isn't guaranteed to work as `this.setState` is asynchronous.
One thing you could do is use the callback argument on setState (see https://medium.learnreact.com/setstate-takes-a-callback-1f71ad5d2296) 

I think adding the saveToSettings call to the callback on the four setState calls might be enough to fix this.
```
      this.setState({whiteHex:  this.decimalToHexString(r * 8) + 
                                this.decimalToHexString(g * 8) + 
                                this.decimalToHexString(b * 8)}, this.saveToSettings);
```

I'm not sure why the orange becomes red though, the colour seems to being stored correctly in the .gbsproj file as `"CustomColors_White": "F0A800",` but for some reason this hex comes through as red.

Here's a few more tests I did with similar issues:
Screenshot 2019-06-29 at 20 21 29
Screenshot 2019-06-29 at 20 20 41

I'm using macOS if it makes a difference.

fydo:
Hello! Thanks for the great feedback. You were absolutely right with regards to the issue being React's state change being async. What was happening in your first screenshot is that you typed in 2,25,30 but since the most recent event wasn't getting applied to the state you ended up with a color that was 2,25,3. 

I just checked in a fix that appears to solve this issue.
fix

fydo:
Just to follow up on my "washed out colors" comment. I found some information here: http://emulation.gametechwiki.com/index.php/Game_Boy/Game_Boy_Color_emulators#Oversaturation

It seems that GBC emulators will attempt to emulate the under-saturation of the GBC screen. I don't know if this is something we want to account for but I am personally happy with how the implementation is at the moment.

chrismaltby:
@fydo yeah I think we can leave the colours as is for the moment but thanks for investigating.
Just to let you know I've just pushed a build that moves this feature into a Settings section now (I've not deleted it! :D)

fydo:
Haha, thanks for the heads up!

Issue #189, opened 2019-05-21
Adding more dialog to existing dialog blocks breaks if a text break happens...


timsamoff:
**Note: for support questions, please use one of the following channels**.

Reddit: [/r/gbstudio](https://www.reddit.com/r/gbstudio)
Discord: [Join Chat](https://discord.gg/bxerKnc)

This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**

Bug.

* **What is the current behavior?**

If a dialog block already exists (and contains text), adding new text to the text box will break if a line break occurs. (Sorry for the terrible description of this behavior, so please watch the GIF:

![dialog_entering_bug](https://user-images.githubusercontent.com/62411/58112280-0c253d00-7ba8-11e9-82be-61cef261f412.gif)

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Add a dialog block to an element's script; add text to it.

Then place the cursor at the beginning of the text and start typing again.

* **What is the expected behavior?**

Text would be entered as expected, but stop when max characters ran out.

* **What is the motivation / use case for changing the behavior?**

This has wasted a lot of time during my development.

* **Please tell us about your environment:**

  - Version: 1.0.0
  - Platform: macOS

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

thomas-alrek:
Just to clarify, is this happening in 1.0.0 or 1.1.0? Because the screenshot seems to be from 1.1.0

timsamoff:
Apologies... 1.1.0, but since 1.0.0.

PR #228, opened 2019-06-22
Restore cursor position after textarea has been edited.

MattTuttle:

This is a bug fix for the dialogue textarea.

The cursor was jumping when editing a section of the textarea. You can reproduce by adding text to the textarea, moving the cursor to an earlier point, and typing more text. This is because the textarea contents are rewritten in the dialogue box.

The textarea saves and restores the correct cursor position when editing.

### DIFF
--- a/src/components/script/ScriptEventBlock.js
+++ b/src/components/script/ScriptEventBlock.js
@@ -30,7 +30,8 @@ class ScriptEventInput extends Component {
   onChange = e => {
     const { onChange, field, value, index } = this.props;
     const { type, updateFn } = field;
-    let newValue = e.currentTarget ? castEventValue(e) : e;
+    let el = e.currentTarget;
+    let newValue = el ? castEventValue(e) : e;
     if (type === "direction" && newValue === value) {
       // Toggle direction
       newValue = "";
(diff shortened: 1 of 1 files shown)


MattTuttle:
Somewhat related. I noticed that when adding a new textarea in the develop branch it puts focus on the first textarea instead of the recently added one. This seems to be a bug prior to this commit.

RichardULZ:
Is this the same issue as #189

chrismaltby:
Looks like a good fix, unfortunately has a few merge conflicts now because of a refactor I'd been working on. Those files should be relatively stable now if you want to update at some point. Thanks

MattTuttle:
It was easier to rewrite the code since ScriptEventInput is now a Component. I'm fairly new to React so hopefully the forceUpdate function doesn't have unwanted side effects.

@RichardULZ Yes, this should fix that issue.

MattTuttle:
I'm not sure which way is better but I separated the textarea field into it's own component. This removes the need for the weird if condition.

chrismaltby:
Excellent, that's much better now. Thanks for your help!

PR #245, opened 2019-07-12
Stop Static multiframe actors animating when moving

### LINKED ISSUE #237: [bug] actors changing frames during move,ignoring wait 1.2alpha1
im moving an actor then setting its position then changin the frame the actor appears to be changing frame during the move, i tried counteracting this with a wait but the wait appears to happen after the frame chaneg

2 frame actor (0), actor move to desired loaction then set to frame 1 

**Expected behavior**
the actor to move to location then change frame 

windows 7
1.2 alpha1

RichardULZ:

Bug fix #237 add check to general actors update so a moving static is not forced to animate.

A multi frame actor set to static would still animate when moving via script.

A multi frame actor set to static will not animate when moving via script, unless set directly.

If the user intended this effect, they could switch to face interaction? 
Made sure that a cycling static actor will still animate when moving.

### DIFF
--- a/appData/src/gb/src/Scene_b.c
+++ b/appData/src/gb/src/Scene_b.c
@@ -587,7 +587,7 @@ void SceneUpdateActors_b()
     {
       if (ACTOR_ANIM_SPEED(ptr) == 4 || (ACTOR_ANIM_SPEED(ptr) == 3 && IS_FRAME_16) || (ACTOR_ANIM_SPEED(ptr) == 2 && IS_FRAME_32) || (ACTOR_ANIM_SPEED(ptr) == 1 && IS_FRAME_64) || (ACTOR_ANIM_SPEED(ptr) == 0 && IS_FRAME_128))
       {
-        if (ACTOR_MOVING(ptr) || ACTOR_ANIMATE(ptr))
+        if ((ACTOR_MOVING(ptr) & actors[i].sprite_type != SPRITE_STATIC) || ACTOR_ANIMATE(ptr))
         {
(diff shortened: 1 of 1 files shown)


appData/src/gb/src/Scene_b.c
```diff
@@ -587,7 +587,7 @@ void SceneUpdateActors_b()
     {
       if (ACTOR_ANIM_SPEED(ptr) == 4 || (ACTOR_ANIM_SPEED(ptr) == 3 && IS_FRAME_16) || (ACTOR_ANIM_SPEED(ptr) == 2 && IS_FRAME_32) || (ACTOR_ANIM_SPEED(ptr) == 1 && IS_FRAME_64) || (ACTOR_ANIM_SPEED(ptr) == 0 && IS_FRAME_128))
       {
-        if (ACTOR_MOVING(ptr) || ACTOR_ANIMATE(ptr))
+        if ((ACTOR_MOVING(ptr) & actors[i].sprite_type != SPRITE_STATIC) || ACTOR_ANIMATE(ptr))
```
chrismaltby: I think this is supposed to be a logical AND `&&` rather than a bitwise one `&`

chrismaltby:
Nice one, looks to work, just got a little change because it looks like you probably want an extra `&` in there

appData/src/gb/src/Scene_b.c
RichardULZ: Good spotting, i'll change that up.
Just checked because i was curious, switching to logical `&&` seems a hair faster than bit wise & here, in bgb cpu.

PR #272, opened 2019-08-27
Add basic sound effects with Play Tone, Play Beep, and Play Crash

gregtour:

This is a feature to add basic sound effects to GB Studio.

This adds three new sound events. See below.

* **Other information**:

These are three simple sound generating functions that enable users to play basic sound effects for games. The crash sound works well for moving objects and entering doors. The beep sound is also useful for sounds when interacting with objects. And the tone event can be used to play simple chimes without creating a mod file.

Please try these features out and let me know if you have feedback.

### Documentation

**Sound: Play Tone**
SoundPlayTone

Plays a tone with a specific frequency and duration. The sound plays on channel 1 and blocks the script until it is complete. The sound might not be heard if music is playing.

**Sound: Play Beep**
SoundPlayBeep

Plays a short beep with a high or low pitch on channel 4. The sound plays in the background. The pitch is set with a number between 1 and 8. The sound might not be heard if music with percussion is playing.

**Sound: Play Crash**
SoundPlayCrash

Plays a crash symbol sound on channel 4. The sound plays in the background. The sound might not be heard if music with percussion is playing.


chrismaltby:
Excellent!! This is great!

I like how you can chain the "Play Tone" events together to make a little jingle too. Not had much time to test yet but it seems like "Play Beep" and "Play Crash" don't wait though, not sure if they should or not really, open to discussion on that. I suppose you could manually add "Wait" commands if you wanted to have a crash, then a pause then a crash again.

Either way could you maybe do a little test project for this one in `test/projects` too? Might be good to have some examples of how they work used one after the other and how they work when music is playing for example. Then I'll do some testing across all the platforms.

Thanks!

gregtour:
Added a test project. This one also causes a game crash on Windows when using the beep function.

RichardULZ:
Here's the method ZGB another GBDK based gameboy tool uses for these kind of sounds,
https://github.com/Zal0/ZGB/blob/master/common/src/Sound.c
https://github.com/Zal0/ZGB/wiki/Sounds

It would be nice to expose more of the sound register functions, as there are volume envelopes (fade in/out) for most channels, and dedicated pitch sweeps on channel 1. 
They also all have a hardware timer called Length load to shut a note off automatically without code.
All these could make for very diverse sounds.

Also, if registers were pushed instead of just pitch, like **NRx3** and **NRx4**, the expensive math could be done during the js build process, rather than when the sound is called on the gameboy.

An example of inputs.
**Volume** (0 to 15), **Vol Envelope** (-7 to +7),   **NR12** +(trigger **NR14**)
**Frequency** (11bit, Round `131072/(2048-x)` where `x` is in Hz ), **NR13+NR14**
**Wave type\Duty** (0-3), **Length** (0-64 max .25 seconds?)  **NR11**+(length on **NR14**)

**Sweep Speed** (0 to 7), **Sweep direction** (-7 to +7) **NR10**

Here are some great references on what each bit of the registers are for.
http://gbdev.gg8.se/wiki/articles/Gameboy_sound_hardware#Registers more Visual
http://bgb.bircd.org/pandocs.htm#soundoverview more Info

PS: Fantastic job getting GBDK 2.96 to compile!

gregtour:
@RichardULZ 
The frequency value is computed in the compile step: https://github.com/chrismaltby/gb-studio/blob/77bf50d5f55c660cecd27f4f4ed8fd323c54c5a4/src/lib/events/eventSoundPlayTone.js#L32

I think each of those sound features can be used from GBT already, so the main purpose for these changes are just to provide some very easy functions for generating sound and possibly allow concurrent playback in some cases.

+1 for the Pandoc links, those are always a good resource.

chrismaltby:
I've just fixed up the merge conflicts on this.

Been having a play with changing NR14 and NR10 in `Script_SoundStartTone_b` could be interesting to expose these but I think it could come at a later date. It would probably need some way of roughly previewing the sound from the editor if the functionality got much more complex though. I always liked how [BFXR](https://www.bfxr.net) had good presets to make common game sounds, maybe one day we can add something like a mini version of that :-D

If this is working okay in Windows since the 2.96 update what do we think about merging it now though? @gregtour @RichardULZ

chrismaltby:
Hey @gregtour, I had another look through this today, working on the UX and decided its simpler to merge the three events into a single "Sound: Play Effect" event which I've now done.

I've also added the ability to preview any of the sounds directly from the editor similar to how previewing music works with a play button on the effect type dropdown. For crash and beep I've just recorded mp3s of the audio direct from the JS emulator so they should sound identical, for tone I'm using a web audio square wave oscillator which doesn't sound exactly right but should be close enough for now.

New sound effect event UI looks like this with the inputs changing depending on which type is selected.

Screenshot 2019-10-12 at 19 11 14

Looks to be working across all the platforms so I've finally merged this in. Thanks for your help!

PR #277, opened 2019-08-28
Comment script event

ManuGamesDev:

Adding a comment script event.

No

* **Other information**:
image
image


chrismaltby:
@ManuGamesDev nice, there was actually an event in place already for this purpose called "Notes" that worked pretty much the same way. I named it that way for consistency with the notes fields on Scenes/Actors/Triggers but actually "Comment" is probably more likely what people would expect it to be named.

I like the highlight colour you've added to yours too, makes it stand out much more.

What do you think, should we just remove eventsNotes.js and use this version instead?

ManuGamesDev:
@chrismaltby to be honest i didn't know that there was something like that already in gb studio lol people on the discord told me there was actually a notes event already, but they also said they liked my comment one better. So yes, I think we should replace it with notes

chrismaltby:
I agree, I've updated your pull request to remove the Notes event and merged that into develop now. Thanks!

Issue #281, opened 2019-09-01
[Bug] "Variables: Add Flags" does not function when using windows10


locherunderscore:
The following script function:
"Variables: Add Flags"
does not work when using gbstudio on windows10.

No values are changed and when outputted via dialog box, the result is always 0.

[Set Flags] will successfully set the flags and [Clear Flags] successfully clears flags.
I want to ask if you could look into how [Add Flags] is implemented in order to find out why this script function is compiled successfully on Macs and is wholely ignored for windows 10 pcs.

pau-tomas:
Here's a simple project to test the case. [bitwise_sample.zip](https://github.com/chrismaltby/gb-studio/files/3563804/bitwise_sample.zip)

The expected behavior would be for the `$01$` variable to alternate between `0` and `1` everytime you talk with the Sage actor.

On Windows 10 (from @LocherGit tests) the value is always `0`.

@chrismaltby I don't have a Windows 10 machine to test, but I'll try to do some investigation.

locherunderscore:
I closed this due to the merge request resolving my issue.
At least the expected behavior was observed on my win10 computer.

PR #279, opened 2019-08-28
Add set timer script, restart timer, disable timer events

gregtour:
* **Please check if the PR fulfills these requirements**
- [X] The commit message follows our [guidelines](/chrismaltby/gb-studio/blob/develop/.github/COMMIT_MESSAGE_GUIDELINES.md)
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This PR introduces a timer script feature.

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**

See below.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

* **Other information**:

* **Documentation**:

**Timer: Set timer script**
SetTimerScript

Execute the specified script after a fixed amount of time. The timer will repeat until it is disabled or the scene changes. The timer does not count down when a script is running or dialogue is open. The timer duration is between 0.5 seconds and 60 seconds. There is only one timer script at a time.

**Timer: Restart timer**
RestartTimer

Restart the timer countdown.

**Timer: Disable timer script**
DisableTimer
Remove the timer and disable the timer script. The timer is automatically removed when the scene changes.

PR #283, opened 2019-09-05
Update win32-x64 platform to GBDK version 2.96(a)

gregtour:

Bug fix.

GB Studio on Windows uses gbdk 2.95 to produce Game Boy binaries while MacOS and Linux versions of GB Studio use gbdk 2.96. Because of these differences, some (newer) features malfunction when built on Windows.

See: #281 #272 #279 

It is possible that for users on Windows, some in-game behavior or performance may change when switching between 2.95 and 2.96. Because the differences between 2.95 and 2.96 are unknown, it is difficult to predict how many issues there could be. There are currently no documented issues.

* **Build information**:

GBDK binaries were built with MinGW64 and MSYS2. The GBDK source code was pulled from https://github.com/gheja/gbdk. 
The following issues are encountered:

- Types of _IO_ssize_t must be replaced with ssize_t
- scpp (sdcc/support/cpp2) must be configured with autoconf 
- The scpp Makefile must be edited to remove syntax errors
- as-gbz80 and link-gbz80 are not copied to the build/bin folder and must be located in sdcc 

* **Other information**:

Ideally, project files would produce identical .gb files when built on MacOS, Linux, and Windows. Even with GBDK 2.96, there seem to be some hex differences from builds on each platform. Builds seem to have more consistent behavior when built with the same release of GBDK, however.


chrismaltby:
Amazing, been wanting to make this change for a long time now!

Is this pull request work in progress though by any chance? I've just booted up my pc to give it a go and it looks like `lcc.exe` is no longer included in `buildTools/win32-x64/gbdk/bin` but https://github.com/chrismaltby/gb-studio/blob/develop/src/lib/compiler/buildMakeBat.js is still referencing that exe to compile. Looks like maybe there's some changes needed to the build script for Windows to make this work.

Do you know by any chance if the binaries are 32bit and will just work on ia32 architecture? GB Studio didn't initially support Windows 32bit but I was receiving quite a lot of requests for it, would be nice not to break compatibility there if possible.

If we can get this update to work I think this should really help with the issues you've been seeing on Windows with the other pull requests too.

Thanks for all the work you've been putting in here, it's been crazy at work so I'm not getting as much time as I'd like to look at this so really appreciate what you're doing.

gregtour:
@chrismaltby It's still a work in progress but it's close. I'll see if I can make a 32-bit build, too.

chrismaltby:
Finally had some time to test on Windows, seems to work great 👍 Still would be interested in a 32-bit build, do you have some build instructions so I can have a go? I can see if I can get hold of 32-bit Windows system (or run one in a VM) and try to compile directly.

PR #284, opened 2019-09-07
Add support for text dialogues with avatars

pau-tomas:

Adds a new optional field to the Text event allowing to select a static sprite and use it as an avatar that displays to the left of the dialogue box.

image

image

Dialogue boxes with avatar can display up to three lines of text and 16 characters per line.

The text dialogue doesn't have this option.

The option to select a sprite to be used as an avatar appears in all the dialogues. It defaults to `None` for any dialogue that already exists in the project.

From my testings the change shouldn't affect existing dialogues.

* **Other information**:

When used as avatars, sprite PNGs must contain the same colors as the background images. This means that the color `#306850` can be used for avatars. However, if the sprite is used as a sprite for an actor or the player, the color will be matched to the nearest color from the background PNGs color list.

Sprites with transparency (the color `#65ff00`) can be used as avatar, but the transparency will be replaced by the color `#e0f8cf` when displayed in the dialogue. Transparency will work as expected when the sprite is used for an actor or player.


src/lib/events/eventTextDialogue.js
```diff
@@ -9,9 +9,17 @@ export const fields = [
     type: "textarea",
     maxPerLine: 18,
     placeholder: l10n("FIELD_TEXT_PLACEHOLDER"),
-    updateFn: trimlines,
+    updateFn: (string, field) => trimlines(string, field.maxPerLine),
```
pau-tomas: I can't find a simple way to enforce the 16 characters per line limit when an avatar is added. @chrismaltby do you have any suggestion?

chrismaltby:
Really nice feature!

I've just pushed an update to your pull request limiting the character limits to 16 per line when using avatars. I'm wondering though if this should always be visible on every `Display Dialogue` event as not everyone will need it.

Here's a few options we could go with:
1. Have this as a separate event, `Display Dialogue With Avatar`. Should just be a case of copy/pasting the current eventTextDialogue.js and making a few changes to each.
2. Having a global option in the Settings section, for `Include Avatars in Dialogue` and only show the Avatar field if it's enabled.
3. We could collapse the avatar dropdown as a link in a similar way to how "Add Notes" is at the moment, probably doesn't save much space though.

I'm leaning towards option 1 but open to discussion. What do you think?

pau-tomas:
Sorry! I forgot about Dialogue Review. Thanks for the update!

I thought about having the option always visible, and I agree with you that it's probably too much. Happy to change it to something else.

My concern about option 1 is that if somebody wants to update their project to support avatars will have to replace all the `Display dialogue` events with the new one. Also in code (almost) everytime a check for `EVENT_TEXT` is done there will need to be `EVENT_TEXT || EVENT_TEXT_WITH_AVATAR`, although this can be solved with a `isTextEvent()` function or similar.

I like option 3, a quick test with the dev tools shows that some good space saving with it.

Screenshot 2019-09-08 at 23 29 18
Screenshot 2019-09-08 at 23 28 55

Another option would be to add `Add Avatar` to the ▼ menu for `Display dialogue` events. It might be a bit hidden, but maybe once it's used once for the field appears by default on all the `Display dialogue` events for that Script?

chrismaltby:
Thinking about it a bit more I think you're right, Option 1 isn't great if you want to add avatars to an existing project and probably will get a bit messy.

I've just pushed an update that uses option 3 to collapse, seem to work quite nicely actually and seems like a good balance between not taking up too much space if you don't use it but not so hidden that people won't find it. I do like your other option of adding an additional icon though, possibly a bit tricky to implement, maybe we could revisit this later.

Let me know what you think of this update, if you're happy could probably merge as it is now.

pau-tomas:
Works great on my testing, so I think it's ready to merge.

RichardULZ:
Windows 10 exporting the Test_TextWithAvatar gbsproj from latest dev build. 
This function crashes in game as soon as you call an event with avatar+text.

pau-tomas:
@RichardULZ could you check if rebasing the changes from https://github.com/chrismaltby/gb-studio/pull/283 into `develop` fixes the issue?

RichardULZ:
@pau-tomas The recompile of GBDK 2.96 from #283 seems to fix the issue!

pau-tomas:
Unfortunately, it seems like without 2.96 on Windows lots of the new features are broken. Hopefully it'll be merged soon :) Thanks for checking and sorry for the trouble!

chrismaltby:
I've merge in the 2.96 update for Windows 64-bit, seems to work for me now 👍Think we'll probably still have issues in 32-bit until we can get a version of 2.96 working there

Issue #289, opened 2019-09-15
A Settings checkbox to set CGB speed fast (improves some audio hitches)


RichardULZ:
**Is your feature request related to a problem? Please describe.**
Temporary improvement to music audio hitches that can occur when drawing text, changing scenes, or doing complex functions. 

**Describe the solution you'd like**
A checkbox option near to Custom Colors that will, on GameBoy Color, run the CPU twice as fast.

Simple as this.
`game.c`
```cpp
#ifdef FAST_CPU
if (_cpu == CGB_TYPE)
{
 cpu_fast();
}
```
I would add it myself, but the .js front end is scary.

**Describe alternatives you've considered**
The real solution is splitting functions known to have slowdown into more chunks based on how much data there is to load, this is difficult due to **wait vblank** being **immediate** while the screen is off. 

This solution is not perfect, as smaller scenes load too fast (wait vbl won't work) and can cause a song to skip ahead, but loading text is better, and i've seen some ambitious games suffering from slowdown.

PR #290, opened 2019-09-15
Add settings option for double speed CPU

pau-tomas:

Add settings option for double speed CPU. 

See https://github.com/chrismaltby/gb-studio/issues/289 for more details

* **Other information**:


src/lib/compiler/makeBuild.js
```diff
@@ -106,6 +106,9 @@ const makeBuild = ({
     } else {
       gameHeader = gameHeader.replace(/#define CUSTOM_COLORS/g, '');
     }
+    if(!settings.gbcFastCPUEnabled) {
+      gameHeader = gameHeader.replace(/#define FAST_CPU/g, '');
+    }
     await fs.writeFile(`${buildRoot}/include/game.h`, gameHeader, "utf8");
 
     // Remove GBC Rombyte Offset from Makefile (OSX/Linux) if custom colors not enabled
```
pau-tomas: 🤔 I think this need to be changed to remove it when both custom colors and fast cpu enabled are not enabled.

chrismaltby:
Seems to work great! Is this ready to merge?

With this enabled I'm not noticing the music skips while switching scenes anymore. Are there any downsides to using the feature at all?

If there's no reason to ever turn it off, I might rethink some of the UI around GBC support at some point and just make a single setting to turn on/off GBC support that states 2x CPU speed as a benefit.

Thanks @pau-tomas and @RichardULZ

RichardULZ:
Reasons not to use it, 
Slight hit to battery power on real hardware, may be harder to emulate on slow devices like phones, not much else, may be best to leave it on.

Reasons to still make a GB only game.
Swappable color palettes when running a gb only game on a real CGB, and faster to emulate on slow devices.

Custom palettes must be enabled when using any CGB feature, so bundling 2x cpu would work great.

I have some modifications to the load scene function that further improve skipping audio in fast cpu mode, such as turning on the display after graphics are loaded, as wait vblank is immediate while screen off, and can cause a complex song to jump 3-4 notes ahead.

With double cpu, Small scenes have no hitch, large scenes have a tiny hitch, text boxes produce no hitch.

Could call GBT twice for large scenes as a crude fix.

pau-tomas:
@RichardULZ do you think this is ready to merge? I can take it out from draft if that's the case.

RichardULZ:
@pau-tomas 
There's a tiny issue with the settings page when gb studio is in a **small window**, 
Enable, palette, then enable Double CPU, This can create an empty space covering the bottom of the screen, pushing the title bar off screen. Only shows when gb studio is fairly small. likely due to reusing some code from opening the palette picker. Only tested on windows 10.

The palette picker should probably be open whenever double CPU speed is on though, may cause confusion if they set a palette, disabled it, but find the palette is still being used.

Other option is combining double CPU speed with GBC palette picker, 
The only real downside may be on slower phones, but that's only noticeable when music is playing, and anyone with music in game will likely turn this setting on...

So, 2 options.
- Fix the current settings page when small, maybe force colors to stay open too... 
- Rename `Export With Custom Palette` to also say `& Double CPU Speed (GBC)` as @chrismaltby suggested, since there is very little downside.

pau-tomas:
@chrismaltby I noticed that Game Boy isn't mentioned anywhere in the app so I was wondering if it's due to ©️  concerns or if it would be ok to rename the `Colors` settings section to `Game Boy Color options` and add the checkboxes for Custom Palette and Double CPU Speed? Maybe having a not stating that enabling any of the options will generate roms only compatible with GBC.

RichardULZ:
@pau-tomas I believe the games that use GBC features are still compatible with original Gameboy hardware, as the system type is checked before using those functions, The extra note may not be necessary.

pau-tomas:
Ah, my bad. Scratch the extra note then.

I guess we won't get into "GBC only" territory until GBS provides a way to use the extra tile maps  and so.

RichardULZ:
Have done some testing, there are practically no downsides to having Double Speed CPU combined with GBC palette.

Testing on a low end phone that stutters with audio, performance was near identical between 1x and 2x speed roms.
As for battery life on real hardware, https://azug.minpet.unibas.ch//~lukas/GBprojects/gbFAQ.html#PowerUsage 
You'll use an extra 7-8 milliamps of power, 10%, Not worth worrying about if it means better sound.

chrismaltby:
Coming back to this I think it might be better to let people decide if they want double speed mode enabled and keep it as separate option.

Just made a few updates to reorganise these settings and fix some of the issues that @RichardULZ was seeing.

- The display issue where the toolbar disappeared when toggling the checkbox on a small screen was one of those weird CSS bugs that fixed itself by adding `transform: translate3d(0,0,0)` to the settings page container forcing hardware acceleration for that `div`

- The double speed mode now uses a default GBC palette if you enable it without checking "custom palette"

- Updated the default GBC palette to match the suggestion from @proximitysound in #308

I think this might be ready to merge now, @pau-tomas do you want to take the pull request out of draft?

Issue #291, opened 2019-09-16
Avatar + Text Overflow Corrupting Background Graphics


SetsuneW:
**Describe the bug**
Filling the text box with Avatars enabled eventually overwrites background tiles.

**To Reproduce**
1. Actor -> Text: Display Dialogue
2. Apply an Avatar
3. Fill text box with 3 lines of "1234567891234567" (the limit in the editor).
4. Build and talk to Actor.

**Expected behavior**
Either the text to display correctly, or the Editor text box to truncate the text, as it did previously to 18/18/16.

**Screenshots**
![GBStudioTextSpriteBug20190916](https://user-images.githubusercontent.com/50024836/64941973-f06c3380-d835-11e9-800a-c3540eca45e6.png)
(I skipped 0 in my text string.)

**Platform (please complete the following information):**
 - OS: Windows 10 Build 18362
 - App version Commit 8366701 (Dev Build)

**Additional context**
N/A

RichardULZ:
This issue was fixed by pau-tomas, and can now be closed.

PR #292, opened 2019-09-16
Skip new line character when adding to dialogue

pau-tomas:

#284 introduced a bug in any dialogue (with or without avatar) that counted the line break characters as a new tile, effectively reducing the max numbers of characters by 2 and allowing the overflow reported in issue #291. 

This fix reverts to the same behavior as before #284 but using a slightly different code.

The image shows how the following text is stored in the tile data for each of those versions:
```
123456789123456789
123456789123456789
1234567891234567
```

image

No functional change but fixes issue https://github.com/chrismaltby/gb-studio/issues/291

The dialogue tiles will overflown if the max character per line was used on all three lines

When rendering text dialogues, instead of using an empty tile for new line character is found just skip it and render the next character in the next line.

N/A

* **Other information**:

### DIFF
--- a/appData/src/gb/src/UI.c
+++ b/appData/src/gb/src/UI.c
@@ -22,6 +22,7 @@ UBYTE text_x;
 UBYTE text_y;
 UBYTE text_drawn;
 UBYTE text_count;
+UBYTE text_tile_count;
 UBYTE text_wait;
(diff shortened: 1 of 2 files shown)


chrismaltby:
Good spot, works perfectly now from my testing. Thanks!

Issue #44, opened 2019-04-24
[Request] Support VGM and GBS Music Format


clsource:
Hello I think this project is awesome.
Here are some ideas.

Mod file format support is great but there are some tools better suited for the job like deflemask http://deflemask.com/ that export gbs format.

https://mmitch.github.io/gbsplay/
http://www.angelfire.com/nc/ugetab/
http://web.archive.org/web/20171128170657/ftp://ftp.modland.com/pub/documents/format_documentation/Gameboy%20Sound%20System%20(.gbs).txt

[GBS2GB117.zip](https://github.com/chrismaltby/gb-studio/files/3112328/GBS2GB117.zip)

VGM is also a good choice https://en.wikipedia.org/wiki/VGM_(file_format)

Here is a player for the gameboy

http://jiggawatt.org/badc0de/dmvgm-4.zip

[dmvgm-4.zip](https://github.com/chrismaltby/gb-studio/files/3112086/dmvgm-4.zip)

Additional References

https://vgmrips.net/forum/viewtopic.php?f=3&t=107
https://github.com/Sanqui/gbvgm
https://github.com/io7m/jvgm
https://github.com/kuma4649/mml2vgm
https://github.com/dzidzitop/vgmtag

Finally supporting midi files may be possible using 

https://sourceforge.net/projects/midi2mod/
or mid2vgm https://vgmrips.net/forum/viewtopic.php?t=1929

Thanks 👍

chrismaltby:
Thanks, lots of useful information there!
Do you know what the CPU overhead is for this by any chance? According to the docs GBTPlayer seems to be around 7%.

If anyone would like to have a go at implementing, these two files are probably a good place to start
https://github.com/chrismaltby/gb-studio/blob/master/appData/src/gb/src/MusicManager.c
https://github.com/chrismaltby/gb-studio/blob/master/src/lib/compiler/compileMusic.js

clsource:
Another way of doing this is creating a converter from `vgm` to `gbt` that should be easier to do i think.

Reference
http://www.smspower.org/Music/VGMFileFormat?from=Development.VGMFormat

Some Libs

In Python
https://github.com/jimbo1qaz/vgmviz
https://github.com/cdodd/vgmparse

In JS
https://github.com/apollolux/vgm-js

Sample VGM FIle

[...]

PR #294, opened 2019-09-19
Feature Volume envelopes in Mod2GBT

RichardULZ:

Feature, Volume envelopes in Mod2GBT via Volume sweep effect A00.
Adding features from #44 

You must set the volume for each level to fade in or out, sounds bad on real hardware or emulators.

You can use hardware volume sweeps on ch1, 2, and 4 for much nicer in game music. 
.
Set at the start of a song or mid way through, can be removed with A00 or a new song.
Asm instructions added gbt_player_bank1 that merge volume and envelope bits for ch1,2,4, these add 4 extra instructions when changing volume to preserve the envelope, and 6 instructions for loading the effect.
A0F = fade out notes, AF0 = fade in notes.
A00, A01, and A10 will stop the envelope.

All previous songs should remain compatible.

* **Other information**:
### Can not compile mod2gbt.c properly
Using Codeblocks on windows 10 with minGW, it compiles, but produces `ENOENT: file not found` in gbstudio as it outputs music to the wrong location and as .asm.
Help is needed @chrismaltby


chrismaltby:
@RichardULZ Nice one.

I've checkout out your branch and compiled your CodeBlocks project, I was seeing the file not found issue but the file that was generated was fine.

I believe the issue is because the previous `mod2gbt.exe` binary is a precompiled one I've had for a few years that has a minor difference to the other platforms in that the output file was always named `output.c` where on Mac/Linux the binary creates `music.c` instead. I don't believe theres any CLI args in mod2gbt to change the output filename so in `compileMusic.js` I detect the platform to determine the output filename

https://github.com/chrismaltby/gb-studio/blob/8366701e7ac7c0642b88cefb549a0a69876dc2e8/src/lib/compiler/compileMusic.js#L50

then rename it and move to the correct location after mod2gbt runs

https://github.com/chrismaltby/gb-studio/blob/8366701e7ac7c0642b88cefb549a0a69876dc2e8/src/lib/compiler/compileMusic.js#L88-L91

It looks like the new build is matching the behaviour on Mac/Linux so actually changing line 50 of `compileMusic.js` to just
```js
const outputFile = "music.c";
```
seems to fix for me.

I'm not too sure on the specific change you made to confirm that's working but it definitely still creates a ROM with working music.

You can confirm the output of your version of mod2gbt by running
```bash
mod2gbt.exe effects_test.mod effects_test -c 4
```
which should make a file called `music.c`
vs running the old binary will make `output.c`

Hope this helps

RichardULZ:
THANK YOU!
That explains a lot, I'll create an example mod song that showcases the new effect, and attempt these changes to get the output working.

As this version of mod2gbt is not up to date with the legacy branch
https://github.com/AntonioND/gbt-player/blob/master/legacy_gbdk/mod2gbt/mod2gbt.c

Would it be worth updating? I think the input arguments are different now.

chrismaltby:
Glad to have helped :-)

Might be worth giving the latest legacy version a go, not sure exactly what the differences are but from a quick look at the codebase it seems the generated file would be `output.c` with this version
https://github.com/AntonioND/gbt-player/blob/3570ae21646796872a2238c4b9777b0c476bdf19/legacy_gbdk/mod2gbt/mod2gbt.c#L202
so the change to `compileMusic.js:L50` would instead need to be 
```js
const outputFile = "output.c";
```
if we compile for all platforms.

And as you say the args have changed
https://github.com/AntonioND/gbt-player/blob/3570ae21646796872a2238c4b9777b0c476bdf19/legacy_gbdk/mod2gbt/mod2gbt.c#L885-L886

so looks like the call would now be

```
mod2gbt.exe modfile.mod label_name bank_number
```

seems like the `-c` flag has just been removed so changing

https://github.com/chrismaltby/gb-studio/blob/8366701e7ac7c0642b88cefb549a0a69876dc2e8/src/lib/compiler/compileMusic.js#L51

to 

```
const args = [`"${modPath}"`, track.dataName, track.bank];
```

might be all that's needed, assuming the output format is compatible with the previous version.

I can probably compile the Mac / Linux versions if you need.

RichardULZ:
I've got it compiling and running thanks to your help!

The new mod2gbt was also naming stuff `constmusic_4c4946b80_Data` instead of `music_4c4946b80_Data`, just missing the space after `const ` in mod2gbt i think. 

Now that's over, it's just my bugs.
Found i was saving Volume sweep as effect 5 instead of 3, sweet, got that out of the way.

And then a more complex bug,
Now that the envelope is working, setting the Trigger register NR14 will start playing a note again.
but... trigger is set every time something changes, Panning, volume, envelopes, not just notes.
https://github.com/chrismaltby/gb-studio/blob/3ef388f40e67f12e5b6c0dac2c4aa538cb93670b/appData/src/gb/src/gbt_player_bank1.s#L230-L244

Every time something changes, it refreshes all the registers and triggers a note to play. I now need to stop that, i don't think it's supposed to do that anyway. Making steady progress.

Will compare old songs to builds in 1.1 to verify they sound the same, It's always had strange quirks that seemed like bugs, so this may improve that too. (eg: Set panning without a note at the start of a song makes a tone, not expected.)

RichardULZ:
May need to revert the Double speed CPU commit i accidentally applied early on.

Will make test projects with 1.1 and 1.2 files to confirm nothing has broken, Likely to use music from the community assets folder. https://github.com/DeerTears/GB-Studio-Community-Assets

PR #296, opened 2019-09-26
Add Custom Events functionality

pau-tomas:
**Uploading it to allow some early testing of the feature.** 

This is a bit of a crazy idea that I wanted to test. 

It allows the user to create events that are a combination of other events and can be then used in any script of the project as any other events.

Procedures/Custom Events have “local” variables and actors. When using the event on another script, those are the input parameters.

The “Add Procedure” button appears at the bottom of the World Editor and opens the Procedure Editor. Already existing procedures appear listed on top of the button.

image

Each procedure has an editable name, a list of input parameters (up to 10 variables and 10 actors) and a script editor like the one present in Scenes, Actors and Triggers. 

The parameter list is automatically populated based on the variables and actors used in the script. The parameters can be renamed to help understand what they do.

Once created, the event be added to any script using the “Add Event” and searching for it. 

image

The event appears as any other event and it allows to populate the parameters with either actors or variables.

**Known issues**

- The feature is called Procedures all across the app. It should be changed to Custom Events but open to suggestions.
- Is it possible to add the `EVENT_CALL_PROCEDURE` event, but it should be hidden as it's just the base Custom Event.
- Is it possible to add a Custom Event inside itself, creating an infinite loop 😨 

----

New feature

N/A

Add functionality that allows the users to create Events at project level and use them in any script.

Projects should keep working as usual. No changes required on the user side.

* **Other information**:


proximitysound:
Screen Shot 2019-09-27 at 9 42 28 AM
This error occurs under two scenarios so far:
Create New Project
Add Scene
Add Procedure

or

Create New Project
Add Procedure
Add Event > Text Event (Don't add anything to the text window)
Add Scene

pau-tomas:
Thanks @proximitysound. That bug should be fixed with the latest commits.

chrismaltby:
This is really amazing! I'd been trying to solve the same problem with a plugin system (more on that later) but that required coding up custom events as JS files similar to those in `/src/lib/events`. I like how this is really easy to use and works very similarly to the existing scripting.

From my quick tests I found a few minor issues but other than that worked great!

1. If I made a procedure and added an actor event, then selected "Actor C", opening the list again showed "Actor C" where "Actor A" should be. It seems to be something to do with the custom names for actors I think but not had time to look into it fully.
Screen Shot 2019-09-29 at 12 22 23

2. As you said, EVENT_CALL_PROCEDURE needs hiding from the event search. You could add it to the deprecated events list at:

https://github.com/chrismaltby/gb-studio/blob/8b4197ec8297136c3c58a8827bc5e4f8188ecfb8/src/lib/compiler/eventTypes.js#L113-L126

Maybe `EventsDeprecated` needs to be renamed everywhere to `EventsHidden` instead though if you add it there.

3. If you want to prevent custom event nesting you could look at EventsOnlyForActors which currently only stops you from adding a Push event to a trigger or scene load (although I believe you can get around it with Copy/Paste so it's not perfect). EventsNotAllowedInCustomEvents may be a bit messy though, maybe it just needs some logic around here:

https://github.com/chrismaltby/gb-studio/blob/8b4197ec8297136c3c58a8827bc5e4f8188ecfb8/src/components/script/AddCommandButton.js#L121-L126

Though if we want to prevent copy/paste working too probably need to pull the logic for what events can/can't be used where out of that React component entirely into a shared function at some point.

In terms of a name for the feature, I think "Custom Events" is good. 👍 Maybe when you show them on the events list they should be prefixed with `Custom Event` e.g. "*Custom Event:* My event" just so they're grouped together.

Oh and a bit off topic but I mentioned plugins earlier, here's an example showing how they work
[pluginsTest.zip](https://github.com/chrismaltby/gb-studio/files/3668009/pluginsTest.zip)

If you unzip that into the root of a project (you might need to reopen it after, though I was working on live reload) it will add some extra assets, a dance command, and a menu item to make a pushable rock.

Screen Shot 2019-09-29 at 13 02 35

I'd been thinking about making an official repo for plugins and adding an equivalent of the Unity Asset Store to browse/download them. Might have to push it back to 1.3.0 though since there's still a lot to do there.

pau-tomas:
Thanks for checking Chris, glad that you like it!

I think 1 is happening because the select component is getting confused with the index of the actor and replacing the first one, I'll take a look. 

Using `EventsDeprecated` makes sense for the `EVENT_CALL_PROCEDURE`, I'll rename it.

For the event nesting the main problem is to avoid the infinite loop on compile. Since what I'm effectively doing is to add all the Events from the procedure into the `output` array the compile process doesn't know were to stop (rightly so). This means that a Custom Event can't be nested anywhere in the Script tree (so `Custom Event 1` > `Custom Event 2` > `Custom Event 1` also ends in infinite loop). 

The naivest solution is to not allow *any* Custom Event to be added into another Custom Event which should be easy to implement. It limits the functionality but I think would suffice for v1.

A better solution, but more complex and not sure how to do it yet, would be for a Custom Event to be compiled only once and have a pointer to it that can be used anytime is called. That means that in the final assembly code there will need to be pre and post call statements to copy the variables and actors into local variables and then those local variables (that potentially have been modified) copied back to the global variables before returning to the main program. (I hope this makes sens).


About plugins, I stumbled upon the feature while checking the codebase (and even toyed a bit with it for another experiment I'm testing for animated tiles :) ). I think is a very powerful feature and complimentary to Custom Events similar to libraries vs functions: Plugins seem more advanced and easy to be shared among the community and Custom Events are organizational structures for a project.

pau-tomas:
I think the feature is ready to be reviewed now. Fixed the known bugs and renamed it to "Custom Events" across the UI strings (didn't changed the code references to `procedure` but happy to do so).

chrismaltby:
Nice one thanks @pau-tomas. I'm on holiday this week but I'm going to review this and the other pending requests when I get back!

chrismaltby:
Just fixed the merge conflicts brought in by some of the recent merges.

It works great with the new sound effect feature, you can make little jingles maybe to play when puzzles are complete that just become a single event!

There's a few bugs that I'm looking into though, if you add an avatar to any text in a custom event it crashes when trying to display that text, also if you add music that wasn't used anywhere else if the game it doesn't end up compiling it. 

It's basically because `compileData.js` uses `walkEvents` to find all the avatars and music used to only build the data that is actually used, and the way you're storing the script doesn't allow it to be read by walkEvents. I think I just need to move the script data to be stored in `children` rather than `args` but then that introduces some issues in the UI, will have a look over the next few days to fix this.

```
{
  "id": "3af93716-ed5d-4ac3-99ae-c214c8f35275",
  "command": "EVENT_CALL_PROCEDURE",
  "args": {
    "procedure": "4b668b45-357c-4b34-accb-75ef02f881c8",
    "__name": "Custom Event: Test"                                
  },
  "children": {
    "script": [
      {
        "id": "11b65b5c-03ad-400a-bc34-2e9caaba3ffa",
        "command": "EVENT_MUSIC_PLAY",
        "args": {
          "musicId": "f50428ab-a084-4591-9bba-2ba10fe7b1c6",
          "loop": true
        }
      },
      ...
```

pau-tomas:
Ah, that's true! missed that case, sorry. 

Another solution that could work if the UI issues are too much would be to check `args.script` in the `walkEvents` methods. But that might not scale if the `children` array is used in other places in the future.

Let me know if you need any help.

chrismaltby:
Hey @pau-tomas no worries!

I've pushed an update that seems to fix those issues now. You're right that I could probably have just updated `walkEvents` to look for `args.script` but I'm trying to get all the events that have nested events using `children` going forwards as all the functions in `helpers/eventsSystem.js` know to look for it and can handle recursively finding/mapping/updating etc.

I found another issue too in `updateEntitiesProcedureScript` where it would update events already in use but only if they were at the top level of the script array, so if any events were nested in conditionals/loops/groups they didn't get updated. I had a function `mapEvents` which recursively goes through the events so I've switched it out for that. While I was at it I added `updateEntitiesProcedureName` so that any renames propagate through to events that have already been created too.

Still need to do quite a bit more testing, but it's looking pretty good for me at the moment.

pau-tomas:
Changes look good. Thanks! 

One thing I noticed is that when the Custom Event doesn't have parameters it looks bad in the Script list. I'm going to add an optional description field that, if present, will be rendered like the text at the top of the `Restore First From Stack` Event.

chrismaltby:
Two more thoughts of things we could do on this feature:

1. Add ability to get to the Edit Custom Event section from an instance of the event.

Screenshot 2019-10-17 at 21 13 59

2. Add a message to instances when the custom event has been deleted (currently they still work when you compile but you can't get to the fields anymore).

Screenshot 2019-10-17 at 21 15 53

I was wondering if there's enough information stored in the single instance to restore the deleted custom event with the same ids even, maybe not.

I could have a look at these at some point but just wanted to see what you think.

pau-tomas:
I thought about adding "Edit Custom Event" but then I forgot. Should be an easy change.

About the Restore option, not sure if there's enough information to recover it, but maybe something can be done using the Redux history and look for the most recent version of the stored procedures before it was removed?

chrismaltby:
Hey @pau-tomas I've been working on this over the last few days and pushed a few more commits.
- Added the dropdown to "Edit Custom Script" from an instance
- Description field is there now working as you described
- Added a middleware for REMOVE_PROCEDURE that first checks if there are any instances and warns you where the event was used giving you a chance to cancel. If you do remove the custom event it also removes all instances. If there was no uses of the event it just deletes with no warning as before.

Really close to being ready now, think I'll probably just go through and switch all references to Procedure with Custom Event just so it's not confusing later

pau-tomas:
I really like those changes, the confirmation dialogue (and removing the instances) is very useful. Excited to have this almost ready. Thanks!

(I pushed a really small change to remove a couple of lines and variables that weren't used when rendering the Event instances)

chrismaltby:
Finally merging this in!

I've renamed all references to procedures to be customEvents so if you have any existing projects they may need migrating by changing the following in your `.gbsproj` file:
`EVENT_CALL_PRODEDURE` becomes `EVENT_CALL_CUSTOM_EVENT`
`args.procedure` becomes `args.customEventId` (note I added 'Id' at the end)
`procedures` becomes `customEvents`
Think that should be all you need to do.

Great work on this feature, I think people should find it really useful!

Issue #177, opened 2019-05-17
Music will not play in-game until a Joypad Input is pressed


timsamoff:
**Note: for support questions, please use one of the following channels**.

Reddit: [/r/gbstudio](https://www.reddit.com/r/gbstudio)
Discord: [Join Chat](https://discord.gg/bxerKnc)

This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**

Bug.

* **What is the current behavior?**

Can't get music to play in a Scene until a Joypad Input is pressed. (Or, at least, that's what my tests have shown.)

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

This can be tested with the GB Studio demo files...

1. Add a Music: Play Track to the Title Screen Scene... Test the game and it won't work.
2. DO the same thing on the Outside Scene and it will.

* **What is the expected behavior?**

Music will play whenever called.

* **What is the motivation / use case for changing the behavior?**

To provide music on Scenes when the game loads.

* **Please tell us about your environment:**

  - Version: 1.0.0
  - Platform: macOS

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

timsamoff:
The music will begin on the Title even if you press a Joypad button that's not being listened for.

timsamoff:
This is confirmed as a Web-only issue -- works fine when emulated.

RichardULZ:
This sounds like an issue with general web features that prevent sounds from auto playing unless interacted with, not really an issue with the html5 emulator.
Try uploading to Itch.io, where they have a built in "Click to play" feature on html games that meets these conditions. 
If this resolves your issue, please close it.

timsamoff:
Is it's "Click to Play" functionality different from mine?

http://samoff.com/CATG/

If so, I'll test it, but in my experience, clicking a button on a webpage is not within the same "domain" interacting within the "app."

In any case, it's a pretty bad UX issue, since 4 of my playtesters (who knew there was music) didn't do anything on the titlescreen for several seconds because they thought the game was still loading.

Is there a way to implement an "invisible click" function or some other form of game-instigated interaction that does not require user interactions to kick the music into gear?

[...]

PR #298, opened 2019-10-02
Fix JS GameBoy emulator audio

RichardULZ:

Bug fix

Audio was mono, had major crackle on chrome android, could cause clicking if in the background on some devices.
Did not play until a button was pressed. #177

Audio is stereo again, and play's better on all devices (optimized, and buffer increased to 4096). 
Sound starts as soon as emulation begins now. 
Overall sound volume is the same as 1.1 (Half volume).

Not likely!

* **Other information**:
Pop fix being done in the gameboy core let me remove the code blob from last dc offset pop fix, which was performance intensive and had channels limited to 1.
Wanted this separate from GBT expansion to get this merged quickly as it improves all web outputs.

### DIFF
--- a/appData/js-emulator/js/GameBoyCore.js
+++ b/appData/js-emulator/js/GameBoyCore.js
@@ -5084,7 +5084,7 @@ GameBoyCore.prototype.GyroEvent = function (x, y) {
 GameBoyCore.prototype.initSound = function () {
 	console.log("INT SOUND")
 	this.audioResamplerFirstPassFactor = Math.max(Math.min(Math.floor(this.clocksPerSecond / 44100), Math.floor(0xFFFF / 0x1E0)), 1);
-	this.downSampleInputDivider = 1 / (this.audioResamplerFirstPassFactor * 0xF0);
+	this.downSampleInputDivider = 0.5 / (this.audioResamplerFirstPassFactor * 0xF0); // Reduced to 0.5 from 1 to half volume.
 	if (settings[0]) {
 		this.audioHandle = new XAudioServer(2, this.clocksPerSecond / this.audioResamplerFirstPassFactor, 0, Math.max(this.baseCPUCyclesPerIteration * settings[8] / this.audioResamplerFirstPassFactor, 8192) << 1, null, settings[3], function () {
 			settings[0] = false;
@@ -5256,8 +5256,8 @@ GameBoyCore.prototype.initializeAudioStartState = function () {
 	this.noiseSampleTable = this.LSFR15Table;
 }
 GameBoyCore.prototype.outputAudio = function () {
-	this.audioBuffer[this.audioDestinationPosition++] = (this.downsampleInput >>> 16) * this.downSampleInputDivider - 1;
-	this.audioBuffer[this.audioDestinationPosition++] = (this.downsampleInput & 0xFFFF) * this.downSampleInputDivider - 1;
+	this.audioBuffer[this.audioDestinationPosition++] = (this.downsampleInput >>> 16) * this.downSampleInputDivider;
+	this.audioBuffer[this.audioDestinationPosition++] = (this.downsampleInput & 0xFFFF) * this.downSampleInputDivider; //had - 1, removing fixes pop
 	if (this.audioDestinationPosition == this.numSamplesTotal) {
 		this.audioHandle.writeAudioNoCallback(this.audioBuffer);
 		this.audioDestinationPosition = 0;
--- a/appData/js-emulator/js/GameBoyIO.js
+++ b/appData/js-emulator/js/GameBoyIO.js
@@ -2,7 +2,7 @@
 var gameboy = null;						//GameBoyCore object.
 var gbRunInterval = null;				//GameBoyCore Timer
 var settings = [						//Some settings.
-	window.location.href.indexOf("audio=true") > -1, //Turn on sound.
+	true,								//Turn on sound. Removed //window.location.href.indexOf("audio=true") > -1,
 	true,								//Boot with boot ROM first?
 	false,								//Give priority to GameBoy mode
 	1,									//Volume level set.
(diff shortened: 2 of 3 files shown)


chrismaltby:
Hi @RichardULZ Unfortunately this update breaks at least desktop Chrome and Safari support when creating the web build and hosting from a web server as it breaks the Web Audio autoplay policy by creating a web audio context before a user interaction. 

https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#webaudio

The issue I think was the replacing of this line with a `true` value.

```
window.location.href.indexOf("audio=true") > -1
```

which causes the audio to be enabled on page load before any user interaction has taken place.

Previously it only did this when audio=true was set in the query string which the In-App Emulator uses (because Electron allows autoplay audio)
https://github.com/chrismaltby/gb-studio/blob/202685ad5efc3a2bedac7b494419b0ff51261a6a/src/index.js#L211

and then in the exported web build this function is called on any touch/keyboard input to initialise the audio if it wasn't already enabled.

https://github.com/chrismaltby/gb-studio/blob/202685ad5efc3a2bedac7b494419b0ff51261a6a/appData/js-emulator/js/other/mobile.js#L72-L96

based on the google link it seems it should be possible to maybe have the context created on page load and call resume() on the first interaction for browsers that don't allow autoplaying audio. If you're saying that the Android browser is allowing it then this might help there, I tried and couldn't get it to work though...

I'm not entirely sure there is a work around to fixing #177 everywhere without maybe having something like a splash screen which I'd like to avoid.

It might be worth splitting out the stereo audio fix in a separate pull request as that seems to be working from what I can tell :-)

RichardULZ:
Ahh, it may be an itch.io autoplay audio fix that's kept it working for me. 
https://richardulz.itch.io/gbstudio-test#1
Will have to test with a plain web server and see if it breaks, but i can definitely change that back for now, the stereo fix is important enough.

RichardULZ:
[FixedPop soundTrue.zip](https://github.com/chrismaltby/gb-studio/files/3732111/FixedPop.soundTrue.zip)
Here's the file used in that itch.io example.

chrismaltby:
Hey @RichardULZ thanks very much for this, I've had chance to have a better look and with commit dd728e98ced87096c7cc8e27b42e7acf2375e7b5 I think I've got autoplay audio in a much better place.

I found you can detect if the audio context was able to start by checking for `window.audioContext.state === "running"` and using that can determine if you're playing from itch.io where, like you say, clicking the start button was enough to allow sound to start.

So as of that commit if you upload to itch.io it should autoplay audio, and if you self host it will continue as it did previously by starting audio on first user input, I also added an `initSound()` call on mouse click too which should help also.

On a complete aside the music in your example is awesome, not sure if you wrote it yourself but if you did and ever wanted to contribute a few extra tracks to the starter project it would be much appreciated :-D

RichardULZ:
Perfect @chrismaltby ! Will test that code out sometime. 
Was going to suggest the initSound on click as a hold out solution but you've gone ahead and done that already too!

Surprised you got your dummy `audio context` buffer to check if it was running, I think the resume function had to be called on `xaudio's audiocontext`, as the dummy one you create on `initSound` is not for playing gameboy audio.

On the aside, That music is me trying to transcribe StarLove by +tek, only got the first 26 seconds...
https://www.youtube.com/watch?v=KgE-4IpSi7c

RichardULZ:
Hmm, Bad problem, Firefox audio is broken. Will investigate.

RichardULZ:
Emergency fix! @chrismaltby 
https://richardulz.itch.io/gbstudio-test?password=1

[FirefoxFix_1xSpeed.zip](https://github.com/chrismaltby/gb-studio/files/3747482/FirefoxFix_1xSpeed.zip)
Use the mobile.js file in this zip, i've reworked some things.

Tested on **Windows** and **2 Android** devices with both **Firefox & Chrome**, in both a **Local web & itch.io**
Androids were **Oneplus3T**, and an Old **Huawei Y330** (dual core 1.3ghz, 512mb ram).

**Firefox** will not autoplay but will play on click or button press (was broken).
**Chrome** will autoplay if engagement was reached (click to start), Buttons will still start the audio if that has not happened.

**Removed** most of the `dummy audio buffer` as it didn't seem necessary.
The old android struggled with **all** cases, but Firefox worked better than chrome, audio was ok when not moving ;P

`setting[0]` is checked again from suspicion we were running `initSound` every keypress (the slow old android was lagging worse than normal without that check).

chrismaltby:
Hi @RichardULZ Thanks for this, I've looked into the issues in Firefox and it looks like the audio context always started with the state being `suspended` rather than `running` even if user input had occurred. It looks like calling `resume()` on the context and waiting for it to start fixes the issue in Firefox and doesn't break Chrome or Safari. Pushed a fix with in 17195e7d19aa056851fc242b85c8e479b74d953d 

I've not got any access to any Android devices today but I'm going to try out a few from the QA department at work next week and see if there's anything in these checks that we can improve further.

RichardULZ:
@chrismaltby Your latest patch in 17195e7 is working fine on both androids, in chrome and firefox, and... audio will autoplay in Firefox now too! Great job!

I'm still not sure why you have these lines, as everything works perfectly without them...
https://github.com/chrismaltby/gb-studio/blob/17195e7d19aa056851fc242b85c8e479b74d953d/appData/js-emulator/js/other/mobile.js#L77-L90

The **android** devices are working fine with the patch, The **old low spec** one is still struggling as expected. Firefox is a lot faster than chrome here, and is very playable, moving around you get audio stutters at the start, but it smooths out when walking in large areas.
Seems like tapping buttons causes the slow down, not the emulation, as walking larger distances has less stutter, something to do with the button code. It's fine on mid-high end phones though.

In even better news, on this old phone, I've tested a rom compiled as both a normal Gameboy Game, and as GBC with double speed mode, the performance is identical, so my fears of needing to leave it off for performance are moot.

As usual, i've updated the rom hosted at https://richardulz.itch.io/gbstudio-test?password=1
It's running a stress test with music and gbc 2x speed.
If you walk in the direction of the arrows it will rapidly switch the scene, great test for music performance.

chrismaltby:
Hi @RichardULZ That's great news and I'll be getting to the 2x request soon :-)

That extra code actually came from here https://gist.github.com/kus/3f01d60569eeadefe3a1 as a fix for audio not working for me on iOS, it's possible it's not needed anymore now though with the recent changes so I'll try removing and see if that breaks anything.

PR #305, opened 2019-10-17
JS Game Boy Emulator: Add gamepad support

bbbbbr:

Feature

Absence of gamepad support.

Adds gamepad support, activates if gamepad button is pressed with js Game Boy emulator has focus.

No changes should be required.

* **Other information**:
When w3c "standard" mapping is available for the gamepad, buttons will map to the expected locations Otherwise the first 4 buttons map to A/B/Select/Start.

Gamepad support co-exists and does not interfere with keyboard and on-screen controls. Polling only starts if a button is pressed on gamepad, and will stop if the gamepad disconnects.

Sound will also start when the gamepad polling is started, similar to keyboard input.

### DIFF
--- a/appData/js-emulator/js/other/controls.js
+++ b/appData/js-emulator/js/other/controls.js
@@ -33,6 +33,51 @@ var btnStart = document.getElementById("controller_start");
 var btnSelect = document.getElementById("controller_select");
 var dpad = document.getElementById("controller_dpad");
 
+
+// HTML Gamepad API support
+// Poll for gamepad input about ~4 times per gameboy frame (~240 times second)
+const GAMEPAD_POLLING_INTERVAL = 1000 / 60 / 4;
+const GAMEPAD_KEYMAP_STANDARD_STR = "standard"
+
+// When gamepad.mapping reports "standard"
+const GAMEPAD_KEYMAP_STANDARD = [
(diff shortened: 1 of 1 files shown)


chrismaltby:
This is great! Works perfectly with my PS4 controllers on a Mac

chrismaltby:
Actually there might be an issue with the enabling of sound when exported for Web and running the game from a web server. I get the "audio context was not allowed to start" warning

Screenshot 2019-10-19 at 14 51 29

I think `window.addEventListener("gamepadconnected", (event) => {` doesn't class as a user gesture in Chrome so it blocks the audio context from starting, there's supposed to be a way to resume it later but I've never been able to get it to work so once this error is seen there's currently no way to get audio to start without a page refresh. I had a look around and I couldn't find any gamepad events that did allow web audio contexts to be created which is a bit odd, so for now I've removed the initSound line.

I've added an initSound() in response to page clicks though, which is allowed and might help a little. I might have just missed the way to do this from a gamepad event though.

chrismaltby:
Actually I've figured out a way to make it work if uploaded to Itch.io without breaking the functionality if just running from a web server browser directly (just falls back to keyboard/touch/click input to start) in commit dd728e98ced87096c7cc8e27b42e7acf2375e7b5. I've added your `initSound()` call back in.

bbbbbr:
Thanks!

Issue #158, opened 2019-05-14
[Request] Multiple choice with more than 2 choices


sittingrisk:
**Note: for support questions, please use one of the following channels**.

Reddit: [/r/gbstudio](https://www.reddit.com/r/gbstudio)
Discord: [Join Chat](https://discord.gg/bxerKnc)

This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**

It would be really useful to have a multiple choice tool that allows for more than two options. As it is now multiple choice sets a variable to true if option A is chosen and false if option B is chosen. Maybe if there are more choices than true then each choice could set a different variable to true and a variable for all other choices to false.

chrismaltby:
I'd been thinking about this one for a while, and will probably look at this after the next release. I'm thinking rather than setting multiple variables like you suggest it would just set the value of a variable to a different number based on the option chosen (so variable would be "0" if the first option is chosen, "1" if the second option, "2" if the third option etc. I think you should also be able to have more options than fit into the box, in which case it will show pagination on the bottom right. Finally I think it's a good idea to have an alternate view where the menu appears on the bottom right of the screen which will allow people to do menus which has been requested a lot.

Rough mock of UI below:
<img width="757" alt="Screen Shot 2019-05-14 at 9 41 37 AM" src="https://user-images.githubusercontent.com/16776042/57683867-8136c680-762c-11e9-8e91-43557c1d6348.png">

KitF:
Would it be possible to have "B" cancel out of a multiple choice? Or would we always need to include a "Cancel" option?

thomas-alrek:
In most GB games the convention when pressing B in a multiple choice dialog is to select the "negative" option. So if the options are "Yes" and "No", B would usually select "No", not cancel it

chrismaltby:
I think making it work like the following should be okay, and also allow the event to be backwards compatible with everyone who's used it previously (this is going to be quite important but also difficult as features are added/modified going forwards)
<img width="896" alt="Screen Shot 2019-05-15 at 10 23 58 AM" src="https://user-images.githubusercontent.com/16776042/57764566-9c243c00-76fb-11e9-8e97-47f4bc241eea.png">
[...]

PR #306, opened 2019-10-18
Add support for menus with more than 2 choices

pau-tomas:
**TODO** 
- [x] Code cleanup (`UI.c` is a bit messy now 😅 )
- [x] Tests
- [x] Migration for Events with conditional fields for 2bb6c79 

----

New feature, requested in #158 

Multiple Choice is limited to two options.

New `Text: Display Menu` Event:
- Supports up to 8 text options, 6 characters per option.
- Option to set the last menu option as "Cancel" item (sets variable to 0).
- Supports two layouts: "Dialogue" for horizontal and "Menu" for vertical.
- Support for two columns of items in "Dialogue" layout.
- Rewritten part of the Dialogue rendering code. `Text: Display Multiple Choice` uses the new code too.

I kept the `Text: Display Multiple Choice` as it provides a feature that the `Menu` can't: up to 18 characters per option. 

Ideally not. The `Text: Display Multiple Choice` Event is now using the new code for flexible menus so it should transition.

* **Other information**:

I cherry-picked 2bb6c79 from one of the develop branches, to be able to support a dynamic list of options. I think it'll require to add a migration step for Events using the old conditional format.

Some screenshots:

image Screenshot 2019-10-18 at 02 02 56 Screenshot 2019-10-18 at 02 03 01 image


chrismaltby:
You legend!! Lots of people are going to be very happy with this one :-D

As part of the branch you cherry picked from I had a work in progress `Case` event which I'll have to finish up because it would go very nicely with this event.

Looks great from a really quick review, only two things I'd suggest, currently if you have "Use cancel as option" unchecked then the you can't pick "0" anymore in the list, **except** you can if you press the "B" button. You followed my design almost exactly I just hadn't though about this :-D

I think we could solve the problem by adding just one more checkbox

[ ] Allow cancel using 'B' (disable/enable the ability to cancel using the B button)
[ ] Last option is cancel (works identically to your current checkbox)

Not sure on the exact wording to use but that seems to fit in English at least.

e.g.
Screenshot 2019-10-18 at 13 30 11

and:
Screenshot 2019-10-18 at 13 30 56

There was a few people a while back wanting the `Multiple Choice` event but without the "B" button functionality, so I think they could use this instead to get the same thing.

I haven't decided yet but potentially we can use this event to replace `Multiple Choice` entirely in 1.2.0 and write a migration to move all events using the old event over to this one like I did for the old Math events into the new combined one.
 
Great work!

pau-tomas:
The problem with migrating `Multiple Choice` to `Menu` is the character limit (`Menu` is up to 6, while `Choice` supports 17). One suggestion I got on discord would be to distribute the 54 characters equally between options (with a upper cap in 16) so:

- 2 options = 17 characters  
- 3 options = 17 characters
- 4 options = 13 characters 
- 5 options =  10 characters 
- 6 options =  8 characters 
- 7 options =  7 characters 
- 8 options =  6 characters 

I'm thinking about it as it might be confusing for the user. One way to mitigate confusion could be adding a character count to text inputs with a `max` field. 

I added the `Cancel on B` option (on by default). Not convinced about the wording either, but I like the idea of the `Last option is cancel` being next to the number of options field.
 
image

chrismaltby:
I'm thinking we forget what I said about `Multiple Choice` and keep them as two separate events. Like you say the character limits would be tricky and potentially confusing.

I think it's not a problem to have a simpler and more limited event while also having a more full featured one for when you need it, it's the same reason that `If Variable True` and `If Variable False` exist when technically `If Variable Compare With Value` can do the work of both.

On the wording it probably just needs "Last option sets to '0'" just for consistency as I think your change to the 'B' button label makes sense.

Screenshot 2019-10-19 at 17 04 20

It's all looking great though!

I'll have a think if there's any good way to dynamically set the character counts based on the number of options though but maybe it doesn't need it for now.

pau-tomas:
In the pasts days I clean up the `UI.c` file a bit (mostly added some functions for duplicated code paths), added tests for the new Event and implemented the migration for the field visibility conditions. So, I think now is ready for review.

chrismaltby:
Hey @pau-tomas  just starting to review this now.

Fixed a small issue where the menu items appeared out of order if any menu items didn't have labels set.

e.g. with the following input:
Screenshot 2019-11-01 at 14 13 02

it was giving:
Screenshot 2019-11-01 at 14 12 33

I've made empty items use the placeholder text instead to match how the multiple choice event works so with the same input it now gives:
Screenshot 2019-11-01 at 14 11 54

Will keep reviewing but looking good other than that so far

Issue #314, opened 2019-11-05
Readability improvements


DuendeInexistente:
**Is your feature request related to a problem? Please describe.**
Events, especially when folded, tend to mix together when glancing at them, making them difficult to read quickly. 

**Describe the solution you'd like**
A) Currently, events have different but uniform colors depending on their folding level. I believe they should have an alternating light-dark pattern to make them easier to tell apart.
B) Empty 'Else' folds should disappear altogether if they are empty and both halves of an If event are folded.
C) If Value events could also have (Small) text indicating their conditions when folded if they don't have custom titles. IE, "If ChangeChar equals 5" or "If ChangeChar=5". This could replace the default title they already have too (If Variable compare with Value)

chrismaltby:
Hi @DuendeInexistente Thanks for this, I agree.

I've started working on making these improvements in PR #318

PR #318, opened 2019-11-08
Feature/usability improvements

chrismaltby:

Implements various usability improvements based on the the suggestions by @DuendeInexistente in #314.

* **What is the new behaviour (if this is a feature change)?**

1. Script events now have alternating light and dark backgrounds to make it easier to see where events start and end

Screenshot 2019-11-08 at 17 59 56

2. It is now possible to remove the Else block on any conditional event using the menu. If any events are included in the else block already they will not be called. Collapsing a conditional block also hides the Else block (and any other branches, e.g. in Switch event)

Screenshot 2019-11-08 at 18 01 33

@DuendeInexistente, your third point around adding the custom titles for conditional statements is a bit more complicated and I haven't implemented it here. I've got an idea that may work but for now a work around is to use the "Rename Event" option.

While making these changes I also added a few more usability improvements

3. If you are in the "Add Actor/Trigger/Scene" toolbar mode pressing Escape will now cancel and return to the "Select" mode

4. Holding down Alt will turn all "Add Event" buttons into "Paste Event" buttons making it much easier to paste events into conditional paths (previously you would need to paste outside of condition and drag in)

Screenshot 2019-11-08 at 18 25 59

5. The comment event added by @ManuGamesDev now sets the event title to match the content and disabled events match the same commented style

Screenshot 2019-11-08 at 18 29 21

Potentially, as it's built on top of #316 so the same issues apply.


src/components/editors/CustomEventNavigation.js
```diff
@@ -21,6 +22,9 @@ class CustomEventNavigation extends Component {
                     selectCustomEvent(customEvent.id);
                   }}
                 >
+                  <div className="EditorSidebar__Icon">
```
pau-tomas: 😍

src/components/script/ScriptEventBlock.js
```diff
@@ -324,69 +326,71 @@ class ScriptEventField extends Component {
...
-        args={args}
-        onChange={this.onChange}
-      />
-    );
+    let label = field.label;
+    if (label && label.replace) {
+      label = label.replace(/\$\$([^$]*)\$\$/g, (match, key) => args[key] || 0);
+    }
 
-    if (field.toggleLabel) {
```
pau-tomas: I think by removing this condition the Avatar field is always visible in Text Dialogue events.

src/components/script/ScriptEditor.js
```diff
@@ -138,8 +138,10 @@ class ActionMini extends Component {
 
   onPasteEvent = before => e => {
     const { id, onPaste } = this.props;
-    const { clipboardEvent } = this.state;
-    onPaste(id, clipboardEvent, before);
+    const clipboardEvent = this.readClipboard();
+    if (clipboardEvent) {
```
pau-tomas: The app crashes when pasting a Custom Event Script via the new `Paste Event` or `Paste Event Before/After`. I think it was crashing before, but changing this line to 

```if (clipboardEvent && clipboardEvent.__type === "event")``` 

should fix it.

PS: I don't think I can add changes in this PR. If you prefer I can do a separate one with this change.

chrismaltby: Oooh, good spot! So actually this wasn't an issue with custom events but with copying any entire script and pasting it where you could paste a single event, this did actually work until my refactor in the last few days. I think the change you suggested would stop the errors but prevent you from being able to paste a full script.

Instead I made this change 5d157ba07cf207c9969ddde6081f954329d8b87a line 177 is the important one, when a full script is in the clipboard you need to read the array of events out the of the `script` field before pasting.

Thanks for your help!

src/components/script/ScriptEventBlock.js
chrismaltby: Ah, oops. This was quite a tricky merge, I knew I'd miss something! Fixed in f7a8232ba7c9f2a4bedfb80a8744b9adec708aad thanks

PR #325, opened 2019-11-12
Add "Self" to Actor select

pau-tomas:

Adds a new `Self` option to Actor Selector fields inside Actor Scripts

When copy and pasting an Actor with scripts that reference itself all the Actor select fields need to be set to the new Actor again.

When selected, "Self" always references the current actor. So when the actor is copied it'll reference the newly created Actor rather than the previous one.

No

* **Other information**:

### DIFF
--- a/src/components/forms/SceneActorSelect.js
+++ b/src/components/forms/SceneActorSelect.js
@@ -22,9 +22,15 @@ class SceneActorSelect extends Component {
     };
   };
 
+  actorName = actor => {
+    const { actors } = this.props;
+    return actor.name || `Actor ${actors.indexOf(actor) + 1}`;
+  }
+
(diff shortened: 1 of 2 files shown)


chrismaltby:
Great idea! It looks like the UI was hooked up but the compile part was missing, I've just pushed a small update that appears to make this work.

I'm gathering a bunch of updates together at the moment and making sure they all work together so I think once I've tested this some more it will be merged as part of the 1.2.0 release candidate.

Thanks @pau-tomas!

pau-tomas:
Oh! Yes, I did a last minute refactor to not have to pass context all the way down the view hierarchy and then I forgot to add the compile part. Thank you!

PR #327, opened 2019-11-15
Cleanup Custom Event instance args after editing

pau-tomas:
Fixes bug where Custom Event variables where still rendered in dialogue even after removing them from the parameter list.

Bug fix

- Create a new Custom Event, add any Event referencing a variable (for example `Variable: Decrement by 1`) and then a `Dialogue` Event referencing that variable.
- Add the Custom Event to a Script and set the parameter variable to a variable with a value different than 0.
- Edit the Custom Event and remove the `Variable: Decrement by 1` Event, but keep the dialogue one.
- Run the project. When the Script is executed the dialogue will still render the variable value.

- Following the above instructions. The dialogue will display 0, as there's no variable passed as parameter.

It shouldn't.

* **Other information**:

An alternative, probably harder to implement but maybe more correct. Would be that if a variable is referenced in a dialogue inside a Custom Event (`$VX$`) it should be added to the parameter list.

### DIFF
--- a/src/reducers/entitiesReducer.js
+++ b/src/reducers/entitiesReducer.js
@@ -608,21 +608,21 @@ const editCustomEvent = (state, action) => {
       "scenes",
       action.id,
       newState.entities.scenes,
-      patch.script
+      patch
     );
     newState = updateEntitiesCustomEventScript(
       newState,
       "actors",
       action.id,
       newState.entities.actors,
-      patch.script
+      patch
     );
     newState = updateEntitiesCustomEventScript(
       newState,
       "triggers",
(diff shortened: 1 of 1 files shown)


chrismaltby:
Nice, that was a tricky one to spot! I've just had a go at your alternative idea also in commit d95749dc37681ed2b9277b4b31b61fa37b4c9b2d seems to work quite well!

pau-tomas:
The change works great. A small (somehow related) nit, would make more sense for Custom Events variables to be referenced with `$PX$` for Parameters rather than `V`?

Issue #176, opened 2019-05-17
MOD file tempo commands (Fxx) not interpreted correctly


timsamoff:
> @RichardULZ and @WAUthethird: I have composed some music that is working perfectly now, EXCEPT that none of the tempo data is saved in the file. Is this possible? Or am I stuck at 125 BPM?

Sorry! Just figured out the Fxx commands!

_Originally posted by @timsamoff in https://github.com/chrismaltby/gb-studio/issues/83#issuecomment-493292900_

timsamoff:
To elaborate: Set some commands in your MOD file to F03 (speed) and F80 (BPM). The music will play at a different speed in the tracker than in the game. The Fxx commands do seem to be working to some extent, just not properly...

Likewise, I have found that once a tempo is set with the first song, the temp can't be changed when a new song begins with new tempo settings.

RichardULZ:
@timsamoff there are a ton of issue threads open now, and i feel this is an issue related to documentation that was recently fixed.
https://www.gbstudio.dev/docs/music/

You can not use bpm, you must use the lower range from 1 - 20 (in hex, the actual value max is 32) for tick speed / notes per frame.


Extra inaccuracies come from mod trackers working in 50 fps/ticks and the Gameboy using 60, which the maker of gpt player has tried their best to fix.

Thread for improving docs.
https://github.com/chrismaltby/gb-studio-site/pull/5

I have also not encountered the speed being locked, every song should set the speed at the start of a song, default speed will not restore, haven't tested if speed can change mid song, but that should work.

If you're ok with this answer, could you close this thread?

timsamoff:
I am using the hex method of controlling tempos. When I get a chance, I'll upload two songs that are giving me trouble to test with.

Issue #172, opened 2019-05-15
[Feature] Ability to preview Music in-engine as if emulated


timsamoff:
**Note: for support questions, please use one of the following channels**.

Reddit: [/r/gbstudio](https://www.reddit.com/r/gbstudio)
Discord: [Join Chat](https://discord.gg/bxerKnc)

This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Currently, music can be previewed from the Scene Script, but it plays all Mods as if they were compatible in the emulator (i.e., it's just a Mod player and doesn't care about format/instruments). It would be nice to be able to hear music as it will sound when emulated, without having to Build and test.

* **What is the expected behavior?**

To preview music as it will actually sound in the game.

* **What is the motivation / use case for changing the behavior?**

Editing efficiency.

* **Please tell us about your environment:**

  - Version: 1.0.0
  - Platform: all

RichardULZ:
It is possible we could butcher the current script tracker to have less features, to more closely resemble in game music. https://github.com/chrismaltby/gb-studio/tree/develop/src/lib/vendor/scriptracker

The grand solution would be to compile a game with one song when play is clicked, but you would loose the immediacy of play, as it would still have to build the song and game.

timsamoff:
This solution would need more consensus, I guess, but it would work for me -- much better than compiling and playing through many minutes just to finally get to the scene where that music is located.

RichardULZ:
Which solution?
1 hacking up the JavaScript mod player for instant rough testing, or
2 A compiler that builds a project with 1 song and emulates it exactly.

I normally have a seperate project just for song tests, but all song get recompiled each build.

You can also press "P" in a scene to set your starting position, provided you're not setting up input scripts at start.

Or set up dummy scenes for each song on start, and press p in one to test, if you can handle compile times.

timsamoff:
Good question. 👍🏼

I'd love for others to help make a decision for best overall UX.

(Btw, I didn't know about the "P" key trick.)

Issue #83, opened 2019-04-27
Inaccurate music in-game?


amysgames:
Hello! So I've been experimenting with this neat tool, and I've noticed a few issues with in-game music. Namely, it doesn't seem like it's playing notes in the correct octave, and it's not detecting tempo changes at all. This only applies when I run the game; the editor itself plays the music no problem. I'm not sure if it's compatibility issues with MilkyTracker, GB hardware constraints I'm not aware of, or something else?

WAUthethird:
Open the .mod file in OpenMPT and see if the issues arise or not.

amysgames:
It sounds the same in OpenMPT as it does in MilkyTracker. I tried saving it in OpenMPT and that didn't help.

I'm also fairly new to tracking, myself, so I'm not sure if maybe I did something wrong?

WAUthethird:
Hmm, I do know that the Hz of the song is changed when compiling the song, so it's normal if it sounds a bit higher or sped-up.
Are the differences very drastic?

amysgames:
It's definitely noticeable. It's playing it a full octave lower than what plays in MilkyTracker, and it disregards my tempo settings entirely (I tried setting it to something different with no change).

The instrument choices are drastically different as well, though I'd understand if there's some hardware limitations there.

JasonB32:
I'm having the exact same issue and I had no idea what was going on, I use OpenMPT as well, and yes Milkytracker plays it fine too, but GB Studio through GBT Player is playing it slower and lower :(

RichardULZ:
I'll make a tutorial on openMTP with Mod2GBT eventually, but right now, here's all the documentation.
Limits of mod2gbt https://github.com/AntonioND/gbt-player/blob/master/mod2gbt/mod_instructions.txt
Always start from a base file, instruments must have same layout. Download effects_test.mod https://github.com/AntonioND/gbt-player/tree/master/mod2gbt
A complete guide from zalo's ZGB which uses the same GBT player.

JasonB32:
Oh if you could please do a tutorial that would be great!!!

amysgames:
Oh, so it _is_ a compatibility thing, then. Alright, good to know.

With those restrictions in mind, I've managed to put together something that sounds a lot closer to what I want, but I guess there's not much I can do about the incorrect tempo (speed 5 is too fast and speed 6 is too slow). Oh well.

Thank you!

RichardULZ:
Try to space your song out differently! 
[...]

PR #328, opened 2019-11-16
Improve in editor music scriptracker.js to act like GBT

RichardULZ:

Bug fix / feature.

Split of #294 to make merging easier.
Related to #172, #83 and partly #176
The in editor music preview is very inaccurate and will play any mod files without GameBoyTracker limits, leading to confusion and not being useful for previewing common mistakes.

Notes use last set volume instead of the default sample volume.
Channel 4 Noise is forced to note C5.
Instruments are limited to their channels, (no pulse waves on the noise track.)
Forced pan to hard left, both, or hard right.
Tick speed changes the tempo to match in game speeds (tick speeds 4 or faster).
Instrument 28 pitched to better match in game noise.
Hard code Pulse and Wave sample instruments 1-14 as small, accurate samples.
(Catch users with wrong instruments in their mod. Sounds nicer.)
Add persistent Volume sweep for volume envelopes (not in GBT just yet ;)

* **Other information**:
Most cases except odd GBT bugs are now reproduced, (some gbt bugs are being fixed)
It is now useful as a fast preview that reveals common mistakes (volume), without building a game.

### DIFF
--- a/src/lib/vendor/scriptracker/channel.js
+++ b/src/lib/vendor/scriptracker/channel.js
@@ -70,8 +70,8 @@ Channel.prototype.reset = function() {
 	this.sample.remain   = 0;
 	this.sample.reversed = false;
 	
-	this.volume.channelVolume = 0;
-	this.volume.sampleVolume  = 0;
+	this.volume.channelVolume = 1; //GBT sets full volume on song start
+	this.volume.sampleVolume  = 1; //GBT sets full volume on song start
 	this.volume.volumeSlide   = 0;
 	this.volume.envelope      = null;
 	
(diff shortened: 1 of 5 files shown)


chrismaltby:
That's great, thanks @RichardULZ.

I'm wondering at some point we could maybe split this into a separate git repo scriptracker-gbt that GB Studio could import as a node module. Could be useful for other projects too! Something to think about

RichardULZ:
Not sure how making it a node module works but sounds neat.

At this stage there's not much more i can do just replicating GBT player bugs (I have a few being fixed in gbt player itself).
The major improvement would be to rip out the Javascript emulators audio code, and get that to work with the script tracker.

A half way point is having it work stand alone just to poke at when playing/making sound effects, seems like the only way to get the right tones, pitch bends, noise sounds, and fades.

Will finish off GBT improvements and volume envelopes first (they're pretty much done), got stuck with pitch bends, they work too, but it would be better if they worked more like a mod tracker for consistency.

PR #333, opened 2019-11-27
Update scroll registers on VBL interrupt

pau-tomas:
Bug fix. 

On scene, sprite position sometimes gets out of sync with background.

old_scroll
(It might be hard to appreciate in the GIF, but the rock sprite moves one pixel up and down as the player walks and the scene scrolls)

`SCX_REG` and `SCY_REG` are now updated on a VBL interrupt. Keeping backgrounds and sprites in sync.

new_scroll
(the rock sprite stays in the same position with the scene scroll)

After some quick tests, doesn't seem to break anything. However I'm not 100% sure of the possible side-effects of moving the scroll updates. Further testing would be good.

* **Other information**:

The change was inspired by ZGB logic for scroll https://github.com/Zal0/ZGB/blob/0f6f59eea1f4787ac76629661d2abc4ce8bb3b7a/common/src/main.c#L39

### DIFF
--- a/appData/src/gb/include/Scene.h
+++ b/appData/src/gb/include/Scene.h
@@ -60,6 +60,8 @@ extern UBYTE timer_script_duration;
 extern UBYTE timer_script_time;
 extern BANK_PTR timer_script_ptr;
 extern UBYTE scene_loaded;
+extern UBYTE scroll_x;
+extern UBYTE scroll_y;
 
 void SceneInit();
 void SceneUpdate();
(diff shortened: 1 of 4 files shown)


RichardULZ:
Nice, Seems you've caught for camera shake too, 
Don't think it needs a particular bank either so it shouldn't be screwing up any processes it interrupts.

chrismaltby:
Nice work! I'm away from home at the moment so don't have much ability to test this but sounds great.

I wanted to investigate ZGB a bit more actually, I ended up coming across the project after most of the game engine was already complete, it helped with getting some of the finishing touches working but I found it too late to fully use it.

For the next GB Studio version I'd like to start looking to expanding the supported game genres and it might be a good time for an engine rewrite built on top of ZGB.

chrismaltby:
Finally got round to merging this, works great thanks!

PR #342, opened 2019-12-04
Collision checkbox for Move Relative & To Variable

RichardULZ:

Feature, Adds a checkbox for enabling collision / disabling noclip with move relative and move to variable.

Always used NoClip for movement events, not fun for some trying to diy custom movement.

For relative and variable movement, you can tick Stop For Collision (defaults off).
This allows for creative uses like tile stepped movement, dash buttons, multi actor movement puzzles (2048?), 
Possibly projectiles? (Move relative with collision untill hit, then move 1 more without collision, and compare?)

May need automatic tests updated due to the extra value pushed.

* **Other information**:
Decided not to add this to "actor move to" so it stays compact, and lightweight.
Will update with a test project tomorrow.


RichardULZ:
Added a test project.
Tile based'ish movement, a Remote controll duck,
following duck, a strange multi move puzzle and A GUN AHHH!

Gun works by disabling player collision, setting bullet to actor position, and moving relative with collision based on direction, once it stops, it moves an extra 1 relative for up/down or 2 for left/right. Then the hitable actors get invoked to check if they're in the same spot. Then enabling player collision again.
You could previously use push actor after moving the bullet in front of the player, but it caused issues for close actors.

May update tileish based movement to repeat when held down like the rc duck, need to test if this affects player triggers...
Will upload the demo to itch.io
Still need to double check yarn test.

chrismaltby:
Hi @RichardULZ great idea!

I did a little testing and came across a bug where if the player is overlapping the actor that you're trying to move (for example when you're stood in front of them) while the checkbox is ticked it seems to block movement.

Not had much chance to look into it but I feel it's maybe an issue with the function SceneNpcAt_b which checks a few tiles for actor overlaps mainly so that you can interact with actors without having to be perfectly lined up with them.

I've added a test scene to the project file that might help narrow the issue down.

RichardULZ:
Ahh yes, the different check between 4 tiles or just bottom 2 tiles. 
This can occur with a random walk npc as well.

If an npc walks into you from below, you will be trapped unless going up, 
similarly if you walk into an npc from below, they will become stuck.

Any change made must be done with caution, as some users may have puzzles where an actor blocking all 4 tiles is useful for blocking a path.

RichardULZ:
Discussion on the Issue/bug from that weird can't move sideways when on top of actor problem.
https://discordapp.com/channels/554713715442712616/570925559346102273/655850150215942144

Perhaps we should look at selecting between 1x2 and 2x2 tile collision, where actors can't overlap, 
could cause issues for backwards compatibility of old puzzles... 

Sliding block puzzle in template requires the weird one we have now... closer to 1x2, but rocks blocking doors need 2x2, does that mean we need a dreaded legacy option? Or maybe we can get 1x2ish working better without breaking old stuff.

Would greatly improve quality of new block puzzles with 2x2.
(feature creep 1x1 collisions...)

chrismaltby:
Thanks for the extra info @RichardULZ 

Sounds like this one might be a little tricky, going to try to make some time to look into this when work calms down over Christmas.

Definitely agree about getting some options for different collision sizes, having movement use 16px tiles has been another big request, maybe this ends up being a little related.

PR #348, opened 2019-12-05
Editor keep scene collision when switching to bg without collision

RichardULZ:
Only keeps if it's the same width, and if there's no other collision currently for that background.

Feature / bug fix

When you change the background of a scene containing collision, it destroys the current collision and either copies collision if a scene with that background has some, or clears it.

If you change the background of a scene to one that has not been used yet, and is the same width, it will keep your current collision.

The original method takes priority, if it finds a scene it can copy collision from, it will.
If the scene is taller, it will work fine, if the scene is shorter, it will split/crop the collision array to size.
If the width changes, it will clear the collision (unless there's a background in use with collision to steal).

* **Other information**:
This was asked for on discord, and when i had the issue and found how frustrating it was, decided to fix it. 
It took me an hour or more just to find where background switching happened...

### DIFF
--- a/src/reducers/entitiesReducer.js
+++ b/src/reducers/entitiesReducer.js
@@ -426,6 +426,7 @@ const editScene = (state, action) => {
 
   // If switched background use collisions from another
   // scene using the background already if available
+  // otherwise keep old collisions if same width
   // otherwise make empty collisions array of
   // the correct size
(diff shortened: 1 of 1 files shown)


chrismaltby:
Works great, thanks @RichardULZ

Issue #353, opened 2019-12-07
Overlay event pauses timer events until removed


Polyducks:
Reported by @krümel (crumb) and tested/reproduced by Polyducks.

**Describe the bug**
'Overlay' events pause all timer events. If you have set a timer event and an overlay appears, it stops all timer events until the overlay is hidden. 

**To Reproduce**
1. Create new file
2. Create new scene
3. Create new actor
4. Reproduce code below

```
timer: set timer script
(countdown time in seconds: 0.5){
    if ($00$==true){
        $00$=false;
        actor_1(x=2, y=1)
    }else{
        $00$=true;
        actor_1(x=0, y=1)
    }
}
Overlay: show
black (x=0, y=4)
```

To see how the timer resumes when the overlay is hidden again, use Overlay:hide at the end of the above script.

**Expected behavior**
Unsure. Perhaps the overlay does not need to interrupt timer events - but if the overlay is computationally expensive it might make sense to pause all timer events. If so, this should be included in the documentation for events which interupt timer events (such as loop forever, multiple choice menus and overlay events).

**Platform (please complete the following information):**
 - OS: Windows 8.1
 - App version 1.2

**Additional context**
The same is true of multiple choice options and loop forever events - I expect this is by design to free up processing time. When the multiple choice menu is exited, or the loop is ended with stop script, the timer resumes.

RichardULZ:
This can be closed now. :)

PR #363, opened 2019-12-15
Fix timer script with show overlay, faster min speed, and performance improvement.

RichardULZ:

Bug fix & feature

#353 Show overlay halts timerscript
https://discordapp.com/channels/554713715442712616/585023243732123676/652881193359900695
Timer script faster please? https://discordapp.com/channels/554713715442712616/570925583421276160/654859741788504106

1. Allows show overlay at the end of a script, timer script still pauses for dialogue or anything during script.

2. Change the minimum timer script speed to 0.25 seconds and step in 0.25 increments instead of 0.05. 
Timer script ticks every 16 frames, about 0.26 seconds, so this is the fastest it will go. 
You can still type 6.712 or anything arbitrary between min max into the field. 
Seems to work fine @gregtour 

3. Performance change, removes checks for emote_timer on movement and timer script, as this should always need a script_ptr to be active, seems to run identically with this not checked.
Any issues you can think of with this change @chrismaltby ?

* **Other information**:

### DIFF
--- a/appData/src/gb/src/Scene_b.c
+++ b/appData/src/gb/src/Scene_b.c
@@ -725,12 +725,6 @@ void SceneUpdateTimer_b()
     return;
   }
 
-  // Don't update timer when UI is open
-  if (!UIIsClosed())
-  {
-    return;
-  }
-
   // Check if timer is enabled
   if (timer_script_duration != 0)
   {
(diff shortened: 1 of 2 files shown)


chrismaltby:
Nice one, thanks @RichardULZ

PR #366, opened 2019-12-25
Add support for multiple palettes for BG and sprites

pau-tomas:
*WARNING* Big patch. 

Opening a (realtively early) pull request to start discussing the implementation of full color support 🎨 . @chrismaltby I think you already started on supporting pngs with more than 4 colors. My approach is different as the palettes are defined in the app (with a palette editor) and can be assigned individually to each Scene background and sprite. 

I think the colored pngs is more straightforward, although it might be hard for users to enforce the palette limits in the background png (each tile can use 1 palette from up to 8). On the other hand the palette editor approach doesn't allow to have different palettes per tile, but allows to introduce Events to programatically change the palette of sprites and backgrounds. I wonder if there's a combination of both that might work.

- Separate setting for Background and Sprites default palette.
- Palette editor. Allows to create new custom palettes to be used in sprites and backgrounds.
- Palette selector in Scene Editor (for scene background palette)
- Palette selector in Actor Editor (for sprite palette)

Known issues:

- Sprite palettes have 4 colors and use the third one for transparency. It needs a way to make this clear to the user in the editor as it might be confusing.
- Sprites fade transition isn't implemented yet.

Settings allows to select one custom palette for both backgrounds and sprites.

Multiple palettes can be used for sprite and backgrounds.

Currently it probably breaks games using custom palette. A migration script will be needed.


chrismaltby:
Hi @pau-tomas, I've just been playing around with this build, nice work!

I think this approach is a great for adding more colour while still keeping it quite easy to manage.

As you say supporting full colour pngs is a little awkward with the limitations that are needed. I've toyed around with a few ideas:

1. Allow full colour pngs, scan the image to determine the 8 colour palettes used, if more than 8 are used just use the first 8 (or ideally the most distinct 8) and assign each tile to the closest matching palette. I've started to get somewhere with this but it's a little tricky.

2. Alternatively I can make it so PNGs still need to use the current 4 colours but in the app I would add a palette painting mode. You would be able to choose 8 palettes per scene and then like you do now for drawing the collisions you would be able to paint the palettes on to the backgrounds. At this point I'd probably also look to start allow providing tileset pngs and creating the entire scene image in app too, similar to RPG Maker.

I'm leaning towards 2. but I'm not sure if I might need some way for 1. to work also for people who'd rather just use Aesprite/Tiled.

For actors I was thinking something very similar to what you've implemented so I think I can probably adapt what you've written here.

Just to let you know where I am with all this though, I'm currently trying to attempt quite a few things at the same time with a big engine rewrite which is now based roughly on [ZGB](https://github.com/Zal0/ZGB). It's not ready to be shared yet but the current major features in varying levels of working are:

- Support for much larger maps (max 2040px in a single dimension right now & max width x height needs to be less than 1,000,000px)
- The Scene type dropdown now works and lets you choose "Top Down" or "Platformer" both very broken but getting there, planning a few other options here too.
- Allows storing 8 colour palettes per scene and using a different palette per tile, (no way hooked up for this to work in UI yet, just hard coded into the compile process while I work on it)

I also made a small CLI to build ROMs by specifying a gbsproj file path, I've got it set up to live reload building/running a test game whenever either the C engine source or the JS compiler source changes, was getting to be a pain refreshing all the time!

Lots of things still need to be done but I'm basically mashing together ZGB and the old engine while also trying to keep the bits that make the game top down or platformer as separate as possible so that I can potentially expose that for people too so they can write new game styles with some C/GBDK (I'll see how that goes, but that's the plan!).

My current priorities are:
1. Get a few of the more urgent bug fixes together and release a 1.2.1 version (done quite a few of these, just need to work out where to draw the line, #342 might end up being 1.2.2)
2. Get the Webpack/Typescript version ready to be used (pretty much is actually, on my Mac at least!). Going forwards new components and helpers would be written in Typescript and I'll start migrating old functionality over as it gets worked on.
3. Get the new engine working.
4. Build the UI for adding colours / tile painting etc.

Once I've got the engine ready I'll be ready to start picking parts from this PR to get the colour UI implemented.

Thanks for your help.

pau-tomas:
I think I lean towards option 2 also. One advantage of it is that games will be compatible with the original GB by default (as backgrounds will be always 4 colors). Back when I was starting hacking GBS I built a very rough pixel editor with the intention of building a tile editor. It didn't look bad, although all the image processing (to, for example, extract the tiles) is done in the main thread so it was a kind of slow with big images:

image

A tile editor will be also useful as a base for a HUD editor (I've [been playing](https://github.com/pau-tomas/gb-studio/tree/feature/hud) with the overlay and `UIDrawText` to see what can be done in that area, but right know is mostly a hack).

Super pumped for the new features! Let me know if you need any help once you start implementing the sprite palettes or anything else.

Related to the new engine, I had a chat the other day on the gbdev Discord #gbdk channel, seems like there's a GBDK bug that corrupts some tiles in specific cases when tiles, map entries, palettes, or any OAM value are modified during a period of active interrupts (music, VBlank, etc.). Zalo was taking a look at how it affects ZGB and it might happen in GBS as it gets more complex and flexible. The discussion is [here](https://discordapp.com/channels/303217943234215948/571488413262544896/662364267000889344) if it might be of help.

blurymind:
Could perhaps piskel be adopted to do the pixel editing ?

blurymind:
btw we are using piskel in gdevelop, which is very similar to this project - a reactjs app running in electron
https://gdevelop-app.com/el/make-8bit-pixelart-games/

Issue #369, opened 2020-01-03
Crash probably linked to Invoke Script command


Atnik:
I've run into a strange bug, which I don't understand myself, so I've attached the project below.

When you run the game, a battle sequence starts, but once the character Hugo finishes his move, the game crashes. This only happens with this character and I've checked the script several times, but it should work like the other characters.
But when you copy the script of the actor Hugo into the "Battle Loop" part in the scene init script it works again (at least it did after I removed the "If Variable has Flag" command at the start of the character script and added it again).

Before trying to find a solution, I got crashes that looked like this:
![grafik](https://user-images.githubusercontent.com/9643886/71745967-072da900-2e6c-11ea-9fbb-74d63b0681a7.png)

But now the game just stops without messing up the graphics.

Here's the project: [Dropbox](https://www.dropbox.com/sh/387nw3mx5dnlrdl/AAAc8e5uuodBLpVio4dzpwZBa?dl=0)
If you go to the scene init script and disable the invoke script command labeled Hugo and enable the one below, you can see that the same script works in the main script.

OS Version: Windows 7 64bit
GB Studio 1.2 (also happens in the dev build)

I know this is all super vague, but I hope it still helps to find the bug.

RichardULZ:
A minor oversight with **Invoke actor**, nothing to do with labels fortunately.

To return after an actor was invoked, we stored up too 8 last `script pointers`,
Later we also stored the `start script pointers`, as that's used as an offset for IF or Switch events,
But i didn't think about actor scripts that might be so long they are in another bank,

So while invoking actor used the standard *start this script* function, which set the `script ptr`, `start ptr`, and `script bank`, we never stored the previous `script bank`s to reload, thus on return, loading garbage data from the new bank, and crashing.

It was lucky it happened with a long script to give me a lead, as it could just as easily happen somewhere where we randomly ran out of space and switched banks, and i would have been stuck digging through labels.


On a different note, the BANKED_DATA_TOO_LARGE warning is pretty vague, should probably error with what script is too large? You can reproduce this with this project file by copying all the actor scripts back into the start script.

RichardULZ:
Added some more tests for BANKED_DATA_TOO_LARGE, now tells you by how much, what script, and to try splitting with invoke actors instead.

RichardULZ:
Merged! Try out the dev build in the github readme and close this issue please.

Atnik:
Tested it and everything worked! Thanks a lot for the help! :)

PR #370, opened 2020-01-04
Fix Invoke actor for scripts in other banks

RichardULZ:

Bug fix for invoke actor, similar to the last invoke actor fix.

#369  Invoke actor will cause a crash on return if the actor scripts are in different script banks.
Rare bug, only dealing with large scripts that happen to be in different banks.

 May be slightly slower for invoke actor events, but will fix a game crash that is not easy to work around. the demo in #369 will work now.

Scripts that call actor invoke frequently in a loop may notice more lag, but it's already pretty slow, label define and label go to will likely be faster if you don't run out of space for a single script. 

* **Other information**:
https://discordapp.com/channels/554713715442712616/570925559346102273/662789686598565888
https://discordapp.com/channels/554713715442712616/575070105919029295/662798184996667405
A check / warning for scripts that are too large and spill into another bank would help, as it creates a similar issue.

### DIFF
--- a/appData/src/gb/include/ScriptRunner.h
+++ b/appData/src/gb/include/ScriptRunner.h
@@ -20,6 +20,7 @@ extern UBYTE script_actor;
 // Max call stack depth
 #define STACK_SIZE 8
 extern UWORD script_stack[STACK_SIZE];
+extern UWORD script_bank_stack[STACK_SIZE];
 extern UWORD script_start_stack[STACK_SIZE];
 extern UBYTE script_stack_ptr;
 
--- a/appData/src/gb/include/ScriptRunner.h
+++ b/appData/src/gb/include/ScriptRunner.h
@@ -20,7 +20,7 @@ extern UBYTE script_actor;
 // Max call stack depth
 #define STACK_SIZE 8
 extern UWORD script_stack[STACK_SIZE];
-extern UWORD script_bank_stack[STACK_SIZE];
+extern UBYTE script_bank_stack[STACK_SIZE];
 extern UWORD script_start_stack[STACK_SIZE];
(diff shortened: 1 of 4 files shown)


RichardULZ:
Added a warning for scripts that can't fit in a single 16384 byte bank, so it's not just `BANKED_DATA_TOO_LARGE`, but also says by how much, what script is too big, and to try splitting with Invoke Actors.

The test project in #369 is a good for this. You can copy all 4 actor scripts into the start script and be just over the limit, disabling some chunks of this script will get it working again.

It seems unlikely for many smaller scripts to be too big for a single bank, but if that is a case, `BANKED_DATA_TOO_LARGE ` will show without the other warnings.

chrismaltby:
Works great from my testing and seems to fix the issues seen in the project attached to #369 

Nice improvement on the error message too.

I've been trying to think for a while if there's a good way to automatically split large scripts across multiple banks but it gets tricky if the split happened in the middle of a loop or if you use label/gotos anywhere. 
You could make the JUMP command include a bank number, and add a JUMP to the start of the next bank at points you need to split but it's awkward because at the time the scripts get compiled you don't know which bank the script will eventually end up in and at which points the script will need splitting. Not sure the performance impact of this either.

Maybe a post script compilation step that before banking the data in compileData.js takes the large script, the current bank and the amount of space left in the current bank and splits it into an array of smaller scripts linked together where the first one fits the remaining space and every additional block fits in an empty bank.

(LARGE SCRIPT, currentBank, spaceRemaining) => [SMALL SCRIPT 1 (smaller than remaining space in current bank), SMALL_SCRIPT 2 (smaller than 16384), SMALL_SCRIPT_N...]

I'll keep thinking about this.

PR #393, opened 2020-02-17
Custom Events 2

pau-tomas:
Bringing Custom Events 2 to GB Studio 2

**TODO**

- [ ] **New in CE2** Compile error when `Set Actor Sprite` is used inside Custom Event
- [ ] **New in CE2** Wrong sprite used when `Item: Launch Projectile` or `Item: Attack` are used inside a Custom Event
- [ ] **New in CE2** Game crashes/glitch when `Text: Display Dialogue` with avatar is used inside Custom Event
- [ ] **New in CE2** `otherActorId` doesn't work as parameter
- [x] **Happens with CE1** invalid variable value when selecting `variable` from the `number|variable|property` dropdown (for example in `Variable: Set To Value`)
- [x] **Happens with CE1** Custom Events contained in Hit scripts aren't updated when Custom Event is updated
- [x] **Happens with CE1** `direction` doesn't work as parameter

Invisible feature change that updates the Custom Events to use a single scripts rather than duplicate the script on each Event call.

The first version of Custom Events worked by duplicating the script
inside the Event on each of its instances. This meant that scripts
could grow too big without users even noticing.

This update makes Custom Events single scripts in memory and each
call to them jumps the script execution to the memory position and
bank where the script is located. Before that, the parameter values
are copied into "local" variables. Once the script is executed the
local variable values are copied back to the parameter ones,

For actors, a new internal C script has been introduced that allows
to set the active actor from a variable value.

Yes. A migration script is included in the PR. It updates existing scripts, as the field names for the parameter on the `EVENT_CUSTOM_EVENT_CALL` Event have been updated.

It also tries to take care of some edge cases that happened due to copy and pasting EVENT_CUSTOM_EVENT_CALL events between projects without copying the called Custom Event. In 1.2.0 the game will run as expected using the `script` embedded in the Call event. But since that's not the case with Custom Events 2. The migration script transforms those EVENT_CUSTOM_EVENT_CALL evens into Event Groups containing the script and adding a comment trying to explain what happened. _maybe is a bit overkill for an edge case but I'm a bit concerned about breaking the migration script due to an unexpected edge case like this 😓_ 

Other than that the feature should be invisible to the users other than reducing the size of the project.

* **Other information**:

TODO:
- [ ] Further testing

This change also opens the door to Custom Events inside Custom Events. It still poses some implementation challenges (ie: for recursive Custom Events as it'll require a variable stack for parameters) but should be simpler to implement than with the old method.

### DIFF
--- a/appData/src/gb/include/ScriptRunner.h
+++ b/appData/src/gb/include/ScriptRunner.h
@@ -185,5 +185,7 @@ void Script_PalSetSprite_b();
 void Script_PalSetUI_b();
 void Script_ActorStopUpdate_b();
 void Script_ActorSetAnimate_b();
+void Script_CustomScriptInvoke_b();
+void Script_ActorActivateVar_b();
 
 #endif
--- a/appData/src/gb/include/data_ptrs.h
+++ b/appData/src/gb/include/data_ptrs.h
@@ -33,6 +33,7 @@ extern const BankPtr sprite_bank_ptrs[];
 extern const BankPtr scene_bank_ptrs[];
 extern const BankPtr collision_bank_ptrs[];
 extern const BankPtr avatar_bank_ptrs[];
+extern const BankPtr custom_events_bank_ptrs[];
 extern const unsigned int bank_data_ptrs[];
 extern const unsigned int music_tracks[];
 extern const unsigned char music_banks[];
--- a/appData/src/gb/src/core/ScriptRunner_b.c
+++ b/appData/src/gb/src/core/ScriptRunner_b.c
@@ -149,6 +149,8 @@ const SCRIPT_CMD script_cmds[] = {
     {Script_PalSetUI_b, 2},            // 0x66
     {Script_ActorStopUpdate_b, 0},     // 0x67
     {Script_ActorSetAnimate_b, 1},     // 0x68
+    {Script_CustomScriptInvoke_b, 1},  // 0x69
+    {Script_ActorActivateVar_b, 2}     // 0x6A
 };
 
 void ScriptTimerUpdate_b() {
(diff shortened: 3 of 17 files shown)


chrismaltby:
Thanks @pau-tomas sounds like a great idea, and should keep the size down in projects that make heavy use of custom events and like you say potentially allow nesting the events even.

There doesn't seem to be anything in this PR that should be a problem in the new engine too. I've kept scripting mostly identical as before but I'm just going through implementing the functionality that the scripting triggers at the moment so even the sample project doesn't even work completely yet (Scene transitions missing / Emotes currently crashes / Can't push actors).

pau-tomas:
Still needs a bit of cleanup, and a proper test of the migration script as I'm not convinced it works in all cases.

After a bit of testing I found one edge(ish) case with this approach that I wanted to mention. 

Since all variable are "global", including the ones inside a Custom Event. If a Custom Event has a `Timer` or an `Attach Script to Button` script using those variables, any call to that Custom Event will modify the value of the variable for those scripts.

For example for this Custom Script:

```
Foo(VariableA, ActorA)
  if (VariableA == true)
    Attach Script to B
      Actor:Emote(ActorA, Shock)
  else
    Actor:Show(ActorA)
```
the first it's called with:

```Foo(true, Actor 1)```

Then on pressing `B` the `Actor 1` will show the Shock emote.

Then the Custom Event is called with:

```Foo(false, Actor 2)```

Even if the script isn't attached to B (because the first parameter is `false`). The value of `ActorA` inside the Custom Event will change to `Actor 2` (because the active actor for `Actor:Show`, in this case `Actor 2` is attached to the variable `ActorA`). 

So the next time `B` is pressed `Actor 2` will show the Shock emoji. Even if that's not the expected behavior.

I'm not sure how much should we care, because it's kind of an edge case and it already happens, to some extent, with local variables and attach scripts.

um3k:
Is adding direct access to global variables (not having to pass them in) within scope of this project?

pau-tomas:
@um3k I'm not planning to add any new functionality with this PR. But using global variables inside Custom Events is something that it's in my TODO list to investigate.

PR #395, opened 2020-03-01
It localization

marcosecchi:
Somehow is seems last localization commit was not merged and some elements are still in english.


chrismaltby:
Thanks @marcosecchi

Issue #404, opened 2020-04-14
build fail: bank overflow (addr 40 > 40)


numonkeys:
Issue:
Game failed to build/run locally.

How it happened:
I hit "run" to test new code.

Expected result:
Game should have compiled and run.

Actual result:
make error 1.

Screenshot:
![Screen Shot 2020-04-14 at 1 37 52 PM](https://user-images.githubusercontent.com/3458886/79272127-d5a67f80-7e55-11ea-8abc-c54e30c78d9c.png)

Notes:
On our 3rd month of straight coding, just did a big update that removed 2 dozen (24) scenes from the game and replaced them with 2 scenes. Was hoping this would significantly reduce game size and complexity with a simpler and more intuitive experience.

However on build to play test, the GB Studio failed to compile.

Not sure if the game is too big (last compiled game.gb file is only 1 MB, from one week ago). Have used 350 variables so far, dev environment appears to max out around 550 so still under limit here. No MODs added to game yet, just scenes and sprites.

Final note -- yesterday when I tried to compile / run this game, the build error was "possibly wrote twice at addr". Today I coded a bit more and even added some more scenes, then got a new error message "bank overflow (addr 40 > 40)".

Any suggestions?

pau-tomas:
Hi @numonkeys, a couple of questions to better understand the issue: are you using/calling a lot of Custom Events by any chance? How big is the .gbsproj file of the project?

Also it'd be very useful if you could zip the project and attach it to this issue, that way we can run some tests locally.

RichardULZ:
Sounds like you have a lot of very complex scripts, or a lot of very unique backgrounds, currently we're limited to 1mb, 64 banks, but it should be possible to raise that.

Have you got some kind of heavy battle script reused in a lot of scenes? That's so far the only thing i've seen to break 64 banks.

As pau-tomas mentioned, if you're using a lot of custom events, that might be an easy way to bloat scripts on 1.2.1, however, it's also being changed, so that code is reused See #393

numonkeys:
Hi guys --

Thx for these questions.

I've been using zero custom events. My brother had set one up but I didn't find it useful (to be honest, I've tried coding my own custom events too but can't figure out how to get them to use in-game variables so have been ignoring custom events altogether).

[...]

PR #408, opened 2020-04-17
Remove string_bank_ptrs array, instead inject string_ptrs directly into script arguments

chrismaltby:
Bug fix for #404 where a project contained enough text to stop the `string_bank_ptrs` array from fitting within bank 5 causing an overflow.

Bank 5 contains an array of pointers to each string in memory. When displaying a string such is in a Display Dialogue event it is looked up by index in the array (16bit value) to find the bank it's stored in (8bit value) and it's offset in the bank (16bit value)

The `string_bank_ptrs` array has been removed entirely and instead every time a string is needed the bank and offset is stored directly within the event.

So where previously a text command contained the following:
```CMD_TEXT, index_hi, index_lo```
where index_hi and index_lo were 8bit values used to construct the 16bit string index which would in turn be used to look up the string bank and offset

It will now contain
```CMD_TEXT, string_bank, string_offset_hi, string_offset_lo```

This adds an extra byte to every text command but greatly increases the amount of text that can be used in game as it no longer fills bank5 with each string added.

Hopefully not, from my testing it seems to work correctly and the project that was failing now compiles as expected.

### DIFF
--- a/appData/src/gb/src/ScriptRunner.c
+++ b/appData/src/gb/src/ScriptRunner.c
@@ -13,7 +13,7 @@ UWORD script_ptr = 0;
 UWORD script_ptr_x = 0;
 UWORD script_ptr_y = 0;
 UWORD script_start_ptr = 0;
-UBYTE script_cmd_args[6] = {0};
+UBYTE script_cmd_args[7] = {0};
 UBYTE script_cmd_args_len;
 SCRIPT_CMD_FN last_fn;
 
--- a/appData/src/gb/src/ScriptRunner.c
+++ b/appData/src/gb/src/ScriptRunner.c
@@ -24,7 +24,7 @@ UWORD script_start_stack[STACK_SIZE] = {0};
 
 SCRIPT_CMD script_cmds[] = {
     {Script_End_b, 0},                // 0x00
-    {Script_Text_b, 2},               // 0x01
+    {Script_Text_b, 3},               // 0x01
     {Script_Goto_b, 2},               // 0x02
     {Script_IfFlag_b, 4},             // 0x03
     {Script_Noop_b, 0},               // 0x04
@@ -62,7 +62,7 @@ SCRIPT_CMD script_cmds[] = {
     {Script_SetFlagValue_b, 3},       // 0x24
     {Script_IfValue_b, 6},            // 0x25
     {Script_IfInput_b, 3},            // 0x26
-    {Script_Choice_b, 4},             // 0x27
+    {Script_Choice_b, 5},             // 0x27
     {Script_ActorPush_b, 1},          // 0x28
     {Script_IfActorPos_b, 4},         // 0x29
     {Script_LoadData_b, 0},           // 0x2A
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: appData/src/gb/src/ScriptRunner.c [hunk 1]
```diff
@@ -24,7 +24,7 @@ UWORD script_start_stack[STACK_SIZE] = {0};
```
RichardULZ, line 24: Took a while, but Menus are broken because max commands is still 6
Line 16, should be `UBYTE script_cmd_args[7] = {0};`
@chrismaltby

RichardULZ:
I can't figure out an easy way to build off this branch, but here's the bank iterative allocation i tried too.
It brought numonkeys bad build from 63 banks down to 57, so it's a worth while space savings.
replaces lines 45-66 of bankedData.js, can reduce the look behind distance from 10 if wanted.

https://github.com/chrismaltby/gb-studio/blob/d0a7fbde776c97e7aad55fc99f49a869dabaf5f6/src/lib/compiler/bankedData.js#L46-L66
```
       // For last 10 banks, until newest bank
   for (this.currentBank = Math.max(0,this.currentBank-10); 
   this.currentBank <= this.data.length - 1; this.currentBank++) {
    if (this.data[this.currentBank].length + newData.length < this.bankSize) {
      const writeBank = this.getWriteBank();
      const ptr = {
        bank: writeBank,
        offset: this.data[this.currentBank].length
      };
      // Bank has room, append contents
      this.data[this.currentBank] = [].concat(
        this.data[this.currentBank],
        newData
      );
      //console.log("Loop Bank:" + this.currentBank + "/" + this.data.length-1 + 
      //" Length:" + this.data[this.currentBank].length + " + " + newData.length);
      return ptr;
    }
   }
    // Current bank is over size, make a new one
    this.currentBank = this.data.length;
    const nextWriteBank = this.getWriteBank();
    const ptr = {
      bank: nextWriteBank,
       offset: 0
    };
    this.data[this.currentBank] = newData;
    //console.log("bank:" + this.currentBank + "NEW BANK! Length:" + this.data[this.currentBank].length);
    return ptr;
  ```

File: src/lib/compiler/compileData.js [hunk 2]
```diff
@@ -279,6 +279,28 @@ const compile = async (
...
+      if(data.startsWith("__REPLACE:STRING_HI:")) {
+        const index = parseInt(data.replace(/.*:/,''), 10);
+        return hi(stringPtrs[index].offset);
+      }
+      if(data.startsWith("__REPLACE:STRING_LO:")) {
+        const index = parseInt(data.replace(/.*:/,''), 10);
+        return lo(stringPtrs[index].offset);
+      }      
+    }
+    return 0;
```
pau-tomas, line 301: Not sure under what circumstances but I observed some cases where the `data` is of type `string` even if it's a number (you can see it happening in the bad build from numokeys if you `console.log(data)` when `typeof data === "string")` before the return statement).

To protect form those cases return could be `!isNaN(parseInt(data))?parseInt(data):0;` instead.

chrismaltby:
Thanks for fixes both! I've implemented them and is working again now.

@RichardULZ thanks for the idea on iterative allocation, I've not included it here but I'm going to put up a separate PR where I've used that method in a minute.

PR #409, opened 2020-04-17
Configurable cart size

pau-tomas:

Feature. Makes the `-Wl-yo` flag that controls the cart size in the makefile configurable via settings. 

Cart size flag is fixed to `64`

Cart size defaults to `64` but can be updated by manually adding `"cartSize": XX` to the `settings` section of the project `gbsproj` file. XX has to be a power of 4.

No. The default cart size value is still 64.


chrismaltby:
Thanks @pau-tomas I've used this to make a slightly different attempt in #410

PR #411, opened 2020-04-18
Fix/invalid custom events actors when pasting external scripts

pau-tomas:

Bug fix

When non-custom events scripts are pasted into custom events
the actor fields referencing scene actors weren't rendered
properly inside the custom event.

This patch adds a fix method that check if the actors are
scene actors and returns "0" (Actor A) if that's the case.

No

* **Other information**:


chrismaltby:
Thanks @pau-tomas looks good, I think you've just accidentally included the cart size commit in with this PR though.

If you remove that I think this should be good to merge.

pau-tomas:
Fixed. Thanks!

PR #410, opened 2020-04-18
Feature/dynamic rom sizing

chrismaltby:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Allows the generated ROM size to be dynamic based on the amount of game content.

* **What is the current behavior?** (You can also link to an open issue here)
All ROMs generated by GB Studio are 1MB in size, creating a maximum amount of content that you can add to your game and using much more space than required for smaller games.

* **What is the new behavior (if this is a feature change)?**
GB Studio determines the amount of banks required while compiling your game and sets the ROM size accordingly. The sample project with no modifications is now 256KB

Supported ROM sizes are:
- 128KB
- 256KB
- 512KB
- 1MB
- 2MB
- 4MB
- 8MB

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

Hopefully not, but I'd appreciate if anyone could test their projects using this branch and report if there's any issues.

* **Other information**:
This PR is heavily influenced by @RichardULZ and @pau-tomas and is a combination of their suggestions in https://github.com/chrismaltby/gb-studio/pull/408#issuecomment-615390717 and https://github.com/chrismaltby/gb-studio/pull/409 so big thanks to both of them for helping with this feature!

Also I've increased the number of banks that can be used for music from 8 to 16, a better way for allocating music to banks is needed but this should help anyone who is running out of space for now.

PR #413, opened 2020-04-21
Fix bad cursor when frame and cursor bank differ

RichardULZ:

Bug fix

If the cursor is dynamically allocated a bank before or after ui border, we would read garbage.
This has been a rare issue before this, where the cursor would be in the next bank over, but has been more apparent with dynamic filling as the tile is small.
#410

Read the cursor bank, have the correct cursor, easy to do, hard to find.

None!

* **Other information**:


pau-tomas:
Good find! You're on 🔥 @RichardULZ

chrismaltby:
I just went to look at this bug in my version and even after reading your code I still couldn't see it! :-D Good find!

PR #415, opened 2020-04-21
Add switch background only event, (to reduce duplicate battle scripts per enemy etc.)

RichardULZ:
Loading a scene bg, but the actors and scripts don't change/stop,
Usefull for battle systems, faster cutscenes, 

Feature
Adds a new event to load any scenes background, without hiding the actors or stopping the scripts.

There is no way to change the background without destroying the actors and scene script, 
Pokemon like battles with complex scripts are duplicated for every enemy type. And there is no way to change the background without fading to white.
Some users have battle scripts that take up an entire bank, and copies of them are difficult to maintain or edit.

Adds a new event to load any scenes background by itself, without hiding the actors or stopping the scripts.
This does not fade the screen before or after the swap, 
if a background is similar, eg switching between monsters, using similar tiles, or canging to a nighttime background, the transition is seamless. 
With complicated scenes some visual artifacts may be visible during the transition, but can be desired or a fade added before in script.

Regular scene switch has not been modified, but parts are reused, so no issues should appear.

* **Other information**:
L10n translation has not been added yet, but otherwise feature works without issues.


chrismaltby:
Hi @RichardULZ I know this is one of the 1.2.1 events I never got around to merging but could probably pick this back up as a 2.0.0 feature. Only thing I'm thinking though is what to do about the colors of the tiles for GBC games?

We could leave the colouring the same as the existing scene image but that might not be what would be expected.

Maybe we could use the colouring from an existing scene that used that image if there was one and fall back to not using the colouring if not. Feels a bit awkward but because I've tied the colouring to the scene and not to the image itself, there's no obvious way to do it.

Or I could just move the colour data to be per image rather than per scene... hmm...

pau-tomas:
A bit of an overkill solution, but what if the scene can be viewed with the different backgrounds in the World Editor and the color tiles can be edited for each of them?

The downside is that it might be confusing since collisions and triggers will apply to all the backgrounds but color tiles won't.

RichardULZ:
This originally for ease of programming actually modified a copy of switch scene event, instead of a background picker, so although saying set background, you would be picking a scene to use, (makes sure its on the map/compile list and everything) it could stay that way.

maxoakland:
Is this something that could be incorporated into version 3?

chrismaltby:
Finally closing this one since the implementation won't work in 3.0 but maybe we could still support this at some point

PR #416, opened 2020-04-22
Fix Variable Set to Value causing Nan / Null if no value (and optimize)

RichardULZ:

Bug fix, Optimization

Variable Set to Value will send nan / null data if 0 is not typed in, despite defaulting to 0...
https://discordapp.com/channels/554713715442712616/570925559346102273/702357708073992243 

If value is greater than 1 use value command,
If value = 1, set true command (faster),
Else, (0 or no value), set false command.
No more Null error message (because we check all the data now).

* **Other information**:
User also notes music compile is slow, but is improved with new mod2gbt i'm working on, seems to slow over time though, perhaps log too long, or something else?

### DIFF
--- a/src/lib/events/eventVariableSetToValue.js
+++ b/src/lib/events/eventVariableSetToValue.js
@@ -16,13 +16,14 @@ export const fields = [
 ];
 
 export const compile = (input, helpers) => {
-  if (input.value > 1)  {
+  const value = parseInt(input.value, 10);
+  if (value > 1)  {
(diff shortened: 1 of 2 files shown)


chrismaltby:
Thanks @RichardULZ works great, just made a minor change to use strict equality and to confirm the value is an Integer first but other than that it's all good.

I also added some additional unit tests to confirm the setValueToTrue and setValueToFalse events were getting called correctly with different inputs.

Issue #103, opened 2019-05-01
Add "SAVE AS" to save the project as another file


Audio-Rochey:
Hello folks,

Often, I want to save a clone of the project and start ripping pages etc out to test an advanced bit of functionality. we only have a "SAVE" function at the moment. It'd be nice to have a "SAVE PROJECT AS" that'd clone everything in the directory to another folder, so I can tinker with that without breaking the main project.

/Dafydd

thomas-alrek:
I agree, however, this is exactly what Git is designed for. If you have your project in source control, you can just create experimental branches, and revert back later.

chrismaltby:
This is finally available in the dev branch now thanks to a recent update by @zdurgan 👍

PR #417, opened 2020-04-23
Feature/save as

zdurgan:

Implements a save as feature. Allows users to save a copy of a modified or unmodified project without reloading.

Also includes an update to the Chinese language pack to include "Save As".

As shown by #103,  currently GB Studio does not support a save as feature.

GB Studio now displays a "Save As..." option in the File drop down. When clicked a dialog opens that allows the user to name a new project. Once the user clicks save, the project is copied to the new location requested and the modified state is updated if it was set. 

The user should not need to do anything to their project for it to work with this pull request.

* **Other information**:

In the developer console, you will get warnings about prop changes and re renders happening with no data change.


chrismaltby:
Excellent! Nice implementation of an often asked for feature!

I made a few small changes as it seemed the check for the project already existing was using a slightly wrong path so never triggered for me and I also made it add the new project to the top of the Recent Projects list in the splash screen. Slightly changed the error message too but I don't know any Chinese to add that translation for you!

Really appreciate you putting in the unit tests for this too 👍

Issue #398, opened 2020-03-15
Can't export or run on Linux


terminaltux:
Hello,

**Describe the bug**
While trying to Run or Export ROM my project on Fedora 31, GB studio just hangs.

When I choose Run or Export ROM, the Build & Run screen will say "Done!" at the bottom, but nothing happens. The Run icon in the top right keeps turning.

After a fresh install of GB Studio, it may Run or Export the project successfully once. But, further attempts fail.

**To reproduce**
1. Start a new Sample Project on a Fedora computer and save it.
2. Try to Run or Export the project
3. If successful, close GB Studio, launch again, and try again.

**Expected behaviour**
GB Studio will export a ROM file, or load the emulator to run the game.

**Platform**
- OS: Fedora 31 (Linux)
- App Version: 1.2.1

**Additional Information**

1. Often, on the Build & Run screen, GB Studio stops at "Copy code bank_7.c".
2. It's possible that Ubuntu works, but Fedora doesn't because Fedora has a newer version of **gcc** (7.5.0 vs 9.3.1) and **make** (4.1 vs 4.2.1). After a fresh install of Fedora, GB Studio worked, but after updating the OS, it stopped working.
3. On an up-to-date Ubuntu LTS 18.04.4 computer, GB Studio will fail with the error:

```
/bin/sh: 1: make: not found
127
```

This can be resolved in Ubuntu by installing gcc and make, ie:
`sudo apt install gcc make`

4. I tried installing the development tools in Fedora, but it didn't resolve the issue, ie:
`sudo dnf install @development-tools`

zdurgan:
I attempted to reproduce your issue on a fresh install of Fedora 31, and after updating and installing the developer tools, as you mention, the Sample Project runs even after closing and launching GB Studio a few times. Are you still having this issue and if you are can you provide some additional information?

terminaltux:
Thanks for looking into it @zdurgan . Unfortunately, yes I am still having this problem.

I think the real issue is, I haven't been able to find a way to get an error message in Fedora. At least with Ubuntu, it displayed an error in GB Studio, so I could find the solution. In Fedora, it just hangs. I've tried running GB Studio from the command line with:
```
gb-studio --help
gb-studio -help
man gb-studio
```
but it didn't provide any info other than:
```
Fontconfig warning: "/usr/share/fontconfig/conf.avail/05-reset-dirs-sample.conf", line 6: unknown element "reset-dirs"
No language pack for user setting, falling back to en
[...]

PR #420, opened 2020-04-30
Use streams instead of read/writes for copying files; may help with #398

asiekierka:
- [ ] Tests for the changes have been added (for bug fixes / features) <- Note that formulating a test for this scenario in particular may be a tad difficult; perhaps a test for copy() as a whole could be useful?
- [ ] Docs have been added / updated (for bug fixes / features) <- After the patch's fix is verified.

After debugging #398 , I found that the application was getting stuck during the recursive copying of build tools. Noticing that using a combination of fs.readFile/fs.writeFile is sub-optimal, I rewrote it to utilize streams instead - which also happened to fix the problem for me.

No. This PR should be invisible to the end-user.

### DIFF
--- a/src/lib/helpers/fsCopy.js
+++ b/src/lib/helpers/fsCopy.js
@@ -28,8 +28,16 @@ const copyFile = async (src, dest, options = {}) => {
       // Didn't exist so copy it
     }
   }
-  const fileData = await fs.readFile(src);
-  await fs.writeFile(dest, fileData, { mode });
+  await new Promise((resolve, reject) => {
+    const inputStream = fs.createReadStream(src);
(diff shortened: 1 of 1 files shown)


chrismaltby:
Really weird issue, I've been trying to reproduce it today in VMs with no luck. Your change doesn't break anything for me though so if it's fixing the issue for you I'll merge and hopefully that will help with the issues seen in #398 
Thanks!

asiekierka:
My Linux setup is a bit atypical, which sometimes helps catch odd issues. To clarify: I managed to get a largely consistent reproduction where the code would never return after the fs.readFile() - it might be an issue with fs-extra's patches, it may be something with my setup, Node version, etc.

Issue #419, opened 2020-04-29
Sprites not displaying properly 


tofa2403:
Sprites are not displaying properly since I had an error occur involving bank 59, it said something about the initialiser, unfortunately I did not write it down, I restarted GB Studio and the error didn't appear when I compiled the game, however sprites were appearing incorrectly, either facing the wrong way or as a completely different sprite, when interacting with certain sprites they changed again to a completely different sprite. 

**To Reproduce**
compile the game and start in scene 351, the issue was also occuring in scene 346, other sprites in different scenes were also effected but had been fine days before.

**Expected behavior**
I expected the sprites to display properly, as you can see in the screen shots, in the project view, the sprite is a lady sitting down at the table, in game view the sprite is completely messed up

**Screenshots**

**Platform (please complete the following information):**
WINDOWS 10
![bug](https://user-images.githubusercontent.com/64530257/80586725-48465c00-8a0d-11ea-9425-8cbfa0e3c154.jpg)
![bug2](https://user-images.githubusercontent.com/64530257/80586730-4a101f80-8a0d-11ea-8ebe-7194974fe616.jpg)

**Additional context**

tofa2403:
I have also noticed that some triggers are pointing to the completely wrong background now too

tofa2403:
I have done some tests and the problem seems to be with the number of unique backgrounds that are included in the game, is there a way to increase the limit ?

tofa2403:
I have also discovered that the sprite problem is not related to the background issue, there seems to be a limit on the number of unique sprites too, if I add a new unique sprite, it breaks the game but if I change that sprite to an existing one it is fixed, would be great if this limit can be increased too

chrismaltby:
Hi @tofa2403 you were right that these were two different issues you hit at around the same time! There was a fixed limit of 256 unique sprites and while you should have been able to have over 256 unique backgrounds there was a bug that was causing some backgrounds to not load correctly once that amount was reached.

I've got a fix for both of these issues in #421 but I just need to do some more testing before I can merge it into the development build.

Here's how it's looking for me now:

<img width="592" alt="Screenshot 2020-05-02 at 21 42 32" src="https://user-images.githubusercontent.com/16776042/80892447-17428180-8cc2-11ea-9418-6f9e83e395b3.png">
[...]

PR #421, opened 2020-05-02
Fix/sprites not displaying

chrismaltby:
Fixes for the issues seen in #419 where sprites and backgrounds were sometimes corrupted.

A project referenced in #419 came across two issues caused by errors in handling large projects.

1. **Sprites were appearing corrupted in some scenes**
  This was caused by an inbuilt limit of 256 unique sprites in GB Studio, though if you went over this limit there were no warnings and the sprites would appear either using incorrect data or corrupted data.

2. **Some scenes were loading incorrect background images**
 While GB Studio did have support for over 256 unique background images there was a bug with the function `ReadBankedUWORD` which causes it to sometimes returns an incorrect value (I believe it's when the value isn't aligned correctly in memory)

I have added support for theoretically 65,536 unique sprites and have fixed the issues with large numbers of backgrounds again giving a limit of 65,536, though I'm sure the ROM would run out of space long before these limits are hit.

It shouldn't, I need to do some more testing on a range of projects before merging though

* **Other information**:

### DIFF
--- a/appData/src/gb/include/Scene.h
+++ b/appData/src/gb/include/Scene.h
@@ -40,7 +40,7 @@
 extern UINT8 scene_bank;
 extern POS map_next_pos;
 extern VEC2D map_next_dir;
-extern UBYTE map_next_sprite;
+extern UWORD map_next_sprite;
 extern ACTOR actors[MAX_ACTORS];
 extern TRIGGER triggers[MAX_TRIGGERS];
 extern UWORD scene_index;
--- a/appData/src/gb/src/Scene.c
+++ b/appData/src/gb/src/Scene.c
@@ -20,7 +20,7 @@ void SceneRenderActor_b(UBYTE i);
 
 POS map_next_pos;
 VEC2D map_next_dir;
-UBYTE map_next_sprite;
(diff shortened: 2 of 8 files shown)


chrismaltby:
In my first round of testing I noticed two more related issues that you probably would have come across soon @tofa2403

1. Setting the player sprite was only allowing the first 256 sprites to be used

2. Saving / Loading the game was only working for the first 256 scenes

I've pushed a fix for both issues though the fix for 2 could potentially cause corrupted game save data for anyone who is in the process of making their game.

Save data was previously stored in RAM as
0: is_save_data_stored
1: current_scene_index
2: player_x
3: player_y
4: player_direction
5: player_sprite_index 

and in this PR is now stored as
0: is_save_data_stored
1: current_scene_index_high_byte
2: current_scene_index_low_byte
3: player_x
4: player_y
5: player_direction
6: player_sprite_index_high_byte
7: player_sprite_index_low_byte

Saving in game will cause the corrupted data to switch to the new format though.

I'll continue testing incase I've missed anything else like this.

tofa2403:
This is great !! Thank you so much for your help @chrismaltby

PR #477, opened 2020-08-26
Add 'Clamp' option in Add/Subtract.

Jarod-Lee:

Feature: Allow add and subtract with no 'wrap' but 'clamp' from 0-255.
If 'Clamp' option is true, the sum will be up to 255, and the difference will be down to 0.
If 'Clamp' option is false, nothing changes.
I think it's handy in some combat system or loot system.

255 + 1 = 0
0 - 1 = 255

If 'clamp' is true:
255 + 1 = 255
0 - 1 = 0

No

* **Other information**:
L10n not completely yet. The 'Notes' text need to be update.

### DIFF
--- a/appData/src/gb/src/core/ScriptRunner_b.c
+++ b/appData/src/gb/src/core/ScriptRunner_b.c
@@ -102,8 +102,8 @@ const SCRIPT_CMD script_cmds[] = {
     {Script_MathMul_b, 3},             // 0x37
     {Script_MathDiv_b, 3},             // 0x38
     {Script_MathMod_b, 3},             // 0x39
-    {Script_MathAddVal_b, 0},          // 0x3A
-    {Script_MathSubVal_b, 0},          // 0x3B
+    {Script_MathAddVal_b, 1},          // 0x3A
+    {Script_MathSubVal_b, 1},          // 0x3B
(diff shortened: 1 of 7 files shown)


RichardULZ:
Yeah, that makes sense, would certainly help a lot with ease of making a battle system.

chrismaltby:
Nice one!

I need to have a think on the wording and I guess the line `Note: Variables have max values of 255 and will wrap if increased above 255 or below 0.` isn't relevant when clamp is enabled, maybe that line needs changing too. Let me have a think and come back to this.

chrismaltby:
I've just pushed some updates on the wording and made an alternate version of the notes text that only appears when using add or subtract with clamp disabled. I've also removed the note from set/divide/modulus since it isn't really needed in those cases. I made clamp being on as the default for new events since I think it's probably the safest option and matches how the standalone increment/decrement events work.

**Subtract with clamp (no note)**
Screenshot 2020-08-27 at 20 23 18

**Subtract without clamp (new note)**
Screenshot 2020-08-27 at 20 23 29

**Multiply** (previous Note)
Screenshot 2020-08-27 at 20 23 56

**Other operations (no note)***
Screenshot 2020-08-27 at 20 23 41

Thanks for implementing this 👍

Issue #458, opened 2020-08-22
Player sprite sheet doesn't revert to default when changing to a new scene


m-masaa:
**Describe the bug**
As of version 2.0, when "Set Player Sprite Sheet" is used to modify the player character, the change is kept in subsequent scenes.
This is problematic when changing the player to a cursor in a Menu for instance as it will remain that way once exiting it and returning in-game.

**To Reproduce**
1. Set a Sprite Sheet to the player using the "Set Player Sprite Sheet" event. (Sprite Sheet must be different than the default)
2. Switch to a different scene
3. See error

**Expected behavior**
The starting Player Sprite Sheet should be used when changing to a new scene.

RichardULZ:
I don't believe this was a thing in 1.2.1
This would break a lot of uses where someone wants to use a scene to change the character you play as.

Just set it to the correct sprite on your way out of the menu scene, i think we only store/restore the active player sprite for save/load events.

Jeltr0n:
My project from 1.2.1 relies very heavily on the player sprite remaining the same between scenes. It seems like it's more inconvenient to re-set the player sprite on every screen than it would be to just set it back to what you want after a menu screen.

chrismaltby:
This is matching the previous behaviour in 1.2.1 but I'd have to agree I had similar issues making the new sample project where I wished sometimes it worked the other way especially when switching between genres.

I could maybe add a checkbox like we have now on the "Attach Script" event to choose which behaviour to use

<img width="357" alt="Screenshot 2020-08-24 at 19 04 34" src="https://user-images.githubusercontent.com/16776042/91080049-d2f63b00-e63c-11ea-9649-f548db79fb20.png">

Jeltr0n:
That sounds great! Would it have to store the old sprite sheet or would it revert to the starting sprite sheet when restoring it though?

m-masaa:
I think my issue is mostly when restoring a previous scene from the stack. In that case, wouldn't it make sense to also revert the player character to the previous sheet?

> I could maybe add a checkbox like we have now on the "Attach Script" event to choose which behaviour to use
This at least sounds good to me for normal scene changes.

Jeltr0n:
[...]

PR #481, opened 2020-08-26
Feature/player sprite persist

chrismaltby:
Feature: Adds the ability to choose if a change to the player sprite persists between scene changes. Mentioned in #458

Internally GB Studio stores the current player sprite index as `map_next_sprite` which is used on every scene change to make sure the correct player sprite is used. Each time a call to the "Set Player Sprite Sheet" happens `map_next_sprite` is updated to the new value so when changing scene the new player sprite will persist.

I've added an additional checkbox matching the one on "Attach Script To Button" which allows this behaviour to be configurable per event.

Screenshot 2020-08-26 at 18 56 11

Unticking this will change the current player sprite but will leave the current value of `map_next_sprite` as is. 

This is useful for scenes where you only temporarily want to change the sprite such as the cursor on the "Player's House" scene or the ship on the "Space Battle" scene where previously a second event was needed to restore the player's sprite when leaving these scenes they can now be set as temporary changes.

Any projects already migrated to 2.0.0 will not migrate their events to have the new persist field and will need manually updating. Newly migrated projects will have the value set to true automatically to match the old default.

* **Other information**:
Should the scene stack also be keeping track of this value?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add additional Animation Speed option "None". Setting this will prevent actor from animating at all. Static actors will now cycle through frames while moving unless this value is set.
 - Add support for macOS full screen mode
 - Add menu item and keyboard shortcut to switch project, opening the recent projects list [@patrickmollohan](https://github.com/patrickmollohan)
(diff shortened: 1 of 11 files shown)


RichardULZ:
That seems like an ok compromise, still wonder if we need a 2.0b1 migration to 2.0b2 or rc, as this will likely break any migrated project that used set player sprite.

chrismaltby:
Agreed, I've just pushed a change which will silently migrate beta1 projects to have this change

Issue #450, opened 2020-08-22
[BUG] PNG color corruption after editing.


PrismaticRealms:
**Describe the bug**
After modifying one of the background PNGs, GB Studio refreshed the updated image but with messed up colors.  (Color bars appearing randomly in the background image.)

**To Reproduce**
1. Create new project based on Sample Project template.
2. Scroll down so that Player's House scene is in view.  Zoom in if necessary.
3. On a Mac, open up Pixelmator Pro (or your favourite image editor but that's what I used.  I also tried with Preview.) and open up room.png background image.
4. Widen canvas to 200 pixels.  Fill in the empty spaces with one of the colors in the background image and save.
5. Return to GB Studio and notice how the Player's House scene now shows the new widened image, but several random blocks in various colors have appeared.
6. Run the game, and notice that the scene displays using these messed up colors.
7. Return to GB Studio and click on the scene.
8. Change the background from room to some other image (cave, for example) and then change it back to room.
9. Notice how the scene refreshes and looks OK now.
10. Run the game again and notice that the scene looks OK here too.

**Expected behavior**
After modifying the background image, the scene in GB Studio should update showing the correct colors.

**Screenshots**
<img width="647" alt="Screenshot" src="https://user-images.githubusercontent.com/56090694/90948707-6d713700-e40f-11ea-9df4-a5ab85f33662.png">

**Platform (please complete the following information):**
 - OS: macOS 10.15.6
 - App version: 2.0.0-beta1

**Additional context**

RichardULZ:
This can be closed :)

PR #485, opened 2020-08-27
V2 Fix, when a scene has changed size, clear colours like collisions

RichardULZ:

Bug fix, colours were not cleared like collisions if a scene is resized.
Easy way out, no dynamic padding to try and keep data.

Color data is garbled after scene resize #450 

Color data is cleared after scene resize.

Nope

* **Other information**:
Migrating pull requests over to main repo.


chrismaltby:
Perfect, must have missed this!

I might have a look at getting it to keep the previous data and pad out the array, I guess that only works if they expanded/contacted the scene to right to the bottom of the image though

Issue #454, opened 2020-08-22
Tile limit overflow problem


hadrianlin:
When a scene goes over the tile limit, it appears fine if no dialog box appears. However, the next scene in which a dialog box appears, it is garbled, even if the new scene is under the tile limit. The previous version of GB Studio did not have this problem.

Windows 10 64-bit. Gb Studio 2.0

PR #486, opened 2020-08-27
V2 Fix, Refresh UI before Scene switch (in case last bg 192+ overwrote it)

RichardULZ:

Bug fix

Unlike 1.2.1, if you load a background between 192-256 tiles, then a background with less than 192 tiles, the text box will be using garbage tiles from the last bg. #454 

Reverts to 1.2.1 where UI tiles are loaded every time before the next scene, in case they were overwritten.
Dialogue still corrupt on scenes 192-256, but fine again once you move to a sub 192 unique tile map.

Users must refresh their engine if ejected for this change to take effect.

* **Other information**:

### DIFF
--- a/appData/src/gb/src/core/Core_Main.c
+++ b/appData/src/gb/src/core/Core_Main.c
@@ -227,6 +227,7 @@ int core_start() {
     BGP_REG = PAL_DEF(0U, 1U, 2U, 3U);
     OBP0_REG = OBP1_REG = PAL_DEF(0U, 0U, 1U, 3U);
 
+    UIInit();
     LoadScene(current_state);
 
     // Run scene type init function


chrismaltby:
Ah oops, thanks for fixing.

I'd suggest I look at preventing the image tiles from even overflowing past 192 but I guess people are probably depending on that functionality now for more complex images when the UI isn't needed

pau-tomas:
Yes, there's projects using the 192 overflow for title screens and cutscenes. 

If we do introduce a hard limit I think we should also offer a (advanced) setting to control tile distribution per scene.

chrismaltby:
Okay no problem I think for now we can leave it as is

Issue #484, opened 2020-08-27
Random Dialogue Box is presented when B is pressed


TheManTheGifTheLegend:
**Describe the bug**
In this code, when pressing "B", a random piece of dialogue will pop up. This was, in 1.2.1 regular build, a random piece of dialogue from anywhere else in the game, but in the dev build it is now random characters (see screenshots)

**To Reproduce**
1)run project
2) avoiding all triggers apart from scene changes necessary, walk Left, left, up, left, on the map (changing scenes on the leftmost part of the screen twice, then topmost, then leftmost), staying on the road as that is where the scene changes are.
3)Walk onto the door of the central-top of the 6 houses on the lower half of the scene
4)Press B

**Expected behavior**
The player sprite changes to a mouse cursor (player was hidden, hitting B increments value $11$ by 1, the script calls the change of player sprite based off of $11$, nowhere is a dialogue mentioned.

**Screenshots**
![image](https://user-images.githubusercontent.com/63155122/91370774-24eabc80-e807-11ea-8776-e05de88dcd54.png)

[ForGitHub.zip](https://github.com/chrismaltby/gb-studio/files/5133335/ForGitHub.zip)

**Platform (please complete the following information):**
 - OS: Windows 10 18363.1016
 - App version [latest Dev build as of 27/08/2020 1:52am]

RichardULZ:
Unfortunately, your project file didn't make it into the attached project zip, all other assets are present.
There is a built rom in builds, but that seems to be from 1.2.1, so won't help to reproduce the bug.

TheManTheGifTheLegend:
> Unfortunately, your project file didn't make it into the attached project zip, all other assets are present.
> There is a built rom in builds, but that seems to be from 1.2.1, so won't help to reproduce the bug.

I'll try again, sorry.

TheManTheGifTheLegend:
[ForGitHub2.zip](https://github.com/chrismaltby/gb-studio/files/5133552/ForGitHub2.zip)
here is the new .zip with the .gbproj inside

RichardULZ:
https://discord.com/channels/554713715442712616/570925559346102273/748323358336090122 
Ahh, this is continued from discord, thank's for getting it into the issue tracker.

The issue is said to be present in 1.2.1, and develop, for your project.
Pautomas has suggested an undiscovered too many script ptrs, or text, But it could also be a simple oversight in scripting somewhere, deep in the project.

Found it,
[...]

PR #487, opened 2020-08-27
Don't pass labels definition to when compiling subscripts

pau-tomas:
Labels defined outside subscripts (like timer or input script) where being passed to the compile. That made possible to use go to those labels without a compile error, leading to broken game behavior (see #484 )

Bug fix

Labels defined outside subscripts (like timer or input script) where being passed to the compile. That made possible to use go to those labels without a compile error, leading to broken game behavior (see #484  )

Subscripts (like timer or input script) receive an empty list of labels, so the scope of go to can only be inside the script.

Maybe. If somebody is using the wider scope for labels in a way that doesn't break the game. But I don't think that's possible.

### DIFF
--- a/src/lib/compiler/compileEntityEvents.js
+++ b/src/lib/compiler/compileEntityEvents.js
@@ -32,6 +32,7 @@ const compileEntityEvents = (input = [], options = {}) => {
         ...options,
         output: eventOutput || output,
         branch: eventBranch,
+        labels: eventBranch ? options.labels : {}
       }),
   };
   const location = Object.assign(


pau-tomas:
@chrismaltby while investigating this bug I realized that https://github.com/chrismaltby/gb-studio/blob/375a77ee1813a340af3b5c73678989900f835739/src/lib/compiler/compileData.js#L101 doesn't do anything as `cmd.true` is an empty array (it should be `cmd.children.true`). 

Since the input subscript is already compiled in the `scriptBuilder` maybe all this can be removed?

chrismaltby:
Thanks @pau-tomas Looks like you're right about that extra compile, doesn't seem to do anything at all anymore. I'll remove it 👍

PR #488, opened 2020-08-27
Fix/platformer collision

RichardULZ:

Bug fix

Jumping while pressing left into a wall would not let you jump
Ladders would work almost 2 tiles off to the right side of the ladder, but not at all left.
Jumping into the ceiling would snap the player down 1 tile more than expected.

- [x] Jumping while walking left in a wall now works the same as right.
- [x] Ladder collision now work 1/2 a tile to the left, (compromise between centered which was 1 tile each side but looked bad) This works great if your ladder collision is only on the left side of a 2 tile wide ladder. This still seems preferable over the original implementation.
- [x] Ladder collision is detected and used from the same coordinates, making it more reliable.
- [x] Jumping into the ceiling won't snap you down 1 extra tile
- [x] Jumping into the ceiling while running is smooth by checking the tile 1 *pixel* lower for left right collision tests.

Users with an Ejected Engine must re eject, ladder solution inelegant, but better than before.

* **Other information**:
**This does not fix Actor interactions being more difficult from the left, related to platformer using 1 tile center, and other modes using 2 tile center for actor facing.**

### DIFF
--- a/appData/src/gb/src/states/Platform.c
+++ b/appData/src/gb/src/states/Platform.c
@@ -52,7 +52,7 @@ void Start_Platform() {
     player.rerender = TRUE;
   }
 
-  tile_x = DIV_8(player.pos.x);
+  tile_x = DIV_8(player.pos.x)+1;
   tile_y = DIV_8(player.pos.y);
 
   grounded = FALSE;
@@ -72,7 +72,7 @@ void Start_Platform() {
 }
 
 void Update_Platform() {
-  UBYTE tile_x, tile_y;
+  UBYTE tile_x, tile_x_mid, tile_y, tile_y_ceil;
   UBYTE hit_actor = 0;
   UBYTE hit_trigger = 0;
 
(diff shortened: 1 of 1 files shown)


RichardULZ:
Attached gifs, if the bg ladder is 2 tiles wide, and ladder collision 1 tile wide, left aligned, looks great, still a bit inelegant.

 PlatformerUpdate2 image

chrismaltby:
This is all much better, thanks @RichardULZ 

I'm wondering if could maybe just detect the two ladder tiles next to each other during compile and ignore the right handside ones now 🤔

RichardULZ:
I probably wouldn't, removes / obscures some control from the editor. 
I may take another crack at it some time, perhaps we should be checking 2 tiles for ladders and subtracting somehow. If someone made a tiny ladder, it would appear a little odd.

But more pressing is Actor Interact being far out in a direction, i think we need a seperate actor infront of player test, as it has a large offset for actors being 2 tiles wide.

Issue #492, opened 2020-08-27
[Beta] "Enable Color Mode?" dialog can be hidden, appears to "freeze" the editor


DeerTears:
**Describe the bug**
Clicking on "Colorize" brings up the "Enable Color Mode?" dialog as expected. If you use Alt + Tab or click on the editor in the back, the dialog is temporarily hidden, but control is not fully restored to the editor. The editor is even able to keep track of all the requests to enable Colorize mode, resulting in a hilarious train of dialog boxes that keep popping up no matter how many times you accept/cancel the dialog.

**To Reproduce**
1. Create a project that does not have color mode enabled
2. Click on "Colorize"
3. Press Alt + Tab or click on the background window
4. Editor will appear to be frozen until the dialog is found again and resolved
5. The editor is also still receptive to further requests to enable colorize mode, stacks these requests and displays them after the current dialog is resolved

**Expected behavior**
Clicking the editor or attempting to do actions on the editor prevents me from hiding the dialog that's freezing the editor, making sure I take action on it before continuing. I expected something similar to the "Do you want to save the changes you made?" dialog when you try to close the editor with unsaved changes.

**Platform (please complete the following information):**
 - OS: Windows 10 64 bit
 - App version 2.0.0 Beta

**Additional context**
This is just an ease of use thing that will prevent headaches like I just experienced, when I was unsure if I wanted to enable color mode on my project. I hesitated to add color mode because I wasn't sure if adding it would kill compatibility with classic GB emulators/systems, which is a reassurance I'd appreciate also being added to the "Enable Color Mode?" dialogue if possible.

chrismaltby:
Thanks for spotting this @DeerTears it's weird because macOS was treating that window as a modal and not allowing the main window to gain focus or accept clicks but I've found a fix which I've implemented in #493
I also improved the wording a little, hopefully should be clearer now!

PR #493, opened 2020-08-27
Fix/color dialog modal

chrismaltby:
Fix: As noticed in #492 "Enable Color Mode" modal was not acting as a modal correctly in Windows and allowed some interaction with main window while opened.

When the modal was opened clicking the main window in Windows would give the window focus, while the UI would be locked it would register clicks that would stack up and all be fired after the modal was closed.
Found a similar issue listed here https://stackoverflow.com/questions/46327538/electron-dialog-box-not-preventing-interaction-with-page

As with the answer from the above link the solution is to use the optional browserWindow argument of the Electron function `dialog.showMessageBoxSync` passing in the main window which forces the dialog box to act as a modal correctly.

I've also changed some of the wording in response to #492 to make it clearer that GB support won't be lost and that you can revert the change later in Settings.

Screenshot 2020-08-27 at 21 05 08

On macOS the dialog box was already acting as a modal and this new change alters the styling to be a sheet rather than a separate window.

Screenshot 2020-08-27 at 21 14 03

No

* **Other information**:

There's a few other modals in `src/lib/electron/dialog` I'll have to go through these and determine which others may need the same update making

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -535,7 +535,7 @@
   "DIALOG_EJECT_ENGINE_REPLACE_DESCRIPTION": "You have already ejected the GB Studio engine to your project.\n\nIf you continue the previously ejected files will be replaced and any changes you made will be lost.",
   "DIALOG_EJECT_REPLACE": "Replace Engine",
   "DIALOG_ENABLE_COLOR_MODE": "Enable Color Mode?",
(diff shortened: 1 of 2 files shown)


DeerTears:
Amazing! Bravo for jumping on this so quickly

PR #496, opened 2020-08-28
Add support for property fields in Custom Events

pau-tomas:

Adds support for Property fields in unions to Custom Events.

Property fields aren't supported

When property is selected in an union field inside a Custom Event a selector appears that allow to choose between the properties of the Custom Events actors (A to J) and the Player. If an actor that isn't the player is selected it gets added to the parameter list.

No

* **Other information**:

I think I tested most of the cases (`$self$`, copy and paste events with actors, etc.) but would be good if somebody else run some more tests.

### DIFF
--- a/src/components/forms/CustomEventPropertySelect.js
+++ b/src/components/forms/CustomEventPropertySelect.js
@@ -1,3 +1,226 @@
-import React from "react";
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import Select, { components } from "react-select";
+import ActorCanvas from "../world/ActorCanvas";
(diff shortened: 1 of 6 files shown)


chrismaltby:
Thanks @pau-tomas it works great. Everything I've tested in game works as expected.

I spotted a small issue with the dropdown indicator causing a full rerender of ActorCanvas each time you hovered on/off the event. It's weird because you're using pretty much the exact same code I did to stop the flickering on the regular property select and I managed to fix it by switching to the version you used in CustomEventActorSelect 🤷 I'm going to do a pass through all the components at some point and check they rerender only when they really need to and try and simplify or split them out into share components where possible.

flicker

Issue #451, opened 2020-08-22
Passive collisions do not work on top down 2D scenes


ultimate-steve:
**Describe the bug**
In Platformer scenes, if the player stays still and an enemy collides, the collision event is triggered. However, if you change the scene type to "Top Down 2D", the collision only triggers when the player tries to move towards the enemy.

**To Reproduce**
1. Create a "Top down 2D" scene
2. Create an actor and set a basic update script to move towards the player
3. Configure an "On Hit" action
4. Test the game and verify that the collision doesn't trigger when the player stays still.

**Expected behavior**
The collision event should fire

**Platform (please complete the following information):**
 - OS: Arch Linux 5.5.13 x64
 - App version 2.0.0-beta1

**Additional context**

raulriera:
Adding my own comment on the issue as well https://discordapp.com/channels/554713715442712616/585023243732123676/748012165436276807

@pau-tomas, do you have a PR up? I could eject my engine and test your fix on my game

jessemillar:
This appears to have come back since @pau-tomas's PR was merged. Passive collisions as described in the initial post do not work in v3.1.0 for top-down or adventure scenes based on my testing.

RichardULZ:
@jessemillar just checking, this is with passive collisions from the player to actors? 

Actor to actor collisions have never worked as it's exponentially more checks, and with scene collision enabled on move events, actors don't go inside the player to cause hit events.

jessemillar:
@RichardULZ Correct. I have an enemy actor and when it goes inside the player it doesn't trigger collision but when the player attempts to move into (or move while the enemy is already inside the player), the collision triggers. I've fixed the issue in my game by using the "is player within x tiles from enemy actor" action after each movement from the enemy actor.

PR #497, opened 2020-08-28
Add actor-player collision in topdown when player not moving

pau-tomas:

Add passive collisions to top down as mentioned in #451 

In TopDown mode actors don't collide with the player when it's not moving. Other scene types like Platformer or Adventure support passive collisions.

If an actor collides with a player, even if the player isn't moving `On Player Hit` and the Actor `On Hit > Player` are triggered.

Maybe. Depends if the lack of collision is important for your project.

* **Other information**:

The solution adds feature parity between TopDown and the other scene types in passive collisions so it should be good to merge. However, while investigating I realized a couple of inconsistencies happening in all types:

- The collision only works when the actor is coming from left or right but not from up or down. 
- Passive collision works if the Actor is moving using `Move To` with `Use collisions` disabled.

I've been using this project to test: [GBS Showcase.zip](https://github.com/chrismaltby/gb-studio/files/5143616/GBS.Showcase.zip).

### DIFF
--- a/appData/src/gb/src/states/TopDown.c
+++ b/appData/src/gb/src/states/TopDown.c
@@ -101,6 +101,14 @@ void Update_TopDown() {
       player.hit_actor = hit_actor;
     }
 
+    hit_actor = ActorOverlapsPlayer(FALSE);
+    if (hit_actor && hit_actor != NO_ACTOR_COLLISON && player_iframes == 0) {
+      if (actors[hit_actor].collision_group) {
+        player.hit_actor = 0;
+        player.hit_actor = hit_actor;
+      }
+    }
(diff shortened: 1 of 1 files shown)


chrismaltby:
Thanks @pau-tomas I need to investigate why the collisions don't work from up or down :-/
> Passive collision works if the Actor is moving using Move To with Use collisions disabled.
I suppose this is because with collisions enabled the actor detects the player is there and stops before overlapping, maybe use collisions could optionally ignore the player... sounds like the UI might get a bit messy though

RichardULZ:
Time for an enable / disable collisions event overhaul?

Issue #517, opened 2020-09-02
Issues


elihaun:
Hello. I am having trouble getting the program to open. I can select the name of the new game, pick a template, etc. but if I hit 'create', it gets stuck on 'creating' forever.  Also, when i change the path, it takes ages to actually change the path.

RichardULZ:
Duplicate of #448

Issue #467, opened 2020-08-25
Cannot create new project in 2.0 beta 


triptych:
**Describe the bug**
Starting GBStudio and trying to create new project results in error screens
**To Reproduce**
Launch GBStudio 2.x beta
click "new"
type in a project name like "foobar001"
keep the default template
keep the default path

**Expected behavior**
Expect a new game window to form with the default template.

**Screenshots**

![gb-studio_StXEUMWX2Z](https://user-images.githubusercontent.com/8691/91125399-83246d80-e656-11ea-800c-fdf834d78aa3.png)
![gb-studio_C7dGp83QCI](https://user-images.githubusercontent.com/8691/91125394-828bd700-e656-11ea-8e9c-e79fc46e1141.png)
![gb-studio_n5YRk7NrFk](https://user-images.githubusercontent.com/8691/91125396-828bd700-e656-11ea-853c-9722119fa56f.png)

![gb-studio_tynGZ277oI](https://user-images.githubusercontent.com/8691/91125400-83246d80-e656-11ea-89e7-d5c55231de74.png)

**Platform (please complete the following information):**
 - OS: Windows 10 64bit
 - App version 2.x beta

**Additional context**

The very first time I created a project it worked. But never again after.

RichardULZ:
Are you running as administrator? That's my only guess right now 🤔

lynxieles:
I have this exact issue on Mac 10.15.4 Catalina on an administrator account.
The version 2.0 beta will not create a new game but 1.2.1 will gladly create a new project using all the same variables (location path, name, etc). will occur with all game types.

RichardULZ:
@lynxieles can you confirm the project path exists when you create a project? 
As there are strange new chromium electon bugs with selecting a folder path. Could be related somehow.

lynxieles:
I can confirm that the folder path exists as I selected it via the browse option.
I did manage to get it to start working if I save only to the default documents folder. I had to restart it to get it to agree to even save there however.

RichardULZ:
@lynxieles When you select the path name via the browse option, is the path name the same as the one you selected? or different/random name of a file? This is the known issue.

lynxieles:
I think it's selecting random file names. I tried it and nabbed a screenshot. it's selecting the .ds_store hidden file on my Mac. It creates fine if I delete the offending addition. I also found that if I delete the addition I have to either copy paste, close and reopen or manually type in a new file path if I want to select a different location.
[...]

Issue #448, opened 2020-08-21
Unable to Create Project


candicetate:
Hello! I just downloaded the 2.0 beta, and I came across a bug. 

When I try to select a folder for a project, the path of the project either doesn't change or isn't correct. The first time I tried, it keeps selecting files within the folders I selected (see video). When I went to take the video, I wasn't able to change the path at all. When I hit the create button, it says creating for maybe 5-10 minutes, but no folder is ever created.

**To Reproduce**
1. Open GB Studio 2.0
2. Click on New
3. Choose a path for the project
4. Click Create

**Screenshots**
Here's a video: https://monosnap.com/file/K649mXvpIbDtVCBjomyawyjZ8P0lRQ

**Platform (please complete the following information):**
 - Windows Version 1903 (OS Build 183621016)

PR #522, opened 2020-09-05
Fix Project path selector to select empty folder, not contents of folder

RichardULZ:

Bug fix

#448 #467 #517 Users can not select a project path or are stuck building a project indefinitely.
Something about the tools changed causing webkitdirectory to provide an array of the file contents of a folder, and not the folder itself.

Replaces default file dialogue with 
electron.remote dialouge.showOpenDialogue({ properties: ['openDirectory'] });
Which returns only the path of the selected folder, and is much faster than the method before it was broken.

* **Other information**:
After days of searching, finally found the correct method
https://stackoverflow.com/questions/36152857/how-to-get-folder-path-using-electron
Figured out the await / async stuff from surrounding code myself :)

### DIFF
--- a/src/components/app/Splash.jsx
+++ b/src/components/app/Splash.jsx
@@ -11,6 +11,8 @@ import createProject, {
 import l10n from "../../lib/helpers/l10n";
 import "../../lib/helpers/handleFirstTab";
 
+const {dialog} = require('electron').remote;
+
 const getLastUsedPath = () => {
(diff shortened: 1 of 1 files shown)


chrismaltby:
Oh yeah I'm seeing that now too... I've always just been using the saved path of my documents directly and never noticed the Electron API changed there.

Had quite a few people mention this to me in different channels so lots of people should be happy about this fix! Thanks @RichardULZ

PR #529, opened 2020-09-06
Feature/redux toolkit typescript

chrismaltby:
Refactor: Replaces the entire Redux store with a Typescript replacement using [Redux Toolkit](https://redux-toolkit.js.org/) removing a lot of the Redux boilerplate and improving type safety going forwards.

The store is using plain Redux and is written in Javascript

The store has been migrated to Redux Toolkit / Typescript with the following changes

- folder `src/actions` has been removed
- folder `src/middleware` has been removed
- folder `src/reducers` has been removed
- `src/store/configureStore` has been migrated to Typescript and is using Redux Toolkit's `configureStore` helper.
- Redux store has been split by feature into `src/store/features/XYZ`. Each feature contains some combination of actions, state and middleware
- Previous reducers have been migrated to use Redux Toolkit's [createSlice](https://redux-toolkit.js.org/api/createSlice). As createSlice uses [Immer](https://github.com/immerjs/immer) internally reducer functions now use mutable style

e.g
```
    setBrush: (state, action: PayloadAction<{ brush: Brush }>) => {
      state.selectedBrush = action.payload.brush;
    },
```
rather than 
```
    case SET_SELECTED_BRUSH:
      return {
        ...state,
        selectedBrush: action.brush
      }  
```
which simplifies the reducers using large nested state 

- Entities are now being managed using [createEntityAdapter](https://redux-toolkit.js.org/api/createEntityAdapter) which is similar to the `add/edit/removeEntity` functions used previously but also generate selectors automatically.
- Each feature's actions are generated automatically from the slices and where a slice isn't required `createAction` and `createAsyncThunk` are used instead as needed.

Quite possibly, 130 files changed across 84 commits so far and most of the data store of GB Studio has changed in this PR so there's likely to be subtle changes in functionality here and there. I'm going to be done some more testing before merging and will be writing some more test cases.

* **Other information**:
I'll be migrating the React front end to Typescript also eventually though it's much easier to do that piecemeal with smaller commits.


chrismaltby:
I'm getting a merge conflict on entitiesReducer.js it's likely I didn't merge in your property fields for custom events fix @pau-tomas Will do a diff of the version I built this from against the `v2beta` branch tomorrow to make sure I migrate your fix across correctly.

RichardULZ:
Just checking, does this change plugins api? Sounds like it will be a lot cleaner to work on though.

chrismaltby:
Oh maybe, custom event plugins and asset plugins should be unaffected (though I need to test them) but it's possible that menu plugins would be affected if people are dispatching actions to the redux store directly from them. Those plugins are currently completely undocumented as I don't really have a stable API there yet. Do you have some idea of things people are doing with plugins and I can see what the migration path would be?

pau-tomas:
AFAIK I'm the only one that created some plugins and they're shared in Discord from time to time. Both of them are events: one for Menu without character limit and the other for If variable in range. The first one is probably the one used the most. 

I always warn people when they ask that plugins are undocumented and can break in the future, I don't know if any of them made into a released project.

chrismaltby:
Thanks @pau-tomas I think both of those should be fine then actually but do you mind posting them here so I can confirm?

pau-tomas:
[Here they are](https://github.com/chrismaltby/gb-studio/files/5184784/Plugins.zip). I also think both of them will work fine still.

chrismaltby:
Thanks Pau, yeah those plugins are still working in this version

um3k:
It seems menu plugins are broken in 2.0 beta. I tested the Add Boulder plugin in both 1.2 and the latest 2.0 build. It worked in the former but not the latter.

chrismaltby:
@um3k I don't know the "Add Boulder" plugin, could you zip it up and post it here, and I can investigate what's needed to fix it. Thanks

um3k:
Sorry, I guess it was called Make Boulder. It's yours, from year or so ago.
https://github.com/chrismaltby/gb-studio/files/3668009/pluginsTest.zip

chrismaltby:
Hi @um3k oops sorry I totally forgot I'd ever shared this!

There was a few things causing that plugin to break in 2.0.0 and in this branch some of the internal structures and functions have changed so some of the plugin code needs updating. With this branch currently I've got the plugin working if I update the code to the following. This is still a bit subject to change though. Eventually I'll need to work on making a stable API for menu plugins. 

```
export const id = "CM_ACTION_BOULDER";

export const name = "Make boulder";

export const accelerator = "CommandOrControl+L";

export const run = (store, actions) => {
  const state = store.getState();
  
  const spriteSheets = state.project.present.entities.spriteSheets.entities;
  const spriteId = Object.values(spriteSheets).find(spriteSheet => {
    return spriteSheet.filename === "rock.png";
  }).id;

  store.dispatch(
    actions.setActorDefaults({
      name: "Boulder",
      spriteSheetId: spriteId,
      script: [
        {
          command: "EVENT_ACTOR_PUSH",
          args: {
            continue: false
          }
        },
        {
          command: "EVENT_END",
        }        
      ]
    })
  );
};
```

PR #530, opened 2020-09-07
Fade to black, and palette fade optimizations/fixes

RichardULZ:

Bug fixes and features!

Fading is hard on cpu, causing many visual artifacts while fading due to calculating each colour and setting after many scan lines.
Setting a palette on scene change can cause the screen to flash a frame of garbage somehow, as palette data is set regardless of fade.
Fade to black has been requested for years, and only recently became viable.

Fade to white and palette calculation is massively simplified, focusing on bg tiles first and applying in bulk, there is now a tightly packed vertical tear near the top if white, or middle if black, instead of random tiles.
A new variable `fade_black` has been added to pick calculation methods, checks to keep screen on have been added, this works in both cgb and dmg thanks to music now being on a timer script, independent of cpu lag.

The fade to white pause has been removed (if index 0, index = 1) to simplify fade to black, however the fade might sometimes be considered "too fast" as a flash, not sure if this is better, or if this change should be reverted and added to the calculate white fade path.

* **Other information**:
Fade to black is not currently exposed as an event.
An event for `Set fade color` or `Set transition type` would be extremely easy, and allow a single event at start to change the fade mode, without editing each scene switch for white or black. This should not require migration, and should allow for a scene to fade to black, then fade in from white.
This could also be considered good to go without an event, as you only need to change 1 number in ejected engine to get fade to black.

Example where each vertical column on the left is a different palette, showing how long each takes in scan lines to update
Before 
image
After 
image


RichardULZ:
Fadetests
Test on Tofa's Oot demake with fade to black, (speed not accurate as a gif)
Note there is screen tearing in the center on gbc fade to black, could be solved by calculating palettes before vblank, and setting after vblank, but is already much better than any fade we had on gbc. 
Note that while Fade to Black is set, instant camera moves will also jump to black instead of white screen off.

chrismaltby:
Hi @RichardULZ I've just added a global setting on the Settings page to choose which fade type to use by default and an event to switch fade settings mid game. Let me know what you think.

I'm seeing the screen tearing too on fade to black but not fade to white (at least not noticeably) I wonder if it's anything to do with fade to white only doing RGB2 once before the loop

```
    paletteWhite = RGB2(DespRight(0x1F, index), DespRight(0x1F, index), 
                      DespRight(0x1F, index));
    for (c = 0; c != 32; ++c, ++col) {
      palette[c] = (UWORD)*col | paletteWhite;
    }
```

where fade to black is doing it every iteration in the loop (inside the call to UpdateColorBlack)
```
    for (c = 0; c != 32; ++c, ++col) {
      palette[c] = UpdateColorBlack(index, *col);
    }
    set_bkg_palette(0, 8, palette);
```

I need to figure out how this is actually working when I get a min to understand what's going on here.

RichardULZ:
Oh there's definitely still screen tearing on fade to white, just quite high up, (about 1-2 tiles from the top)
The solution would be calculating values, then waiting for vblank, then setting bgp at the start of vblank with the calculated values, that could add weird pauses, but a check at start of frame seems inelegant as well...
This needs to work for palette switching in game too, which probably screen tears differently too.

Fade to white is 10 times faster now, due to RGB2 being fairly expensive with all the variable juggling it has to do.
It now generates a mask once, and apply's the same white bit mask to all the rgb colours, can't do that with fade to black, as it's bit shifting the actual rgb values down, not just a simple mask.

chrismaltby:
Ah I see, I ended up making a little HTML/JS page to understand the maths but I see what you mean, I was hoping we could do something similar to the white mask but in reverse with AND rather than OR, it was faster but the colours weren't great and it still did the tearing.

So instead I've just pushed an update the uses another buffer for the current palettes and moved the `set_bkg_palette` and `set_sprite_palette` calls into the vblank interrupt that fire only if the palettes have changed. I might be speaking too soon but from what I can tell this seems to have removed the tearing!

This is slowed down gif of what I'm seeing in bgb now, fade to white is looking the same too
ezgif com-gif-maker

RichardULZ:
Yep, looks perfect.
I guess if palette dirty is only ever set by cgb exclusive commands, there's no point in doing a check for dmg/cgb, should be fine.

chrismaltby:
Just put in an engine changelog for this, all seems good to go so merged now! Thanks for your help! 👍

PR #537, opened 2020-09-15
Patch 1

laetemn:
Language update and Brazilian Portuguese translation corrections for GB Studio version 2 BETA.


chrismaltby:
You just had a missing quote in the file causing this file to not be able to load correctly but this was a quick fix which I've done in fe28e771cb0c4c1310f5134fd9c1bf857289f88f 👍

PR #548, opened 2020-09-19
Add Actor parenting

um3k:

New feature for setting a parent actor for a given actor, allowing them to move in unison.

Actors are entirely independent of each other, and there is no reliable means to 
make them move synchronously to simulate a larger sprite.

Adds a "Parent" field to all actors, which, when set, will make that actor move
synchronously with the selected actor, maintaining its relative position as saved
in the editor. It also matches direction and animation frame.

Possibly, it will certainly invalidate ejected engines. A default value for the parent field
has not been explicitly configured, but it seems to default to self anyways, which will maintain
existing behavior.

* **Other information**:

Current implementation is entirely in C, and results in a noticeable slowdown in DMG (non-color) mode. I think this could be remedied by migrating it to assembly code, but currently that is just beyond the borders of my skill set.

### DIFF
--- a/src/components/editors/ActorEditor.js
+++ b/src/components/editors/ActorEditor.js
@@ -32,6 +32,7 @@ import { actorSelectors, sceneSelectors, spriteSheetSelectors } from "../../stor
 import editorActions from "../../store/features/editor/editorActions";
 import clipboardActions from "../../store/features/clipboard/clipboardActions";
 import entitiesActions from "../../store/features/entities/entitiesActions";
+import ActorSelect from "../forms/ActorSelect";
 
 const defaultTabs = {
   interact: l10n("SIDEBAR_ON_INTERACT"),
@@ -445,6 +446,18 @@ class ActorEditor extends Component {
                 />
                 <div className="FormCheckbox" />
                 Pin to screen
+              </label>
+            </FormField>
+
+            <FormField>
+              <label htmlFor="actorParent">
+                <ActorSelect
+                  id="actorParent"
+                  value={actor.actorParent}
+                  //direction={argValue(args.direction)}
+                  //frame={argValue(args.frame)}
+                  onChange={this.onEdit("actorParent")}
+                />
               </label>
             </FormField>
 
--- a/src/components/editors/ActorEditor.js
+++ b/src/components/editors/ActorEditor.js
@@ -451,6 +451,7 @@ class ActorEditor extends Component {
 
             <FormField>
               <label htmlFor="actorParent">
+                Parent Actor
                 <ActorSelect
                   id="actorParent"
                   value={actor.actorParent}
--- a/appData/src/gb/include/Actor.h
+++ b/appData/src/gb/include/Actor.h
@@ -55,6 +55,7 @@ typedef struct {
   BankPtr hit_2_ptr; // 39
   BankPtr hit_3_ptr; // 42  
   UBYTE movement_ctx; // 45
+  UBYTE parent_actor; // 46
 } Actor;
 
 typedef enum {
--- a/appData/src/gb/include/ParentedActors.h
+++ b/appData/src/gb/include/ParentedActors.h
@@ -1,0 +1,8 @@
+#ifndef PARENTED_ACTORS_H
+#define PARENTED_ACTORS_H
+
+#include <gb/gb.h>
+
+void PositionParentedActors();
+
+#endif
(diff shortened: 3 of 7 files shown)


patrickmollohan:
Great addition! This would make for really fun boss fights. I didn't notice much slowdown on DMG mode (though maybe in addition to assembly, the newer gbdk would help?); however, when testing with my existing projects, the actors seem parented to the player, even though it shows "self".

um3k:
I see. I'll have to figure out why that is happening.

patrickmollohan:
I've noticed that it can be fixed by clicking the dropdown and reselecting "self" for every actor, so I think your code's logic is sound. Perhaps if the parent property doesn't exist on the actor, it's set to 0 (the player)? Perhaps a migration would be needed for existing projects.

um3k:
Yeah, that seems likely. I'll have to figure out how to make a migration
script.

src/components/editors/ActorEditor.js
```diff
@@ -448,6 +449,19 @@ class ActorEditor extends Component {
               </label>
             </FormField>
 
+            <FormField>
+              <label htmlFor="actorParent">
+                Parent Actor
```
pau-tomas, line 454: needs inernationalization

pau-tomas (commented):
Nice work! I really like this feature.

I tested it a bit and found a couple of issues:

- If the player has children actors to its right or bottom it can't move towards that direction.

actor_parent_move_bug

- If the parent and the children don't have the same amount of frames the children animation won't display correctly, showing empty frames (and, potentially, other actors frames depending on how the VRAM is allocated). Not sure how to solve other than maybe adding a warning at compile time, but that won't cover the cases where the spritesheet of the children or parent actor is updated via event to one with less or more frames as the related one.


I'd like to think more about how to better fit this into potential future features (like metasprites or prefabs). Don't want to corner us in a place where we can't easily migrate projects using this feature functionality to future ones, but also don't want to do any premature optimization as I like the simplicity of this solution.

chrismaltby:
Really like this feature, and love how simple the implementation is. Definitely interested in getting this in the app, thought might be more of a 2.1 feature. Looking good so far. I'm also getting the issue where actors are by default parenting to the player (though the dropdown is saying Self).

I think the dropdown probably just needs a few options to change it slightly for this use case:
- First value in the dropdown should be None and this should be the default
- Self, and the current actor shouldn't appear in the dropdown at all

I've been working on some wireframes for a little ui refresh at some point (they're very rough though), I could see this feature being quite prominent on the UI with a Pin button which could include the "Pin to Screen" functionality too. Keep the UI as it is for now but once I get around to making these updates I'll definitely look to make this working something like this wireframe.

Screenshot 2020-09-26 at 13 13 38

um3k:
Thanks for the notes, @pau-tomas and @chrismaltby. I'm pleased that you like the idea and implementation, I personally had my doubts, so it's reassuring. I've had the project on the backburner for the past week or so, but hoping to get back on it soon. I do have a few ruminations on it I'd like to mention.

1. Currently, it coopts start_position to store an actor's relative position to its parent. I don't think this will cause any issues, as start_position does not actually appear to be used anywhere else. However, it is an obstacle for, say, being able to move the child actor relative to its parent. In my mind, it would be better to store the relative position in the pos field, so that it would work natively with the Position and Move events, but that would necessitate applying the offset in assembly. I severely doubt my ASM skills are up to snuff for doing that. I don't really want to add another position field, since memory is at a premium.

2. I very much like that UI mockup. I'm wondering if it would be a good idea to rebrand this as "Pin to Actor", to sort of unify things. But then, does synchronizing things like direction and frame make sense? They are absolutely important parts of the feature. But I also wonder if they should be toggleable, because I can think of scenarios where you wouldn't want them linked.

3. I will try to wrap my head around how the ActorSelect UI code works so I can make the necessary modification.

4. The issue with the player not being able to move with sprites to the right, bottom, and left is due to the player colliding with the sprites. I'm not sure how to handle this. On one hand, I could leave it up to the end user to disable collision, though that might be confusing. I could perhaps disable collision specifically between child and parent actors? I'll have to look at the code.

5. Should it be possible to alter parenting via events? I don't think it would be terribly difficult to implement, once everything else is sorted out.

beatbesmer:
Will such an solution also allow to dynamically "attach" another sprite?

Imagine a player picking up an item (switch the player sprite to  the "pickup"- animation, move the object- sprite over his head [to position of player: x/y+2, just missing a "round" movement type]) and carrying it over his head bevor throwing it away and "detach" this relationship (... and yes, i'm referring to our green friend invented by Miyamoto, which picks up pots)...

um3k:
That functionality is not currently implemented, but should be fairly
trivial to add.

PR #556, opened 2020-09-21
Fix instant compile

RichardULZ:

Bug fix

Adding any new local variable will force the engine to recompile (60+Seconds), due to TMP_VAR_1 and 2 being indexed at the end of the list, modifying in Data_ptrs.h

INSTANT COMPILES EVERYWHERE!
TMP_VAR_1 and 2 are now added to variable list after the first 100 global variables, so will always be var 101-102, never causing an update when new variables are added.

Can't see how.

* **Other information**:
Thanks a ton to @pau-tomas For some super sleuthing while investigating the issue,
image

### DIFF
--- a/src/lib/compiler/compileData.js
+++ b/src/lib/compiler/compileData.js
@@ -666,6 +666,8 @@ export const precompileVariables = (scenes) => {
   for (let i = 0; i < 100; i++) {
     variables.push(String(i));
   }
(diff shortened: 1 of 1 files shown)


chrismaltby:
Hahahahahahahaha... oh... no... I can't believe that... Thanks both

PR #560, opened 2020-09-23
Fix platformer direction change rerender

um3k:

Fixes a bug where the player in platformer mode would not update its facing direction until a few frames after reversing its direction of movement.

If you walk in a direction in platformer mode, then quickly switch to the other direction, you character will start moving in the new direction, but it will take a few frames for the sprite to change the direction it is facing. This is due to a missing player.rerender = TRUE in the left and right input sections of the script.

The player sprite reverses direction immediately.

Not that I can imagine

* **Other information**:

### DIFF
--- a/appData/src/gb/src/states/Platform.c
+++ b/appData/src/gb/src/states/Platform.c
@@ -125,6 +125,7 @@ void Update_Platform() {
         pl_vel_x -= WALK_ACC;
         pl_vel_x = CLAMP(pl_vel_x, -MAX_WALK_VEL, -MIN_WALK_VEL);
       }
+      player.rerender = TRUE;
     } else if (INPUT_RIGHT) {
       player.dir.x = 1;
       if (INPUT_A) {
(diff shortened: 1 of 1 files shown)


RichardULZ:
Nice spot, Any idea on performance? 
Does this force it to re render the player sprite every frame the left or right button is held in platformer? Or just when you change from left-right? Can't think of many other ways.

um3k:
That's a good question, I hadn't really thought about that. I think it's as long as the button is held. Jump uses INPUT_B_PRESSED for an instantaneous action, maybe there's a counterpart for left and right I can use. On investigation, yes, those exist. I will try them and update if it works.

um3k:
Using PRESSED seems to work nearly as well, with the exception of when the player is holding right, presses left (which overrides it and reverses the direction), and releases left. Then you have the delayed update again. I doubt that would be much of an issue on real hardware though, since it's not as easy to press opposite directions.

chrismaltby:
This feels much better now, and using the INPUT_PRESSED macros should have very little performance impact. Thanks @um3k!

PR #572, opened 2020-09-27
Reduce GBSProj file size using Nano ID and zlib for tile data

chrismaltby:
This is an experimental change to reduce the file size of the GBSProj format.

Unique ids are stored as UUID v4 values and raw tile data is stored uncompressed in JSON

- All 36 byte UUIDs (given to every actor/scene/trigger/asset/event/etc) are replaced with 16 byte [Nano IDs](https://zelark.github.io/nano-id-cc/) which should allow someone to make a project the same size as the sample project every hour for 3 million years (dedication!) before they have a 1% chance of an id collision. This cuts all ids sizes down by over 50%.

- All collision and tile color data has been compressed using [zlib](https://nodejs.org/api/zlib.html) and stored as a base64 encoded string rather than a raw byte array in the JSON data. This gives a huge saving on the storage of these values e.g. the collisions in Sample Town from the sample project go from 60775 bytes down to 355 bytes (!!) and overall halves the file size of the sample project (1.2Mb to 591Kb).

It shouldn't, the compression happens just before saving and the decompression happens immediately after loading so very little code is touched in the change. Opening a project that isn't compressed falls back gracefully so you can still open the project but it will be compressed on next save.

If anyone was editing the GBSProj collison/tile data manually (maybe as part of some custom tools) they will need to first convert the stored base64 strings to a buffer and then uncompress them with zlib to get the previous format, and reverse this process before saving. 

Nano ids will only be used for newly created entities and can be used alongside the existing UUIDs in a project with no issues.

* **Other information**:
While there doesn't seem to be any issues with this update I'm absolutely not going to include it until at least a 2.1 release. The format once migrated will not be usable by older versions so I'd rather concentrate on getting 2.0 as stable as possible before introducing a change like this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "image-size": "^0.7.1",
     "json-loader": "^0.5.7",
     "lodash": "^4.17.14",
+    "nanoid": "^3.1.12",
     "normalizr": "^3.4.0",
     "prop-types": "^15.7.2",
     "raw-loader": "^4.0.0",
@@ -79,7 +80,6 @@
     "rimraf": "^2.6.3",
     "semver": "7",
     "ts-flood-fill": "^1.0.2",
-    "uuid": "^3.3.2",
     "vm2": "3.9.2"
   },
   "devDependencies": {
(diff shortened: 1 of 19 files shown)


Boundarybreaker:
Is this gonna be reimplemented in 3.x at some point?

mxashlynn:
I'm also curious to know about the future of this issue, if it has one.

Issue #569, opened 2020-09-27
Adjust actor limit counter from 10-30 based on scene size


RichardULZ:
**Is your feature request related to a problem? Please describe.**
Only stated in the changelog, new users sometimes expect their single screen scenes to allow up to 30 actors at once, where the limit is 10.

**Describe the solution you'd like**
Dynamically change the displayed actor limit between10 when a scene is 144-160, 360 tiles large, to 30 when over 1080 tiles? (3 screens tall?)

**Describe alternatives you've considered**
Figuring out if a scene has more than 10 actors in a screen width to warn? 
Or a simpler compile time warning? 

**Additional context**
I suppose advanced users could hide and show actors within a single screen, working around the 10 actor at once limit, but the limits being 30 per map, and 10 on screen is confusing to document either way.

maxoakland:
> I suppose advanced users could hide and show actors within a single screen, working around the 10 actor at once limit, but the limits being 30 per map, and 10 on screen is confusing to document either way.

I'm actually doing that in my game right now!

PR #575, opened 2020-09-28
Warn if actors are clustered too close together in a scene

chrismaltby:
Implements changes to scene actor warnings based on the enhancement request from @RichardULZ in #569 

Since 2.0.0 the number of actors allowed per scene has increased to 30 but the actual number allowed on screen at once is only 10 due to limitations of the hardware/tradeoffs in the game engine. This causes some confusion when more than 10 actors in a scene are placed too close together as some of them will not be visible while playing.

1.  Reduces the limit of actors in the scene actor count to 10 if the scene is the size of a single GB screen `160px x 144px` (although this is just a warning and can be ignored if needed for advanced use cases)

2. Introduces a new warning when over 10 actors are clustered together such that they would all be on screen at once 
Screenshot 2020-09-28 at 18 41 29

No

* **Other information**:

1. I'm not sure 100% on the wording of the message (and I've just noticed from looking at the screenshot I used "onscreen" on one line and "on screen" on another!) Anyone have any suggestions?

2. The calculation isn't 100% correct as actors don't immediately get removed when they go offscreen but it catches a lot of cases that would fail


pau-tomas:
Because some of those errors might not affect the game (if used "wisely") I'm wondering if the messages might be distracting for some users. At the same time it should be clear that they might have breaking consequences. I drafted a version of the error with a warning icon that displays/hides the error messages on click. This could be used for all the counters (and potentially have an icon for errors too). 

warning_scene

Also, maybe there should be some rules about the color applied to the entity counters:

- Orange + warning icon: over limit but things might still work (ie: more than 10 actors in a big scene)
- Red + error icon: over limit and things will break or igoner (ie: more than 30 triggers, or more than 30 actors in a big scene, more than 25 frames)

chrismaltby:
Yeah I agree about the error message, it's also a bit weird that there's already errors/more info that get displayed on hover in a tooltip (which also isn't localised I've just noticed). I think I'll combine all these together to do something like your screenshot 👍

chrismaltby:
I took your idea @pau-tomas and merged the existing tooltip text together into a single popup that appears either on click or after a short hover over any of the scene stats.

Rather than add an icon I've just set the background color of the stat which hopefully will be noticeable enough but not too distracting if you want to ignore it.

While making the change I also finally got around to localising the text too.

Screenshot 2020-09-30 at 20 25 59
Screenshot 2020-09-30 at 20 25 36
Screenshot 2020-09-30 at 20 25 31
Screenshot 2020-09-30 at 20 36 20

Issue #390, opened 2020-02-04
Selected Music tracks disappear if they update


RichardULZ:
**Describe the bug**
OpenMPT, like gbstudio, renames working files to .bak before writing an new file.
This short timespan where the music file is missing (renamed) can cause gbstudio to loose previously assigned music files, and requires you to re assign the track wherever it was used.

**To Reproduce**
1. Assign a music play event with a .mod
2. Rename the music .mod file
3. See the file selection disappear in music event.
4. Rename back to the original file name
5. Track remains empty in gbs event.

**Expected behavior**
Preferably the missing assigned tracks would turn red in gbs music event, and found again when a track is re added.

Else, a small time buffer before clearing tracks that have been removed.

**Screenshots**

**Platform (please complete the following information):**
 - OS: [ Windows 10 ]
 - App version: [1.2.1 (Issue since 1.0?)]

**Additional context**
GBstudio's `src/lib/project/WatchProject.js` seems the likely place this happens, but perhaps the fix would be best in the `music play event`, to store the selected song name and check for updates when missing, else how would you tell if a removed song was used previously at all.

You can work around this by saving .mod outside the project, and replacing the music file in windows manually, or reopening an older version of the gbsproj. In most cases this is slower than reassigning the track in gbstudio, unless it was used more than once.

This is a similar issue to #387 just without the crash ;)

Issue #387, opened 2020-01-27
Missing backgrounds cause editor crash on asset reload.


RichardULZ:
**Describe the bug**
Missing backgrounds cause editor to crash on asset reload, even if the assets have been fixed.

**To Reproduce**
1. Delete or rename a background
2. Asset is auto reloaded with blank white 256x256 square, (no path name).
3. Undo the delete / rename, outside of editor, or leave it missing (fix both cases)
4. Try to manually reload assets.
5. Editor crash with `"path" argument must be of type string. Received type undefined` `path.js:L39C:5`

**Expected behavior**
Background is found and reloaded nicely, or left missing, no crash.
Dimensions and collisions are retained while the file is missing (test if they persist through save), 
Actor positions are retained (currently fine, may break with #386).

Not important, perhaps text saying `Background "path/name" missing` inside the blank background?

**Screenshots**
![BackgroundMissingCrash](https://user-images.githubusercontent.com/34431457/73145421-5f299980-4112-11ea-8440-5f77771ce0f5.gif)
![BackgroundMissingCrash2](https://user-images.githubusercontent.com/34431457/73145565-1d4d2300-4113-11ea-852a-234fca961f6e.gif)

**Platform (please complete the following information):**
 - OS: [Windows 10]
 - App version [1.2.1 / dev]

**Additional context**
Similar to #386

RichardULZ:
This no longer crashes the editor, but will not find the background again once it is replaced, tough to fix.

Issue #574, opened 2020-09-28
When renaming a background it corrupts the project.


RedstoneGuyLol:
**Describe the bug**
When ever i rename one background, the background in the application turns black and some of the actors, collisions, and triggers change their X and Y positions

**To Reproduce**
1. Go to the project
2. Click on assets then click backgrounds
3. rename any background (it has to be used)
4. Go back to the GB Studio application 

**Expected behavior**
The background getting set to default.

**Screenshots**
![error1](https://user-images.githubusercontent.com/63218795/94467162-88fafa80-0190-11eb-8831-72e8ce2ab596.PNG)
![error2](https://user-images.githubusercontent.com/63218795/94467164-88fafa80-0190-11eb-8209-ef31843287f4.PNG)

**Platform (please complete the following information):**
 - OS: Windows
 - App version: 2.0.0 Beta

chrismaltby:
Hi @RedstoneGuyLol I've had an attempt at fixing this here #576 I just need to test this on all the different platforms and make sure it doesn't break anything but so far seems promising.

RedstoneGuyLol:
> Hi @RedstoneGuyLol I've had an attempt at fixing this here #576 I just need to test this on all the different platforms and make sure it doesn't break anything but so far seems promising.

Ok, i hope this will get patched soon.

chrismaltby:
Hi @RedstoneGuyLol I've merged this fix into the latest dev build and it's going to be included in beta4 :-)

RedstoneGuyLol:
> Hi @RedstoneGuyLol I've merged this fix into the latest dev build and it's going to be included in beta4 :-)

Thanks!

PR #576, opened 2020-09-28
Allow renaming project assets while preserving id

chrismaltby:
Attempt to fix issues seen in #574 where renaming assets causes ids to be regenerated and any scenes/events referring to those ids would be broken.

Asset file watching is handled in https://github.com/chrismaltby/gb-studio/blob/v2beta/src/lib/project/watchProject.js using the module [chokidar](https://github.com/paulmillr/chokidar) which handles the complexities of cross platform file watching. In chokidar when a watched file is renamed it fires an unlink (delete) file event followed by an add file event with no way to determine that this was a single rename action (and it looks unlikely that this functionality will change https://github.com/paulmillr/chokidar/issues/303). This caused a rename to first delete the asset, then it will be readded, at which point it gets a new id and all previous references are broken.

Using fs.stat I am fetching the inode value for each asset which appears to stay the same after a rename, when a file is deleted I store the asset data in a lookup table by inode and when a new file is added I check if the inode matches a recently deleted one and if it does I reuse the previous id and any additional previous data that might be useful to keep (such as the music speed conversion flag). With this change renaming an asset has a small flash of the asset being deleted before it reappears again a second a later with the scene collisions and actors positions unchanged.

As the inode values also get written to disk in the GBSProj file it's also now possible to rename an asset while GB Studio is closed and it will fix the links when you next load the project.

Maybe, I've only tested this on Mac so far. I'm interested in seeing if this works correctly on Windows. I believe Windows uses 64-bit inode values so I'm using the bigint arg when calling stat `await stat(filename, { bigint: true });` and storing the value as a string (since JSON doesn't support 64-bit values), without testing I can't be sure this is working. I'll do plenty before merging this.

Linux should be okay but also needs a lot of testing.

### DIFF
--- a/src/lib/helpers/clone.ts
+++ b/src/lib/helpers/clone.ts
@@ -1,0 +1,3 @@
+export const clone = <T>(input:T): T => {
+    return JSON.parse(JSON.stringify(input));
+}
--- a/src/lib/project/loadBackgroundData.js
+++ b/src/lib/project/loadBackgroundData.js
@@ -2,6 +2,7 @@ import glob from "glob";
 import { promisify } from "util";
 import uuid from "uuid/v4";
 import sizeOf from "image-size";
+import { stat } from "fs-extra";
 import parseAssetPath from "../helpers/path/parseAssetPath";
 
 const TILE_SIZE = 8;
@@ -13,6 +14,8 @@ const loadBackgroundData = projectRoot => async filename => {
   const { file, plugin } = parseAssetPath(filename, projectRoot, "backgrounds");
   try {
     const size = await sizeOfAsync(filename);
+    const fileStat = await stat(filename, { bigint: true });
+    const inode = fileStat.ino.toString();
     return {
       id: uuid(),
       plugin,
@@ -22,6 +25,7 @@ const loadBackgroundData = projectRoot => async filename => {
       imageWidth: size.width,
       imageHeight: size.height,
       filename: file,
+      inode,
       _v: Date.now()
     };
   } catch (e) {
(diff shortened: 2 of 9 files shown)


RichardULZ:
Awesome, should fix #387 too (which was not crashing in 2.0beta, but still had issues) 
Will have to try these out.
I don't think we've ever fully squashed all the asset reload bugs for everyone, just moved them around, as they are so hard to track down when they do happen (haven't seen any issues so far, but pretty sure there's at least one user on beta that has...) could be a further issue with dropbox or something.

chrismaltby:
@RichardULZ I've just done some testing on Windows and so far this seems to be working pretty well. The only issue I came across was Music files sometimes working and sometimes not, and I tracked it down to a one line config issue which was sometimes causing the add file event to fire before the delete file event because I wasn't polling the file for changes for long enough,.

I've just fixed this in the latest commit so now whenever a music file is changed it checks every 100 milliseconds to make sure no further changes happen until at least 5 seconds have passed using the awaitWriteFinish options in chokidar https://github.com/paulmillr/chokidar#performance I could probably reduce this time (it only waits one second for image assets) but I figured it doesn't need to be anywhere near as instant to update the music so it's probably better to err on the side of caution.

I need to test if this fixes #390 too. It's possible this will help there.

chrismaltby:
I've tested this on Windows, Mac and Linux now and it seems to be working well on all three!

Also from what I can tell it does seem to fix both #387 (undoing a deleted file restores the links in GB Studio) and #390 (So far been unable to get OpenMPT to break music links with this build)

PR #577, opened 2020-09-28
[UI] Add "tabs" type to event field

pau-tomas:
Updated Timer and Attach Input scripts to support it

Feature

The scripts inside `Attach Script to Button` and `Set Timer Script` behave in a different way than other Events (like all the conditional ones). They have different scope for labels and, in the case of the button script can persist across scenes. But there's no way to differentiate that fact.

Add a UI change that makes those scripts visually consistent (to some extent) with the scripts that behave like them (`On Update`, `On Init` and so) by adding a tab with a name on top of them.

No (although there's an ugly hack using `undefined` in the tabs condition to have the default script to show in all projects with no value for `__scriptTabs`)

* **Other information**:

There might be other ways to differentiate those scripts. And I'm not even sure there's a need for that, but wanted to test how it'll look.

Also, the tabs are a new type of field for Events that combined with `conditions` allows to organize the Event fields in horizontal blocks. Which means that if one day we want to go down the rabbit hole we could do this...


chrismaltby:
That's a good idea, I'll check this out later. Definitely like how this opens up the On Release / On Hold options too!

chrismaltby:
It was looking slightly wrong on the light theme

Screenshot 2020-09-30 at 20 54 21

but I just added a 1px left margin to the first and it's looking good now

Screenshot 2020-09-30 at 20 54 07

Thanks!

Issue #557, opened 2020-09-22
Vertical screenshake


coasterspaul:
**Is your feature request related to a problem? Please describe.**
Currently, `Camera: Shake` only does horizontal screenshake. This is generally fine, but there's a few cases (something heavy falling; in my specific game, shooting a powerful cannon upwards) where I think a vertical screenshake would look a little better. 

[Assuming, of course, it wouldn't be that much harder to add. I'm not familiar enough with the hardware, GBDK, or the GB Studio engine to know if there's some obscure limitation that would make vertical screenshake a pain to put in.]

**Describe the solution you'd like**
Renaming the current event to `Camera: Shake Horizontal` and adding a similar `Camera: Shake Vertical` event would probably the simplest way to go about this. Though you could stick to one event with an added option, to be fancy. I don't really care either way.

coasterspaul:
I should've checked to see if this had been asked for already again. Looks like #175 also mentions vertical screenshake (as well as asking for even more options.) Feel free to close this one and move discussion there if you want, but I do think this needed to be mentioned again.

Issue #175, opened 2019-05-16
[Feature] Camera Shake Vertical ++ Control


junkajii:
**Note: for support questions, please use one of the following channels**.

Reddit: [/r/gbstudio](https://www.reddit.com/r/gbstudio)
Discord: [Join Chat](https://discord.gg/bxerKnc)

This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [x] feature request

* **What is the current behavior?**
Camera Shake is horizontal and, we only have the time control.

* **What is the expected behavior?**
I hope to choose if the shake is vertical and, the offset and speed controls.

* **What is the motivation / use case for changing the behavior?**
I've looked at all Pokémon Generation I attacks (Red, Blue, Yellow), most attacks have horizontal shake, but have offset, velocity, and time differences, and in one attack the shake is vertical.

* **Please tell us about your environment:**

  - Version: 1.1.0
  - Platform: [Windows]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

00:31 until 00:45, you can see the shake variations.
https://youtu.be/4aBT7rjZoIE?t=31

10:00 here's the vertical shake.
https://youtu.be/4aBT7rjZoIE?t=600

Xeyler:
I'm working on a PR to add this feature now! 🎉

This involves some (kind of important?) design decisions. Specifically, what are the desired parameters for the screen shake script?

I'm thinking should_x_shake, should_y_shake, shake_length(seconds), and intensity.

Does this sound agreeable?

RichardULZ:
Not super trivial as we added a work around for incorrect tile loading during a screen shake, I believe this separated scroll x and screenshake x to have a third x variable in v2 beta?

Exposing more options, it may be ideal to have a "should shake" or just use "shake length" is not 0 to determine if we should be shaking the screen? Then further features like speed/intensity/axis won't have a performance impact unless the screen is shaking, apart from the extra variables that would be required in y axis.

Screen shake operations might currently be done with 16bit operations to the overall map position, it might be worth investigating if that could be done on the final 8 bit screen scroll instead, wrapping from 255 to 0 should look no different if we're within the loaded tiles.

Xeyler:
[...]

PR #581, opened 2020-10-01
Optional vertical shake

Xeyler:

Note: I was unable to find in-depth documentation alongside the code. I would be happy to update the relevant doc sources if someone could point me towards them!

This PR implements the "vertical screen shake" feature which has been requested several times. See: #175, #557 

Currently, GB Studio only supports shaking the camera horizontally.

This feature adds X and Y checkboxes to the camera shake command, allowing the user to shake the camera on the horizontal axis, vertical axis, or both axes.

The way I've introduced this change, the camera shake command behaves the same way by default until the user has changed the axes options.

* **Other information**:

Here's what a new camera shake with the horizontal and vertical axes enabled looks like! Note that my gif recorder slows down the motion.

xy-shake

Please let me know if any of the changes I've made are unsatisfactory. This is my first pull request on this repo, so I'm not as familiar with the codebase as I would be otherwise.


chrismaltby:
Great! Thanks I've had a quick skim through the code and it looks good so far. I'll do some testing soon and will let you know how I get on.

In terms of code documentation there's not really anything just yet, it's been on my todo list for a long time but to be honest I should probably just remove that line from the the PR template for now!

Xeyler:
Ah, I'm glad you like my PR! 😄🍰

I've just realized that with the changes I've made, entities don't shake on the vertical axis. I'm not sure how I missed this... And come to think of it, it's weird that my changes would introduce this behavior... I'll take another look at the engine when I catch a second.

If you find any other oversights or issues, please let me know! I'll be happy to rework my changes.

Xeyler:
Alright, with my latest commit, actors now shake on the vertical axis. And I've just checked: With my changes to the tests, `npm test` reports all are passing.

With that, I think this PR is ready to merge unless there are further comments about coding style, design decisions, or test coverage. 

Otherwise, thanks for having me! I've haven't used GB Studio much at all to be honest. In my projects, I've used straight assembly. Learning more about GBDK and GB Studio has been super cool... I'll have to experiment with them more! :heart: :video_game:

patrickmollohan:
Great work! No issues as far as I can tell from the projects I've tested this with.
One suggestion: I'd either make the X and Y checkboxes inline with each other (should just need to add `width: "50%",` to `"shouldShakeX"` and `"shouldShakeY"` fields in `src/lib/events/eventCameraShake.js` ), or perhaps a better approach, have a dropdown to select between "Horizontal", "Vertical", and "Diagonal". This would be more consistent with other areas of GB Studio, such as the dropdown in "Actor: Move To" and would eliminate the case where neither X or Y is checked (as I'm typing this, it occurs to me I haven't tested what that does).
Aside from that, it would be amazing to have this in GB Studio!

Xeyler:
The case where neither X or Y is checked is a strange one. Typically, it just won't shake at all (unless I'm wrong about my own code :sweat_drops:). 

However, if the project was started in a version of GB Studio which doesn't have the changes of my fork, then it's opened in my fork, the X and Y checkboxes won't be checked. But by default, only the horizontal axis will shake (to preserve the expected behavior).

I should spend some time to fix it so that when older projects are opened, the X checkbox is ticked and the Y checkbox isn't. And perhaps I should change it to be a dropdown so that the non-shake shake isn't possible?

patrickmollohan:
Just implemented the dropdown:
```
const l10n = require("../helpers/l10n").default;

const id = "EVENT_CAMERA_SHAKE";

const fields = [
  {
    key: "time",
    type: "number",
    label: l10n("FIELD_SECONDS"),
    width: "50%",
    min: 0,
    max: 10,
    step: 0.1,
    defaultValue: 0.5
  },
  {
    key: "shakeDirection",
    label: l10n("FIELD_DIRECTION"),
    width: "50%",
    type: "select",
    options: [
      ["horizontal", "↔ " + l10n("FIELD_HORIZONTAL")],
      ["vertical", "↕ " + l10n("FIELD_VERTICAL")],
      ["diagonal", "⤡ " + l10n("FIELD_DIAGONAL")]
    ],
    defaultValue: "horizontal"
  }
];

const compile = (input, helpers) => {
  const { cameraShake } = helpers;
  let seconds = typeof input.time === "number" ? input.time : 0.5;
  let shakeDirection = input.shakeDirection;
  let shouldShakeX = true;
  let shouldShakeY = false;
  switch (shakeDirection) {
  	case "horizontal":
  		shouldShakeX = true;
  		shouldShakeY = false;
  		break;
  	case "vertical":
  		shouldShakeX = false;
  		shouldShakeY = true;
  		break;
  	case "diagonal":
  		shouldShakeX = true;
  		shouldShakeY = true;
  		break;
  	default:
  		shouldShakeX = true;
  		shouldShakeY = false;
  }
  // Convert seconds into frames (60fps)
  while (seconds > 0) {
    const time = Math.min(seconds, 1);
    cameraShake(shouldShakeX, shouldShakeY, Math.ceil(60 * time));
    seconds -= time;
  }
};

module.exports = {
  id,
  fields,
  compile
};
```
From there, the localisations would need "FIELD_HORIZONTAL": "Horizontal" and "FIELD_VERTICAL": "Vertical" added in the respective languages.

chrismaltby:
Thanks @Xeyler this is working great! I've just made the changes @patrickmollohan suggested making the event use a dropdown, and I've added this to the engine changelog so if anyone had ejected their engine they can see which files have been updated.

PR #582, opened 2020-10-01
Add arm support

patrickmollohan:

Support the arm64 architecture.
Tested on a PinePhone running Mobian.

Now also supporting the armhf architecture.
Tested on a Raspberry Pi 3 B running "Raspberry Pi OS", formerly known as "Raspbian" but was rebranded to something less catchy because...reasons?
Note: the Pi 3 B does not have enough RAM to properly build GB Studio, so build it on a different system.

Cannot run GB Studio V2 Beta on arm64 devices. Or on the Pi.

Can run GB Studio V2 Beta on arm64 devices. And now can run on the Pi.

Not that I'm aware of.

* **Other information**:
Build by running `npm run make:linux-arm64`
9pkjdzde9eq51
And also `npm run make:linux-armhf`


blurymind:
one step closer to running it on android :)

chrismaltby:
Nice one, thanks @patrickmollohan! Do you think this build would allow it to run on a Raspberry Pi? I've not got a phone I could test this on but I might be able dig one of those out from somewhere :-)

patrickmollohan:
Thanks! I think it should run on the Raspberry Pi 3 and above, as long as the arm64 kernel is installed. When I initially tested on my Pi 3 a couple days ago, it said it couldn't run due to Raspbian running the armhf kernel. I plan to add support for that as well tonight with any luck and test on all the Pi devices I have here. I would imagine it should work on any arm64 device if it could run on the PinePhone.

patrickmollohan:
Added support for armhf. A few things to note:
electron-forge expects the --arch flag and the folder to be named armv7l as opposed to armhf, but the npm script is make:linux-armhf to be consistent with the Pi.
Also, although the tools themselves seem to run well on the Pi, I cannot build a game in GB Studio. I get the error:
`
Error: ENOENT, buildTools/linux-arm/tools_version not found in /usr/lib/gb-studio/resources/app.asar
`
I'm not sure why it is looking there for the tools_version file, when it should be looking in buildTools/linux-armv7l where it is. This isn't an issue with my arm64 version.

patrickmollohan:
Alright, fixed! Now building games on my Pi 3!
Changed after-copy.js to copy from buildTools/linux-arm instead of buildTools/linux-armv7l if the architecture is armv7l. I originally tried modifying src/lib/compiler/ensureBuildTools.js with a similar fix, but it would give me the path error when building the music, most likely due to src/lib/compiler/compileMusic.js. Instead of changing two files, I figured it would be easier to maintain if only one was changed, so I went for after-copy.js.

chrismaltby:
Thanks @patrickmollohan I've not had chance to try this out yet but I've just bought a new Raspberry PI 4 to test this out when I get some time :-)

patrickmollohan:
Awesome! I imagine it should be an even better experience on a Pi 4!

chrismaltby:
Hi @patrickmollohan Good news, I managed to get this to work on my Pi4!

I also had to compile from another Linux machine though apparently even 8GB isn't enough to build 😮 doesn't help I was also very low on disk space! I'm just thinking though since it seems we're quite close on getting GBDK 2020v4 support in (just one known bug in the compiler we're trying to get fixed now) it might be worth just holding off merging this until that its ready first and we can launch ARM support alongside 2020V4. It should just be a case of recompiling GBDK with the new version and the fixed SDCC then dropping it in over these folders

patrickmollohan:
Awesome!
That's really strange that 8GB isn't enough. Maybe it isn't a RAM issue? GB Studio builds fine on my laptop and PinePhone, both only have 4GB RAM. Strange.
I agree that would be wise to hold off merging until the new version of GBDK is implemented. Really great work, the both of you! I wish I could help more, but it's a touch above my head at this point. In any case, very excited to see that implemented!

patrickmollohan:
Sorry this took so long, @chrismaltby, had a few things come up. Both armhf and arm64 versions of GBDK-2020 have been updated.

JE55E6arlow:
By any chance is there a download link? I'd really love to see a .deb, because I have a Raspberry Pi 3 B+, and want GB Studio to work on it.

patrickmollohan:
@JE55E6arlow sorry for the delays; didn't see any notification for your comment. 
I built the .deb and .rpm files for armhf and arm64 against the latest version of GB Studio v2 and uploaded them [here](https://easyupload.io/m/otdc14); however, I haven't had the opportunity to test them. If there are any issues, let me know and I'll attempt to resolve them!

HuntingPhoenix:
Hi! Self admitted newbie here. Did support for ARM get added? I didn't see the files in buildTools in other branches so I wasn't sure (or beta builds for RPi).

patrickmollohan:
@HuntingPhoenix I closed the pull request due to inactivity. It's been ready for quite a while but was never merged. Should @chrismaltby be interested, I will update GBDK to the latest versions across all platforms, including ARM, and submit a new pull request, but for the time being, I don't see the need for it to hang out in limbo.

kedzie1138:
@patrickmollohan would love to try this out on my arm64 based Chromebook if you are able to share the .Deb again. :)  (The above link expired)

PR #583, opened 2020-10-01
Set smaller min width for main window

patrickmollohan:

Sets a smaller min width on the main window, allowing GB Studio to run maximised on devices with lower resolution screens (such as phones) or allow GB Studio to be tiled on medium resolution screens (such as low-end laptops).

Only medium resolution screens can run GB Studio maximised, and only high resolution screens can tile GB Studio.

Min width of the main window has been adjusted to compensate for lower resolutions, allowing more devices to run GB Studio maximised or tiled.

Not that I'm aware of.

* **Other information**:
Screenshot from 2020-10-01 10-33-35

### DIFF
--- a/src/components/library/Toolbar.css
+++ b/src/components/library/Toolbar.css
@@ -39,7 +39,6 @@
 }
 
 .Toolbar__FixedSpacer {
-  width: 20px;
   flex-shrink: 0;
 }
 
@@ -106,7 +105,6 @@
 
 .Toolbar__DropdownButton .Toolbar__DropdownButton__Triangle {
   width: 10px;
-  margin-left: 5px;
   flex-grow: 0;
 }
 
@@ -160,7 +158,7 @@
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-  padding: 0px 10px;
+  padding: 0px 5px;
 }
 
 .Toolbar__Search {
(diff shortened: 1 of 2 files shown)


chrismaltby:
Hi @patrickmollohan unfortunately this PR introduces some cropping of the UI on the Mac build as there's slightly less room because the window buttons are displayed inline in the toolbar

Screenshot 2020-10-01 at 21 20 05

See how the window title only displays a single character and the play button is mostly cropped on the right of the screen.

About the absolute smallest we can get without requiring some design changes is 720px wide.

Screenshot 2020-10-01 at 21 20 02

If you shrunk the search input and section button when the width reached a certain cutoff point you might just be able to fit to 600 though.

patrickmollohan:
O gotcha, whoops! I tested on Linux and Windows but my Hackintosh is down atm. I originally tried 720px for my PinePhone, but even though the screen's resolution is 720x1440, it would not resize to fullscreen. I tried 700px next, which worked on my PinePhone, but I realised it wouldn't do the tiling like shown on my laptop, which is why I dropped it to 600px. I'll try getting my Hackintosh back in operation and play around with shrinking the search input and section button like you said.

patrickmollohan:
I don't have my Hackintosh back up and running just yet, but I added the 80px padding temporarily in the css file so I can approximate what it looks like (removed before submitting the commit). Changed the 600px min width to 640px (half of the width of 1280px wide monitors). When the window is above 720px, everything is normal. Below that, a few things are shrunk: the selector, the search bar, and the fixed spacer to the left of the build button. The selector could be further reduced by eliminating the arrow and the search bar could be even smaller (cutting out the ellipses). Looks a lot more functional on my end, but will need to be tested on macOS to make sure the build button isn't off-screen still.

chrismaltby:
Thanks @patrickmollohan that's looking good now on my Mac :-)

Screenshot 2020-10-06 at 20 51 52

I'm starting to work on a UI refactor at the moment and I've already put in a test for a 600px wide toolbar in that version to make sure I don't cause a regression here when I eventually switch over https://ui.gbstudio.dev/?path=/story/components-toolbar--app-toolbar-small

PR #586, opened 2020-10-01
Add movement presets to OnUpdate if empty

chrismaltby:
Add feature to quickly choose movement script preset similar to 1.2.0 movement types

Since 2.0.0 the simple but not very flexible movement type dropdown has been replaced with the ability to create custom movement with the "OnUpdate" script. When migrating an old project the movement is converted for you but if you want to create a new actor with the old movement types you now need to understand how to script it manually or copy and paste and existing actor.

Adds a few preset buttons to generate quick movement scripts matching the previous movement types when editing an Actors "OnUpdate" script.

Screenshot 2020-10-01 at 20 55 51

These buttons only appear if the script is currently empty.

No

* **Other information**:

These two events probably only make sense on top down scenes, it might be good to have these buttons be context sensitive on the current scene type so that Platformer and Shoot Em' Up scenes contain different presets.

### DIFF
--- a/src/components/editors/ActorEditor.js
+++ b/src/components/editors/ActorEditor.js
@@ -9,7 +9,7 @@ import ScriptEditor from "../script/ScriptEditor";
 import DirectionPicker from "../forms/DirectionPicker";
 import { FormField, ToggleableFormField } from "../library/Forms";
 import castEventValue from "../../lib/helpers/castEventValue";
-import { DropdownButton } from "../library/Button";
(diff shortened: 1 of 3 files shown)


skRapKlan:
I think these events (random walk and random rotation) could be very useful in platformer and shooter levels too! I can think of a bunch of stuff they could be used for right when I read this..! Please implement them for all game modes. What if you want to have a spider crawling randomly on the wall? And maybe have a tab to select the directions you want the actor to move selecting random walk?

Issue #543, opened 2020-09-19
[Feature] Update gbdk-2020


patrickmollohan:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Updating gbdk-2020. There have been several pull requests that would seemingly speed things up, such as [faster sprite handling](https://github.com/untoxa/gbdk-2020/commit/fcd2ada57b076bb3ef8829cc492009289b6a66a5).

**Describe alternatives you've considered**
N/A

**Additional context**
I have compiled the latest gbdk-2020 for Ubuntu and get the following output when trying to build a game in GB Studio 2 Beta:
```
Preparing variables...
Preparing strings...
Preparing images...
Preparing ui...
Preparing sprites...
Preparing avatars...
Preparing music...
Preparing scenes...
Preparation complete
Compiling events...
Unlink _gbsbuild
Copy default engine
Looking for local engine in assets/engine
Local engine not found, using default engine
Generate header: data_ptrs.h
Generate data file: data_ptrs.c
Generate header: banks.h
Generate data file: bank_6.c
Generate data file: bank_7.c
Generate object file: bank_6.o
Generate object file: bank_7.o
music_track_101_ approx size in bytes: 6056
music_track_102_ approx size in bytes: 263
Approximate Music bank sizes: 1
Music bank for each track: 8, 8
data_ptrs.c rewritten with new song banks

lcc -Wa-l -Iinclude -c -o obj/main.o src/main.c
lcc -Wa-l -Iinclude -c -o obj/asm.macros.o src/core/asm.macros.s
lcc -Wa-l -Iinclude -c -o obj/Scroll_a.o src/core/Scroll_a.s
lcc -Wa-l -Iinclude -c -o obj/Camera_a.o src/core/Camera_a.s
lcc -Wa-l -Iinclude -c -o obj/gbt_player_bank1.o src/core/gbt_player_bank1.s
lcc -Wa-l -Iinclude -c -o obj/Actor_a.o src/core/Actor_a.s
lcc -Wa-l -Iinclude -c -o obj/gbt_player.o src/core/gbt_player.s
lcc -Wa-l -Iinclude -c -o obj/Math.o src/core/Math.c
lcc -Wa-l -Iinclude -c -o obj/Scroll.o src/core/Scroll.c
lcc -Wa-l -Iinclude -c -o obj/Sprite.o src/core/Sprite.c
[...]

PR #588, opened 2020-10-02
GBDK-2020 v4.0 support

untoxa:

Makes core compatible with the new GBDK-2020 v4.0 release

Multiple errors, described here: https://github.com/chrismaltby/gb-studio/issues/543

Nice and smooth compiling.

Changes to Actor_a.s are not fully compatible with previous version of GBDK-2020, however, if you pass definition of _shadow_OAM symbol into the linker, it should compile and run.

* **Other information**:

Two commits: changes to Actor_a.s (backward-compatible if _shadow_OAM is defined) and changes to Makefile (breaks compatibility, because of the new style of making GBC rom, random byte patching is not supported by makebin, you should use dedicated commandline parameters instead).


untoxa:
ps: GBDK-2020 v4.0 binaries for all platforms are not included into this pr. Should be taken here: https://github.com/Zal0/gbdk-2020/releases

chrismaltby:
Amazing!! Thanks for all the hard work @untoxa I'll test this out over the weekend. Very much looking forward to moving to the latest GBDK after all the good stuff you've been saying about it :-)

untoxa:
unfortunately seem to be not that smooth i hope: https://github.com/chrismaltby/gb-studio/issues/543 but looking forward for your help with debugging that. a bit strange, looks like some troubles at startup? at least it builds ok.

untoxa:
seem to work fine now

untoxa:
after SDCC #11925 workarounds are no longer needed, but upcoming GBDK-2020 now has ability to use "read of uninitialized WRAM" exceptions in BGB - that helps to find bad code, if you have those enabled in BGB settings.

chrismaltby:
Great, thanks for the update @untoxa it sounds like waiting for this next GBDK-2020 is the right thing to do now 👍 
I've got a branch ready to go with the LCD interrupt change you suggested on Discord too!

untoxa:
you should check carefully for 'inaccessable VRAM" issue, because if you write your LCD ISR in C it can return at unsafe LCD state and mess the timing of a function that is executed in the context of main() at that point. can you point me the commit?

RichardULZ:
bbbbbr approached me about the Wrote Twice at Addr error changing/going away with gbdk2020v4, and that they will try to make a tool that checks if data was overwritten, but can't detect if it's overflowed into an empty bank naturally. 

With our dynamic packing in 1.2.2 and 2.0, we usually check bank sizes are ok before they go to gbdk, but I have suspicion there's a few rare off by one cases where we don't catch it. Just got one about mod2gbt again, but that might be 1.2.1.
Including the discord transcript for consistency.
NotesOn Bank Overflow Changing

bbbbbr:
See this issue in GBDK for additional info about the multiple-write checking :
https://github.com/Zal0/gbdk-2020/issues/72

untoxa:
1. there had never been "overflowing into the next empty bank" checking, such checking is not needed.
2. automatic code arrangement mechanism can be added to gb studio. that is not very complex, the idea is described below:

for example, compile banks gbdk example. you will see this picture in the objects (bank_1.o): 
```
XL3
[cut]
S b_bank_1 Def000001
[cut]
A _CODE_1 size 1C flags 0 addr 0
S _bank_1 Def000000
[cut]
```
we can see here, that in the source file there is a __banked function bank_1() and the bank number of it is b_bank_1 that equals to 1. and that function uses 1C bytes of CODE_1 area.
so.
1. compile each script separately (better they be in one bank, all with `#pragma bank 255`)
2. load all objects into post-processing utility, make a dictionaty of used sizes in _CODE_255 area by each script
3. toss scripts by banks, try fit them best
4. modify object files: rename _CODE_255 into _CODE_N where N is the new bank number for the script and modify b_<SYMBOL_NAME> global symbol value to N (Def00000N).
5. link modified objects.

if banked functions are called through the trampoline, there will be no problems at all, and you will never get bank overflow for the code. you can also report "script too long" if it does not fit into 16K, but that is unlikely.

the same thing can be easily performed with data, but you should invent the way to pass bank number, maybe _bank_<SYMBOL>. the code in C will be:

SWITCH_ROM_MBC1(__bank_Tiles);
set_bkg_data(..., Tiles);

if you need arrays of pointers, then:

const farptr_t pointers[] = {{__bank_Tiles, Tiles}, ...};

untoxa:
to make array of __banked functions [`func1()`, `func2()`, `func3()`] you have to do this:
```
void _wrapper_func1() __nonbanked { func1(); }
void _wrapper_func2() __nonbanked { func2(); }
void _wrapper_func3() __nonbanked { func3(); }
const fn_t * funcs[] = {&_wrapper_func1, &_wrapper_func2, &_wrapper_func3};
```

untoxa:
that all together make full-automatic user-transparent bank management possible (and quite easy).

untoxa:
I updated the core, cleaned out most of PUSH_BANK/POP_BANK for CODE. BankManager for data is not necessary too, you can just switch banks when you need it. if it is necessary to save an restore bank context you can do somethins like this:
```
UINT8 _save = _current_bank;
SWITCH_ROM(bank_we_need);
<payload>
SWITCH_ROM(_save);
```

untoxa:
sorry. wrong button.

untoxa:
some comment about `void ScriptTimerUpdate_b()`
that function is called as banked from the core code from the `ScriptTimerUpdate()`, and as nonbanked, when gets called indirectly by pointer from script runner.
there is no problem, because banked and nonbanked functions are compatible WHEN THEY HAVE NO PARAMETERS. so we can define all functions in script_cmds[] array as banked, since not one has any parameters. so we can call them directly from the code, if there is such need without any wrapper (e.g. `ScriptTimerUpdate()`), and indirectly as long as we switch banks manually before call.
`ScriptTimerUpdate()` is not needed, i left it only because of the name without `_b` suffix.
i suggest to rename `ScriptTimerUpdate_b()` into `ScriptTimerUpdate()`

untoxa:
@chrismaltby i suggest to merge this pr, until it get too far from the v2beta branch, you may not merge makefile changes, and everything else will work with gbdk-2020 v3.2 if you pass `-Wl-g_shadow_OAM=0xC000` (define the _shadow_OAM in addition to .OAM with the same value)

untoxa:
checked, safe to merge.

RichardULZ:
Added changelogs and got it working with toxa's latest GBDK2020 v4.0.1 build for windows at least.
https://github.com/untoxa/gb-studio/pull/1

We don't have to use the Win7 Compat layer anymore! Performance on 4.0 was 6+ minutes with it on, but removing it has brought it down to 40 seconds, slightly faster than gbdk 3.1 with compat.

Not including gbdk in the pull just yet, as the changelog bloats out into the hundreds... But it is required for this pull to work!

Need help building gbdk2020 dev for mac @chrismaltby or @pautomas 
Toxa sugested "to compile (gbdk) from the develop branch.
and take sdcc from here: http://sdcc.sourceforge.net/snap.php
you may take #11952 - it is safe"

Here are the 2 builds from toxa, just place them in buildTools\win32-ia32\ ect, and update the tools version.

[gbdk-2020-4_0_1-linux.zip](https://github.com/chrismaltby/gb-studio/files/5540490/gbdk-2020-4_0_1-linux.zip)
[gbdk-2020_4_0_1-win.zip](https://github.com/chrismaltby/gb-studio/files/5540492/gbdk-2020_4_0_1-win.zip)

Just fyi, even without benchmarking I'm seeing something like 1.5 to 2x!!! performance increase on DMG!

untoxa:
gbdk-2020 v4.0.1 release: https://github.com/Zal0/gbdk-2020/releases/tag/4.0.1

chrismaltby:
Amazing! I've just dropped in the latest Mac binaries and so far this is working perfectly! Compile times are good and benchmarks looking great! Just give me a few days to test and review the engine changes but I think this is ready to go.

> Also required disabling the quick bank .o generation, as sdcc generates very different files, also changing a few compile flags.

I've got a branch somewhere where I got this working again with the latest GBDK, I can have a look at getting this working again as it was making a big difference on compile times for large projects

Thanks so much for everything you've done to put this together @untoxa and @RichardULZ (+ everyone involved in GBDK2020)

RichardULZ:
Toxa was considering whether we could split scripts into their own pushed arrays, and switch to compiling with farptrs, including the bank and address as symbols to be replaced by the compiler baked into scripts. 
That would be another final speed up for scripts.
On top of this, they noted the compiler could handle the symbols and bank packing, so if scripts were not passed in 1 chuck they would also get full debugger mapping.

But understandably, we're talking 1-0.5 seconds per data bank at the moment (large projects with 60 banks are back to 30 second compiles...) even when just compiling a single massive array, compared to the almost instant bank data .O fake conversion in JavaScript.

Less related to compile, now that functions can directly call banked functions, without push/pop, we could/should start directly referencing _b functions from core main, instead of a function to a function.

untoxa:
you can eject engine in gb studio 2, but you can not edit scripts or graphics or else data because it is a blob of bytes. if everything will be defined separately, sources will be readable! also scripts might be human-readable with the help of macros. compiling speed will be the same, if you afraid about that.

that also simplifies gb studio JS code. much easier to debug too.

it's not only about compiling speed.

emission of object files directly is bad.

chrismaltby:
Cool no problem I'll leave it as is.

Interested in this idea for human readable scripting. I think one of the things that stopped me going that route in the first place is working out how to automatically split the scripts between banks. Right now I treat sprite/tile/music/scripts as just data and can calculate exactly how much can fit in each bank and split them accordingly, I can see packing the data being a lot more complicated if it isn't just simple arrays. If we can make this readable though I'm all for it!

One for a different PR though.

I think there's lots of engine improvements this new GBDK should allow us to make which is very exciting (I just need to work on improving my C & ASM skills)

untoxa:
i suggest some macros similar to this: https://github.com/untoxa/rhythm/blob/master/src/song.c (that is a hUGETracker song to C export, you may patch songs in text!)

RichardULZ:
Hey this is a terrible idea I just thought about, multi core compiling. 
Whether we could allow 2 or more batches to process more C files, leaving the linker last.

The half second delay for converting simple bank files might be mostly command line overhead?

untoxa:
make --jobs 10

chrismaltby:
Can definitely try compiling in parallel once this is merged, either using make/msys2 or avoiding packaging any extra dependencies and just running the compiles in parallel from JS. To be honest though it's pretty fast since the Win7 Compat layer was removed.

So I was just about to merge this, I've got a branch with the binaries included in a1f478282dc94cbcf44930cdb21c816cadcb6acb but I noticed the binaries from the GBDK 2020 release page look like they are only 64-bit for Windows. It looks like the `lcc` binary is 32-bit but the SDCC ones aren't. I'm just going to set up a 32-bit Windows VM and build it from source, looks like the Mac build is using SDCC 11941 and Win64 was using 11952 so I'll probably use 11952

untoxa:
Chris, are you sure you need 32-bit windows build? You hardly find such a pc.

chrismaltby:
I wish :-) 10% of my user base is running Windows 32-bit! I don't think they'd be happy if I dropped support.

I've just found my 32-bit Windows disk image though so hoping it shouldn't be too bad to get it running

chrismaltby:
I ran out of time to get my Windows 32-bit dev environment set up for building last night but I've got it built now and confirmed working!

PR #592, opened 2020-10-05
Remove xcode tools dependency for Mac by using bash script rather than Makefile

chrismaltby:
Replaces the Makefile with a generated Bash script when building the app for Mac/Linux to remove dependencies needed to build a game.

On a Mac building without the full Xcode app or the Xcode Command Line Tools installed throws errors preventing a build from completing, there are also cases where some users have Xcode installed but their specific environment prevents it from working. There is currently a section in the documentation giving a fix https://www.gbstudio.dev/docs/build/#troubleshooting but many people do not find this and ideally GB Studio should not require any third party applications to be installed manually.

The only actual dependency on Xcode was that it installed GNU Make which is used on the Mac/Linux builds to run the build scripts. On Windows to prevent requiring a similar dependency we instead auto generate a make.bat file which was already handling things like the incremental builds by checking for the existence of compiled object files. This PR updates the script that was previously generating the make.bat file to also allow generating a make.sh Bash script which is now used in the Mac/Linux builds.

With this change a fresh Mac OS install can run GB Studio and build games without a single other application being installed.
 
No

* **Other information**:

I've kept the Makefile as it's still useful for anyone who exports their project source

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -598,6 +598,10 @@
   "GAMETYPE_PLATFORMER": "Platformer",
   "GAMETYPE_ADVENTURE": "Adventure",
   "GAMETYPE_SHMUP": "Shoot Em' Up",
-  "GAMETYPE_POINT_N_CLICK": "Point and Click"
+  "GAMETYPE_POINT_N_CLICK": "Point and Click",
 
+  "// 20": "Compiler -----------------------------------------------------",
+
+  "COMPILER_COMPILING": "Compiling",
+  "COMPILER_LINKING": "Linking"
 }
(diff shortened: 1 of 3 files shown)


RichardULZ:
On a related note, certain Windows users that are provided a computer for school have issues with building.
Some system administrators lock the use of command prompt, 
And other users get ` error: Spawn EPERM`  permissions error at the start while compiling, either from not having access to the temp folder or a strict antivirus. 

Providing the ability to change the build directory, say to a flash drive or documents folder, has a small chance of fixing the second issue.
I really enjoy the simple make.bat for windows, but i have seen zalo somehow use make commands with another utility for building.

Issue #502, opened 2020-08-29
Character moves in top down mode even with scripts attached


POWERUPT:
**Describe the bug**
I am porting a game I made in GB Studio 1 to GB Studio 2, and one scene uses a modified version of this [text input system](https://toxworks.itch.io/text-input-system-in-gb). It turns the player character into a cursor, and uses scripts to lock it to the positions of the background letters. When the A button is pressed, it enters a character based on the cursor's position. The issue is that this is broken in GB Studio 2 because, instead of the "Attach Script to Button" action replacing other inputs, it sends it, then the input is done, in this case the cursor moves by itself in addition to being moved a certain distance.

**To Reproduce**
1. Make a blank project in GB Studio 1
2. Attach scripts to each directional button that puts up a dialogue box that says what was pressed.
3. Test in v 1. Each button will bring up a dialogue box, but the arrow will stay still
4. Migrate to v 2 and test. The arrow will move and the dialogue box will appear.

**Expected behavior**
As previously stated, movement should be cancelled by the script.

**Screenshots**
Screenshots will not accurately portray the issue, but I will link a video if requested.

**Platform (please complete the following information):**
 - OS: Windows 10
 - App version: 2.0

pau-tomas:
Seems like a duplicate of #452

Issue #452, opened 2020-08-22
Attach button moves player


Veund:
attaching script to button doesn't disable player movement in 2.0 beta

i used the player as a cursor to make a menu incrementing a variable by one then setting players actor position when a dpad right is pressed, the player sets the position but then keeps moving from position until the dpad is lifted and pressed again 

this is in 2d top down mode  

win 10 64 bit 
2.0 beta

project attached

[Bakers Life.zip](https://github.com/chrismaltby/gb-studio/files/5112226/Bakers.Life.zip)

RichardULZ:
Note this is slightly related to #457 (joypad script fire on scene load button held) and #468 (joypad script fire after dialogue close)

eishiya:
@RichardULZ I'd be interested in a fourth option to address this:
4. Events to disable/re-enable the default functionality of individual buttons, ideally implemented with a selector that lets you select all the affected buttons in at once.
In addition to solving this issue while also allowing inputs during scripts when desired, it would be useful for things like disallowing the player from moving in a particular direction (e.g. not letting the player move left during a chase sequence).

RichardULZ:
@eishiya gbs 1.2.1 had mode 3 all the time without a checkbox, If you wanted to disable the default action, you would just use an empty "attach script to joypad", and you couldn't use it until it was detached. 

So it might not be necessary to have another event, depends on how the fix is implemented.

I think the issue is from the change to run things synchronously,
The previous side effect of disabling input was caused by only being able to run 1 script at a time, and never allowing player movement during a script, along with previous joypad buffers which were changed (causing the other issue).

patrickmollohan:
I would personally love to see the checkbox of option 3 so that you can choose whether to disable the default behaviour of the buttons when attaching scripts. This would be especially handy for the "A" button, which is the "interact" button. Currently, in order to attach a script to the "A" button and also replicate something similar to interacting, you have to check if a player is in a specific position relative to the object you want to interact with, check if they are facing the direction of the object, then "attack" the object with a completely transparent sprite, triggering an event with the "On Hit" section of the object. That's kinda clunky, but it works. It would be much easier, though, to just have that checkbox that allows me to not disable the default behaviour of the "A" button but still be able to attach a script to it.

PR #593, opened 2020-10-05
Override default button input when attaching script

pau-tomas:

Bug fix / feature ?

In 1.2 using "Attach script to button" disabled the default behavior for the button. So, attaching a script to A will disable the interaction with actors.

In 2.0 when attaching, the engine will try to run the two scripts in the background which lead to some unexpected issues like the player moving even if the directional buttons had an script attached ( #452 , #502 ).

This patch restores the 1.2 behavior. Attached scripts override the default input behavior except when a dialogue/menu is shown.

Maybe. Some people might be using the new default behavior of 2.0 (see below for potential solution and issue).

* **Other information**:

It'd be simple to add a checkbox to the attach input script with the option to override or not the default behavior. This would work ok for the direction buttons, but not for `A` as without this patch the default behavior (ie: interacting with an actor) is only called if the button is pressed repeatedly. 

The implementation is a bit convoluted and repetitive and I'm not a big fan of exposing `ui_block` to `Input.h`, but I couldn't think of a more elegant way.

### DIFF
--- a/appData/src/gb/engine_version
+++ b/appData/src/gb/engine_version
@@ -1,2 +1,2 @@
 # Don't remove this file, it is used to determine if your ejected engine is out of date.
-2.0.0-e5
+2.0.0-e6
--- a/appData/src/gb/include/Input.h
+++ b/appData/src/gb/include/Input.h
@@ -4,36 +4,44 @@
 #include <gb/gb.h>
 
 #include "BankData.h"
+#include "UI.h"
+
+#define ALLOW_DEFAULT_INPUT_RIGHT (ui_block || !GET_BIT(input_override_default,0))
+#define ALLOW_DEFAULT_INPUT_LEFT (ui_block || !GET_BIT(input_override_default,1))
+#define ALLOW_DEFAULT_INPUT_UP (ui_block || !GET_BIT(input_override_default,2))
+#define ALLOW_DEFAULT_INPUT_DOWN (ui_block || !GET_BIT(input_override_default,3))
+#define ALLOW_DEFAULT_INPUT_A (ui_block || !GET_BIT(input_override_default,4))
+#define ALLOW_DEFAULT_INPUT_B (ui_block || !GET_BIT(input_override_default,5))
 
 /* TRUE if left is being held on dpad */
-#define INPUT_LEFT (joy & J_LEFT)
+#define INPUT_LEFT ALLOW_DEFAULT_INPUT_LEFT && (joy & J_LEFT)
 
 /* TRUE if right is being held on dpad */
-#define INPUT_RIGHT (joy & J_RIGHT)
+#define INPUT_RIGHT ALLOW_DEFAULT_INPUT_RIGHT && (joy & J_RIGHT)
 
 /* TRUE if up is being held on dpad */
-#define INPUT_UP (joy & J_UP)
+#define INPUT_UP ALLOW_DEFAULT_INPUT_UP && (joy & J_UP)
 
 /* TRUE if down is being held on dpad */
-#define INPUT_DOWN (joy & J_DOWN)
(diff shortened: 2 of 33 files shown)


patrickmollohan:
I've been looking forward to this! A great way to disable player movement, for instance, by attaching an empty script to the D-Pad buttons. Awesome! Seems to work on my projects quite well.
This is somewhat related: I made an adjustment to the Script_SetInputScript_b() with your modifications so the same script can be attached easily to multiple buttons.
```
void Script_SetInputScript_b() {
  UBYTE input, index;

  input = script_cmd_args[0];

  if (script_cmd_args[1]) {
    SET_BIT_MASK(input_script_persist, input);
  } else {
    UNSET_BIT_MASK(input_script_persist, input);
  }

  SET_BIT_MASK(input_override_default, input);

  index = 0;
  for (index = 0; index != 8; ++index) {
    if (input & 1) {
      input_script_ptrs[index].bank = script_cmd_args[2];
      input_script_ptrs[index].offset = (script_cmd_args[3] * 256) + script_cmd_args[4];
    }
    input = input >> 1;
  }
}
```
eventInputScriptSet.js would need its `defaultValue` property changed from `"b"` to `["b"]` in the `input` field. I don't know if this would break anything, but with simple tests such as an empty script or shaking the camera when one of several buttons is pressed seems to work as expected. Also tested removing the script and that works when removing the script from all buttons or some of them. I don't know if this warrants a separate pull request or if you think it'd be worth implementing, but I was hoping to get your input (pun semi-intended) since you probably know the input system better than I do.

pau-tomas:
Thanks for testing it @patrickmollohan !

I like the idea of being able to add one script to multiple inputs and I think it should be a separate PR. I'm not sure if it'll require a migration script to update all the existing single inputs (`"b"`) to arrays of single item (`["b"]`).

Bastani:
This would help a lot in my project, I've been using a Boolean variable to counteract the current behavior.

chrismaltby:
Thanks @pau-tomas In testing there's one little bug I found is if you set a Text / Menu event to a direction button the ALLOW_DEFAULT macros return true for 1 frame, because the input script sets ui_block to true which allows the normal movement in the scene's update script just for that frame.

Screenshot 2020-10-06 at 20 46 06

If you update Core_Main.c to move `HandleInputScripts()` to be after the scene update it fixes it:

```
    if (!script_ctxs[0].script_ptr_bank && !ui_block) {      
      PUSH_BANK(stateBanks[scene_type]);
      updateFuncs[scene_type]();
      POP_BANK;
      HandleInputScripts();
    }
```

Do you think this will cause any issues for anyone though? I can't think of any cases where the order of these should introduce any new bugs but it's not impossible.

I guess it could become this instead but feels pretty ugly to do an extra check there if the other way works.

```
    if (!script_ctxs[0].script_ptr_bank && !ui_block) {   
      HandleInputScripts();   
      if (!ui_block) {
          PUSH_BANK(stateBanks[scene_type]);
          updateFuncs[scene_type]();
          POP_BANK;
      }
    }
```

Once you make that change though these seems to work great!

Are you also okay to update `src/lib/project/ejectEngineChangelog.ts` to add a `2.0.0-e6` to that list with the list of files that's changed so people know what to update in their ejected engines?

Thanks!

pau-tomas:
Thanks for checking Chris! 

I can't think of a case where changing the order of the calls might break things either. If anything it seem to make things more consistent in case, in the future, we introduce the option to not override the default behavior, as instead of being called if the button is pressed repeatedly it seems to do it always.

chrismaltby:
I'm all for automating that process if you've got an idea how 😁 I've never used Github actions before either so very interested to see if this would be possible!

pau-tomas:
Merged #595 (thanks @patrickmollohan!) and updated engine version and changelog (manually, for now :D )

chrismaltby:
Awesome thanks @pau-tomas and @patrickmollohan all works great!

Issue #591, opened 2020-10-05
Make "Set Background Palette" options clearer


coasterspaul:
**Is your feature request related to a problem? Please describe.**
In the `Palette: Set Background Palette` event, leaving a slot as "None":
![image](https://user-images.githubusercontent.com/11203426/95098555-720a5a00-06f4-11eb-9ebb-26417b40bee6.png)
actually sets it to the scene default. [Thanks to um3k for the clarification.] This is pretty useful for stuff like palette cycling - you don't have to set every single palette in every event - but a lot of people don't seem to know it's there.
(I think some people assumed "None" just meant the DMG palette, which isn't helped by the color preview.)

**Describe the solution you'd like**
Change the text for "None" to "Scene default" or similar. And change the color preview - either:
- hide it entirely
- set it to all white, which would look the same as getting rid of it entirely in the light theme, and it'd probably still make sense in other themes
- somehow actually get it to show the scene default for that palette slot, if that's even possible

**Describe alternatives you've considered**
Also, a "Keep As-Is" option, leaving a palette set the same instead of resetting to the default, would be useful for even wilder color-changing, assuming that's possible with how the event's set up. Since there'd be no way to tell what colors that would be ahead of time, just either:
- hide the preview entirely
- set it to all white, which would look the same as getting rid of it entirely in the light theme, and it'd probably still make sense in other themes

chrismaltby:
Hi @coasterspaul I've just made an update to improve this in #596 which should hopefully work better now. I just need to do some testing before I merge that but from a quick try it seems to be working much better.

A keep-as-is feature would be quite nice, I think it might be a bit tricky with they way palettes are handled though.

coasterspaul:
And yet you managed to add the keep-as-is feature to beta 4 anyway? Thanks so much!

chrismaltby:
Hah, yep I figured out a way to do it without changing too much in the end :-D

PR #596, opened 2020-10-07
Fix "none" value for palette select inputs in palette events

chrismaltby:
Implements a request from #591 making the palette select inputs on palette events clearer

As seen in #591 palette events have a default label of "None" with the swatch preview showing the default DMG colors when actually choosing this value would use the scene default colors defined on the Settings page.

The palette update events have been updated to closer match how they work in the Scene editor sidebar. The default label now says "Global Default", the palette number is prefixed before each label, and the swatch preview shows the global palette that would be used.

Screenshot 2020-10-07 at 21 33 52

Screenshot 2020-10-07 at 21 34 08

No

### DIFF
--- a/appData/src/gb/engine_version
+++ b/appData/src/gb/engine_version
@@ -1,2 +1,2 @@
 # Don't remove this file, it is used to determine if your ejected engine is out of date.
-2.0.0-e7
+2.0.0-e8
--- a/appData/src/gb/include/Palette.h
+++ b/appData/src/gb/include/Palette.h
@@ -30,5 +30,6 @@ extern UWORD SprPaletteBuffer[32];
 extern UWORD BkgPaletteBuffer[32];
 
 extern UBYTE palette_dirty;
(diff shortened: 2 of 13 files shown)


Bastani:
Is it possible to keep the none functionality so that it only changes the palettes that were set to?

I say this because I have one state in a game where there are 200 different palette combinations which would come down to 16 if we didn't have to specify each palette.

chrismaltby:
Hi @Bastani I initially thought that would be quite complicated with the current way palettes are handled but I think I've figured out a way this could work.

I've just updated this PR to now have an extra option "Don't Modify" which can keep the existing value for a palette allowing you to only switch a few palettes at once.

Screenshot 2020-10-10 at 15 51 37

RichardULZ:
That looks like a pretty smart approach, allows any number of palettes to be updated or not updated, i would have thought about a single palette change event but this is about as efficient, and doesn't use another event.

PR #619, opened 2020-10-15
Update Timer script even if On Update scripts are running

pau-tomas:

Bug fix

Timer scripts aren't executed if there's On Update scripts (that don't contain a Wait event) in the scene.

Timer scripts run even when On Update scripts are present.

🤷 I'm not sure of the side effects of the change. But the assumption is that `script_ctxs[0]` is used only for non-background scripts so `On Update` ones are never assigned there.

* **Other information**:

N/A


chrismaltby:
Yeah looks like that was a mistake from when I migrated the timer code over to the new engine. You're right that script_ctxs[0] is the only script we should be checking for there, good spot!

PR #620, opened 2020-10-15
Bugfixes for Attach Script to Button 

pau-tomas:

Two small fixes for regressions introduced by the recent changes on Attach Script to Button

1) When `Attach Script to Button` doesn't persist across scenes, the override default behavior does. Making default behaviors unavailable in following scenes after attaching a script.
2) `Attach Script to Button` allows for no button to be selected which results on the script being attached to all buttons. Same happens with `Remove Attached Script`, but in that case the script is detached from all the buttons.

1) Override flag bit is unset when the script is remove on scene change.
2) When only one button is selected, clicking that button doesn't deselect it.

For 2 if somebody is using the behavior to attach/remove form all buttons, it'll still work. Until they select one button, then they won't be able to deselect it again and they'll have to select all the buttons in order for the event to behave as before.

* **Other information**:

N/A

### DIFF
--- a/appData/src/gb/engine_version
+++ b/appData/src/gb/engine_version
@@ -1,2 +1,2 @@
 # Don't remove this file, it is used to determine if your ejected engine is out of date.
-2.0.0-e9
+2.0.0-e10
--- a/appData/src/gb/src/core/Input.c
+++ b/appData/src/gb/src/core/Input.c
@@ -41,6 +41,7 @@ void RemoveInputScripts() {
   for (i = 0; i != 8; ++i) {
     if (!GET_BIT(input_script_persist, i)) {
       input_script_ptrs[i].bank = 0;
+      UNSET_BIT(input_override_default, i);
     }
   }
 }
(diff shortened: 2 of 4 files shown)


chrismaltby:
Thanks @pau-tomas the persist across screens bug definitely fixed now. You were so close with the other one but not quite 😁 If you kept clicking the last button it would toggle off sometimes still.

You had
```js
if (value.indexOf(input.key) > -1 && value.length > 1) {
  onChange(value.filter(i => i !== input.key));
} else {
  onChange([].concat(value, input.key));
}
```
but you just wanted

```js
if (value.indexOf(input.key) > -1) {
  if (value.length > 1) {
    onChange(value.filter((i) => i !== input.key));
  }
} else {
  onChange([].concat(value, input.key));
}
```
when you clicked the last button you were actually causing the else path so the array became `["down", "down"]` and then if you clicked a second time it would see that the button is in the array, and now the array has length of 2 so it just filters them both out and you have an empty input array again, oops :-)

pau-tomas:
Ah! That was my first version and then I decided to "optimise it" clearly not paying much attention. Good catch!

PR #625, opened 2020-10-18
Configurable Engine Fields

chrismaltby:
Borrowing from an idea that @pau-tomas worked on pre beta1 this is a feature to allow customising values from the game engine (such as platformer velocity/acceleration/gravity values) without needing to eject the engine.

Changing any of the hard coded values in the platformer scenes requires ejecting your game engine and editing the values in `Platform.c`, which also requires keeping the engine up to date when GB Studio updates modify the engine files.

This update adds an `engine.json` file to the game engine with a fields array that works very similarly to the event fields array when defining scripting events allowing you to define a number of customisable values that will appear on the Settings section. The default game engine includes exposed values for the Platform scene allowing you to edit this values without ejecting the game engine.

Screenshot 2020-10-18 at 17 38 24

In @pau-tomas's original version these fields were sliders, and before merging these will definitely be something similar, the `engine.json` file will allow you to specify a UI element to use (slider/number field/dropdown/checkbox etc)

When you eject your engine the `engine.json` file will be ejected also allowing you add your own exposed values for your version of the game engine.

This PR is still work in progress and accepting feedback if people have suggestions. The plan is for it to include the following features:

- [x] Define an `engine.json` schema file for listing engine fields **DONE**
- [x] Add Engine Props editor to Settings section generated using `engine.json` data **DONE**
- [x] Store engine overridden values in `.gbsproj` file **DONE**
- [x] Compile engine props into `data_ptrs.h` and set the values in `data_ptrs.c` **DONE**
- [x] Update `Platform.c` to use compiled values  **DONE**
- [x] Allow specifying UI elements to display on Settings page with limits such as min/max values **DONE**
- [x] Reload fields from ejected `engine.json` when file changes **DONE**
- [x] Add new event for modifying an engine prop dynamically (will be a little tough as the props support a mix of 8-bit and 16-bit values **DONE**
- [x] Allow modifying an engine prop value using variables (select two variables for hi/lo when updating 16-bit prop?) **DONE**
- [x] Remove `engine_version` file and replace with with version field in `engine.json` **DONE**
- [x] Expose an 8px/16px grid option for TopDown **DONE**
- [x] Change "Default Fade Style" to be implemented as an engine prop **DONE**
 
If you've ejected your game engine, you'll need to re-eject and merge your changes

* **Other information**:

I think it's outside of the scope of this PR but the end goal here is to extend this into plugins so that it would be possible to create a plugin that defines a new scene type (as a C file) and a list of new engine props (in JSON) that you can drop into any project (and share with each other) to add custom game modes without needing to eject the game engine at all.

* **UPDATE 22/10/20** 

I've had a big rethink about how the settings page looks to make sure it will still be manageable as the number of fields here will grow.

Screenshot 2020-10-22 at 18 03 15

This a little inspired by the project settings section of Xcode and allows searching by property name.

@pau-tomas I also changed the Cart Type selector to just be a simple dropdown now. Can probably just add the Rumble On/Off as either a checkbox under this setting or an extra item on the dropdown, whichever you think makes the most sense.

Screenshot 2020-10-22 at 18 05 50


pau-tomas:
That new settings page looks super fancy! ❤️ 

Will take a look at what's best for the rumble options as not all cartridge types support it. Thanks!

pau-tomas:
I've been playing with the build and so far it works great!

`variablePair` is going to be a fun one to add to Custom Events. If you want to leave it as "not implemented" for now I can take a look once it's merged.

I know it's still WIP and it's a tricky one. But I wonder if the concept of `hi`/`lo` and having to select two variables might confuse people? Maybe if a WORD is needed there could be a single selector for 1 variable and a read-only field with the variable after the selected one. Then a note in the event saying something like "Since we need to store a number bigger than 255 so we'll be using two variables".

src/components/script/ScriptEventForm.js
```diff
@@ -53,6 +56,71 @@ class ScriptEventForm extends Component {
...
+              variable: "0",
+            },
+          });
+        }
+        if (command === EVENT_ENGINE_FIELD_STORE) {
+          if (is16BitCType(engineField.cType)) {
+            return [].concat(
+              eventCommands,
+              {
+                key: "value",
```
pau-tomas, line 103: Loading those dynamically will make more difficult (maybe even not possible) to identify the fields as variables inside a Custom Event. Because the `type` field won't be available neither in the event instance nor in the event declaration :/

chrismaltby:
Thanks for reviewing this @pau-tomas got to admit I haven't got around to testing this from within a custom event yet so that's a good point on ScriptEventForm problem. I'll have a think of some ways I can make this work!

On the subject of variable pairs I wasn't 100% sold on the current implementation. I think you're probably right, having a single dropdown that just picks two consecutive variables when you need a 16-bit value is a much nicer option, I'll give this a go, should make it easier for custom events too 👍 There will be some weird edge cases though where if you selected Local5 or Temp1 there will be no consecutive variable to use for the low value, maybe I just need to filter them out from the dropdown.

If I'd have designed it now maybe all variables would have been 16-bit from the beginning 😁  Once we have this ability it could open the door for a 16-bit maths event too which could be useful.

chrismaltby:
Hi @pau-tomas I've made a mock up of how we could replace the `<VariableSelect />` component with a single new component that can handle both 8-bit and 16-bit variables. The tooltip would appear if you hovered, or clicked on the left of the input and would allow for some text to explain the difference between 8-bit and 16-bit values and how they're being used.

Screenshot 2020-10-26 at 18 35 30

pau-tomas:
I love that one! It gives away some of the technicalities (8-bit/16-bit) but makes it simple and with a clear explanation. I prefer the ":" option as it can't be confused with an addition.

RichardULZ:
Wow, never imagined you attempting 16bit variables in 2.0!

chrismaltby:
Well this PR ended up being a lot bigger than I anticipated 😅 

@pau-tomas from what I can tell custom events should be working with the engine field events now.

I removed the concept of variablePair and just added an optional parameter to the existing variable input to specify that you want to use 16-bits, only the hi variable is stored and the compile function grabs the next variable to use as the lo value.

I've also added a new field to custom event variables called `type` which is used to keep track of which variables have been used in a 16-bit calculation in the custom event so that when you use the event it gives you a 16-bit dropdown instead.

I've just got a few more tests to write, I need to do the engine changelog, and then I still need to replace the engine_version file checks with the new json format but nearly finished now if you want to give it a try again.

pau-tomas:
Thanks for working on that Chris! 

I won't be able to test the latest changes until the end of the week (I'm moving to Spain for a couple of months... if lockdown allows me) but I have been working a local build of the Custom Events 2 branch based on this branch from a couple of days ago and things seemed to work well over there.

chrismaltby:
No worries @pau-tomas It's giving me some time to try writing tests for the React components (which I've never actually done before!) From what I can tell it's all working but I'll hold off on merging for now as I think you have a much better understanding of the edge cases to watch out for with custom events 😄 

I hope everything goes well with the move, and these new lockdown rules I hear they're about to bring in don't cause you too much trouble.

Very much looking forward to seeing what's new in `Custom Events 2` by the way!

RichardULZ:
Hey @chrismaltby I just pushed a fix to this branch, not sure if I should keep making fixes for here, or the current v2beta if we need to put engine fields on hold for a little bit.

It's getting a bit unwieldly to track all these bugs, wondering if you could open the projects page to collaborators so we can jot down todo lists?

chrismaltby:
Hi @RichardULZ thanks for the PR, I've just merged it here for now hopefully won't be much longer before this in the V2 branch though. If a feature isn't affected by this PR, it could probably just put it straight into v2beta but Fade was covered here so I think merging to this branch made sense.

I was just playing around with permissions on the repo (you might have got an email that's probably now revoked) and it looked like GitHub wouldn't let me add users to be able to edit the project pages without also giving full unrestricted access to the entire repo :-/ I might need to move the repo to be owned by an organisation to get more control and possibly need to start paying GitHub, I'm not sure yet.

Rather than mess around with that I thought we could set up a team on Trello, I think we can have unlimited users but just a 10 board limit which I think will be fine. I've created a team and some initial boards and a public roadmap (nothing in it yet) at https://trello.com/b/lJgVq4Dl/gb-studio-public-roadmap

I think I've got both of your email address @pau-tomas and @RichardULZ so I can send you both an invite. Let me know if there's anyone else you'd like added (or I think we can just generate an invite link to share with the moderators on Discord)

pau-tomas:
Finally had some time to play with the build and Custom Events 1 and things seem to be working fine so far.

I found a small issue not related to Custom Events: Updating "Grid Size" via Event doesn't seem to work when selecting `16` (either using dropdown or variable).

chrismaltby:
Thanks for looking into this @pau-tomas I hope the move to Spain went well! One of the Senior Devs at work is trying to get back to Spain also this weekend!

Any chance you can upload a project where the grid size update doesn't seem to be working? In my testing I hadn't seen that and I've tried but not succeeded at reproducing the issue. I've merged but if there's something that needs fixing around the grid update event I can fix that separately.

pau-tomas:
The move went well, thanks! We're already settled :) Hopefully they'll be able to arrive without problems too!

Here's an example project, with two scenes one setting the prop value with a variable and another from the dropdown: [Grid16Prop.zip](https://github.com/chrismaltby/gb-studio/files/5505465/Grid16Prop.zip)

chrismaltby:
That's good to hear :-)

I've checked out your sample project, the issue is a little awkward but basically with the starting position you were using the player is actually off the 16px grid when the event fires which causes the player to get stuck. 

Because of the way the GB Studio coordinate system works the X coordinate needs to be an even number and the Y coordinate needs to be an odd number when switching to the 16px grid dynamically mid scene (because the top/left coordinate for a 16px grid would be [x=0, y=1]. If the grid is already set as 16px when loading the scene the `Start_TopDown()` includes some code to snap to the 16px grid but not sure a good way to do this automatically if you change mid scene (maybe could keep the previous frame's value and compare and snap if changed to 16px since the previous frame...)

The other option is maybe the function `Start_TopDown()` could fire after the scene start event is finished, but I'm pretty sure that will bring in new issues. Or there could be `PostStartScript_TopDown()` and now it's getting messy... Hmm not too sure what to do about this yet.

Issue #573, opened 2020-09-27
Projects dissappear when shooting sideways


DuendeInexistente:
**Describe the bug**
Shooting up and down works fine, but to the sides most of them disappear in one or two frames of appearing. 
[This is the game's folder in question, including a rom build.](https://github.com/chrismaltby/gb-studio/files/5289007/Untitled4.zip)

**To Reproduce**
Shoot sideways.

**Screenshots**
[Here's](https://youtu.be/R3WSuKO9TZU) a recording of the error.
**Platform (please complete the following information):**
 - OS: Ubuntu 20
 - App version Latest v2 beta build as of a few hours ago.

BTTRG:
The reason the bullets are disappearing is because they are set to collide with the player. Deselect this and it will work.

PR #628, opened 2020-10-20
Reduce projectile collision width to fix #506 & #573

### LINKED ISSUE #506: Projectiles hit the player one tile (16x16) away from the sides
**Describe the bug**
When a projectile is launched towards the player, it will collide even when it is just grazing the player from one 16x16 tile away.

**To Reproduce**
1. Launch a projectile towards the player while being shifted by one 16x16 tile.
2. See error

**Expected behavior**
The projectile should continue without registering a collision with the player.

**Screenshots**
Projectile2

**Platform (please complete the following information):**
 - OS: Windows 10
 - Beta 2.0

**Additional context**
Project file: [ProjectileTest.zip](https://github.com/chrismaltby/gb-studio/files/5146977/ProjectileTest.zip)

RichardULZ:

Tiny Bug fix

#506 actors moving up or down have a width of 5 tiles if on grid, instead of 3, due to a value of +-16 each side including equal to. 
#573 Is user error where a player fired projectile can collide with the player.

Reducing to 15 will fix this on grid to 3 tiles horizontally, but is extremely tight in adventure mode.
Reduced a further 3 px each side to 12, assuming projectile with a 2px border each side & Actor having 1px leeway each side, this feels pretty good.

Attacks with a sword will not do crazy sweeping damage horizontally, some testing may be required if this needs to be pushed to 13+- to better attack enemies at the tip of a sword.

* **Other information**:
Should probably increase vertical too, as 8+- is barely 3 tiles? Some note of missed shots in shootemup.
@pautomas  asked if this could ever be expanded/performant enough to take bounding per actor, however this would need both actor/projectile bounding to be correct...
In this context, exposing projectile bounding may be wise where a sword/arrow/bullet could be slimmer in certain directions, than a ball or swipe. Changes to that only affects per pixel modes like adventure or platformer.

### DIFF
--- a/appData/src/gb/engine_version
+++ b/appData/src/gb/engine_version
@@ -1,2 +1,2 @@
 # Don't remove this file, it is used to determine if your ejected engine is out of date.
-2.0.0-e10
+2.0.0-e11
--- a/appData/src/gb/src/core/Projectiles_b.c
+++ b/appData/src/gb/src/core/Projectiles_b.c
@@ -118,8 +118,8 @@ void UpdateProjectiles_b() {
           continue;
         }
 
-        if ((projectiles[i].pos.x + 16 >= actors[a].pos.x) &&
-            (projectiles[i].pos.x <= actors[a].pos.x + 16) &&
+        if ((projectiles[i].pos.x + 12 >= actors[a].pos.x) &&
+            (projectiles[i].pos.x <= actors[a].pos.x + 12) &&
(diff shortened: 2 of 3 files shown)


chrismaltby:
Great, thanks @RichardULZ! Definitely think there's a lot of room to improve collision detection still. I'm hoping once we get on to the new GBDK there will be a bit more cpu to play with to get this right.

PR #633, opened 2020-10-23
Don't add sprites to the sprite scene list if not used

pau-tomas:

Bug fix

Reported [on Discord](https://discord.com/channels/554713715442712616/585023243732123676/769298433202061364) a project was failing on compile with the following error:

image

After a bit of debugging, the error was due to a spritesheet field being set to `None` in a commented event in a completely different script from the same scene.

Even if the spritesheet didn't exist, it was added to the scene sprite list with index `-1`. Then, when compiling a call to `Actor: Set Sprite Sheet`, the call to `getSpriteOffset` looks in the global sprite list for a sprite with index `-1`, which is `undefined` and then tries to access `size` on that sprite, causing a crash.

If a sprite found in a scene script can't be found in the `usedSprites` list simply skip it and don't add it to the scene sprite list.

No

* **Other information**:

There's should probably be some extra guard in `getSpriteOffset` to make sure the  case for `undefined` sprites is covered there too. However I'm not sure we can provide a good error message that points the user to how to fix the issue as we don't have enough information about why the sprite might be `undefined`.

### DIFF
--- a/src/lib/compiler/compileData.js
+++ b/src/lib/compiler/compileData.js
@@ -1156,7 +1156,7 @@ export const precompileScenes = (
         const spriteIndex = usedSprites.findIndex(
           (sprite) => sprite.id === spriteId
         );
-        if (memo.indexOf(spriteIndex) === -1) {
+        if (spriteIndex !== -1 && memo.indexOf(spriteIndex) === -1) {
           memo.push(spriteIndex);
         }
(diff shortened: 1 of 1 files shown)


chrismaltby:
Thanks @pau-tomas definitely want to work on improving some of those error messages at some point!

Issue #568, opened 2020-09-26
Enabling Color Mode causes graphical glitches


terminaltux:
**Describe the bug**
When I create a large background (like 320x288px) and move the camera instantly, graphical glitches appear on the screen during the transition if Color Mode is enabled.

If Color Mode **is not enabled**, it works fine:
![DMG_No-Issues](https://user-images.githubusercontent.com/50606357/94347753-15bc8180-fff4-11ea-9f2f-81450402476b.gif)

If I Enable Color Mode, the graphical glitches appear:
![GBC-glitch-01a](https://user-images.githubusercontent.com/50606357/94347764-2cfb6f00-fff4-11ea-9530-55e001db5452.png)
![GBC-glitch-02a](https://user-images.githubusercontent.com/50606357/94347766-2f5dc900-fff4-11ea-9337-1aef9949d4ab.png)
![GBC-glitch-03a](https://user-images.githubusercontent.com/50606357/94347767-31c02300-fff4-11ea-8a81-4c02b4267a79.png)

**To Reproduce**

**Easy Method:**

1. Download my sample project [here](https://mega.nz/file/YbhShIQT#nlIEpvrxPLbYpdkz-6IF3531Yr_jz1eVxB9z2QiyShA).
2. Run the game to see that no graphic glitches appear
3. Enable Color Mode and run again.
4. They are quick flashes, but you should be able to see the garbled graphic during the camera move transitions.

**Create Your Own Method:**

1. Create a background image that is larger than 256 x 256px. Mine is 320 x 288px.
2. Add an animated actor or two.
3. Add a Camera: Move To event that goes from one 160x144px section, to another.
4. Set the speed to Instant.
5. In the Settings, check Enable Color Mode.

**Expected behavior**
The camera will move to the specified X,Y coordinate without flashing garbled graphics/tiles while doing so.

**Platform (please complete the following information):**
-OS: Fedora 32, 64-bit
-App version: 2.0 beta2 (itch.io build)

RichardULZ:
Note you forgot to mention the use of fade to black which is key to this issue.
Related to a recent change i didn't thoroughly check, the fade is only applied on next vblank to avoid tearing on gbc, however the screen rerender is happening before that activates. 
Will need to test an extra wait vblank in scroll c, but that may not cut it, alternative force palette refresh as part of update, possibly optimise screen instant black to fill registers with 0 instead of calculating black.

terminaltux:
[...]

Issue #638, opened 2020-11-01
Loading more actors on screen causing a crash, even if they are hidden.


RichardULZ:
**Describe the bug**
https://discord.com/channels/554713715442712616/570925559346102273/772217829814304828
User on discord has a platformer scene with 21 actors, the arrangement makes an easy to reproduce crash where shooting enemies and collecting coins does not reliably remove them from the actor pool, and when reaching the center ladder, the new actors on the right load in causing a crash, despite not being over 10 actors. 

**To Reproduce**
1. Download the project from discord,
2. Start the game and go in the trash can
3. shoot the enemy on the RIGHT, Then continue LEFT and down.
4. Once you get to the ladder in the center, it will likely crash.

Not shooting the first enemy on the right, and immediately moving left, may not crash, but still produce artifacts like stuck sprites (coin doesn't hide after exploding) or half sprites where they shouldn't be.

**Expected behavior**
Actor that is hidden + collision disabled + update script stopped is correctly removed from the actor pool. 

**Screenshots**
![image](https://user-images.githubusercontent.com/34431457/97793577-ad943780-1c52-11eb-8058-590cfdf143c6.png)
![Crash](https://user-images.githubusercontent.com/34431457/97793571-96554a00-1c52-11eb-8314-c1f56cfa6e0a.gif)

**Platform (please complete the following information):**
 - OS: Windows 10
 - App version 2.0 beta 5 and feature/engineprops branch

**Additional context**
Re rendering the screen with 2 instant camera moves also fixes the actor pool, but clearer/cleaner ways to remove or add actors to the active pool should be investigated.

RichardULZ:
This issue might be more confusing than I thought, The actor pooling system just activates or deactivates the actors id, from 1-30, so there shouldn't be any issues with overwriting an actor slot like i first thought. 
It's still a very repeatable bug at least, but nothing seems to help so far.

RichardULZ:
Ahh, i think it has something to do with how actor scripts are assigned a thread, disabling the enemies update scripts seems to fix the issue, (still some odd coins not disappearing glitches once in a while...)
Perhaps my original thought was close, just for the script pool instead...

PR #639, opened 2020-11-01
Fix instant camera move with Fade to black

RichardULZ:

bug fix on the Engine props branch (as this already modifies fade to black)

#568
A bug from old Fade to black vsync change, where instant camera moves could show 1 frame of garbage.

Force black palette for everything before the re render occurs in scroll.c, bypassing the palette switch.

Not likely.

* **Other information**:
#625 for engine props/fields pr

### DIFF
--- a/appData/src/gb/src/core/Scroll.c
+++ b/appData/src/gb/src/core/Scroll.c
@@ -6,6 +6,7 @@
 #include "DataManager.h"
 #include "GameTime.h"
 #include "FadeManager.h"
+#include "Palette.h"
 #include "data_ptrs.h"
 
 INT16 scroll_x = 0;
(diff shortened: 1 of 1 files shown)


chrismaltby:
Thanks @RichardULZ part of me wonders if this should be in a function in either Palette.c or FadeManager.c but since we only need it once probably okay to have it here for now and could refactor this later. I'm hoping with the new GBDK that we'll be able use `inline` functions and at that point could probably start refactoring some things in the engine, making more smaller functions without having performance issues. For now this works though!

From testing this though I saw a new issue, not sure if it's been reported already but the instant camera move was keeping the player visible after the camera move even if they should be offscreen.

ezgif-3-9aecd3c19b89

RichardULZ:
That's weird, I was suspecting re render would recalculate the actor pool, (which I think get's dirty after normal use, issue #638 ) But perhaps the player isn't considered at all for that? The player being off screen is not a case you typically expect.

PR #641, opened 2020-11-01
Increase limit of textMenu when using Dialogue

Bastani:

This change to the textMenu allows up to 16 characters to be used when using the Dialogue menu style. It trims to 6 characters when using the default "menu" option or greater than 4 menu options.
test

Using the Dialogue menu style causes menu options to be trimmed regardless of the space available.

When using the Dialogue menu style, if the number of options is 4 or less, the space available is increased from 6 to 16.

Not that I'm aware of.

* **Other information**:
~~While this change does fix the text being shown, the eventMenu.js script will need to be changed to allow for actual use. I am using custom plugin scripts to counteract it.~~

### DIFF
--- a/src/lib/compiler/scriptBuilder.js
+++ b/src/lib/compiler/scriptBuilder.js
@@ -445,9 +445,17 @@ class ScriptBuilder {
   ) => {
     const output = this.output;
     const { strings, variables, event } = this.options;
-    const menuText = options
+    let menuText;
+    if(options.length > 4 || layout === "menu"){
+      menuText = options
       .map((option, index) => trimlines(option || "", 6, 1) || `Item ${index + 1}`)
       .join("\n");
+    }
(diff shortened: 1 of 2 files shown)


pau-tomas:
While I like this idea, I think the point you make in `Other information` means that this change shouldn't be merged as is. Sorry.

One option to make `textMenu` more flexible for plugins would be to add an optional parameter (defaulting to `6`) to the function in `scriptBuilder.js` to modify the max character in the `trimlines` call.  That way a plugin event could be created that supports more characters but is limited to 4 items and "menu" style.

Bastani:
@pau-tomas So I just updated eventMenu to follow what is needed. It will only trim to 16 characters when it is Dialogue and there are 4 or less options. Otherwise it will trim to 6 characters.

PR #646, opened 2020-11-06
Use A button for jump and B to accelerate in Platform mode.

gmarty:

This change change the platform mode commands to use button A to jump and button B to accelerate. This is in line with most commercial plaftormer on the platform and provides a better user experience.

Currently, jump is assigned to button B and accelerate to button A which is counter intuitive to seasoned gamers.

This change change the platform mode commands to use button A to jump and button B to accelerate.

Yes, this change is breaking the experience for users making games on the v2 branch.

* **Other information**:
This is my first patch, so it may be crude.
Also, I haven't discussed this with any of the maintainers and they may want this as a configurable setting, in which case we'll  need to discuss the requirements (e.g. setting at the game level or scene level...).


DeerTears:
I like that this is happening, what you've done should be the default, but we should probably add the option to keep the old button layout. There's the new editor properties feature #625 that might(?) be helpful to swap these buttons with a checkbox. Something like "Platformer: Jump with B" starting in the "false" position.

I think it would be unfair to change the GB Studio control scheme at the mercy of everyone who's used to it by now, without giving them the option to go back. Changing this would actually break some projects that rely on "B" as the "interact" button across scenes. The "Attach Script to Button" event with "persist between scenes" checked would cause any events (like an 'interaction' sound effect) to occur during a jump, if this change is merged. But this is a bigger problem that a new event type like "if scene type" or "set variable to scene type" would help fix, something for another issue.

gmarty:
I thought it would  be acceptable to do since the platformer mode is still in beta stage, but I understand it could mess with projects attaching a script to the button.
I'm happy to do the change you suggested. I have to wait for https://github.com/chrismaltby/gb-studio/pull/625 to land in the `v2beta` branch first. I had a look at the WIP and it looks pretty straightforward to add new fields. I can do something similar to what is done for `fade_style` and use a field of type `select`.

chrismaltby:
I just went through a pile of Game Boy games to check and the count was 7 games using A to jump and 1 game using B (Oddworld Adventures strangely enough), I'm inclined to agree, feels like it's almost a bug that GB Studio is using a flipped button layout.

I think this probably actually the right change to make and while it might require some updates in projects using the beta I think this is a change that should be made. While it would technically be possible for this to be configurable now as an engine field I think this should probably be kept using the method you used @gmarty as making it dynamic could introduce slow down in one of the scenes that is already quite CPU intensive. Like you say this is a beta and some breaking changes at this point can be expected, I just wish I'd spotted this sooner...

Looks like there's some merge conflicts now though since I pushed the engine fields update, are you okay to fix these?

gmarty:
Ugh, this rebase went horribly wrong, I forgot how to git. Can someone help here, otherwise I can just start a fresh branch and PR as the changes are simple.

chrismaltby:
Hah, oh dear... Wish I could help but I'm not sure what you've done there 😀 I think it might just be easier to branch off v2beta again and make a new PR!

There's a bit of a discussion that overlaps the idea in this PR by the way over at #648

gmarty:
Implemented in https://github.com/chrismaltby/gb-studio/pull/652.

PR #647, opened 2020-11-06
Keep the camera centered on the scene when shaking.

gmarty:

This is a minor polishing patch. It keeps the camera centered on the scene when shaking.

At this time, the shaking of the camera only reveals the right hand side or the bottom of the scene when it shakes (depending on horizontal, vertical or diagonal shaking).
In a game I'm working on, I wanted the shaking to reveal briefly a border on a static scene that would be otherwise invisible because it is located outside of the screen. The current shaking would show only the right and the bottom of that border which doesn't look good.

This patch keeps the camera centered on the screen when it shakes, thus revealing part of the scene located on top, bottom, right and left.
Before the screen offset when shaking would be from 0 to 5 pixels, now the range of offset is (-5, +5) pixels.

It's a purely cosmetic change and is hardly noticeable in most cases.

* **Other information**:


chrismaltby:
Thanks @gmarty this is working great!

PR #649, opened 2020-11-07
Feature/navigator

chrismaltby:
Feature, replaces the "Navigation" section on the right hand sidebar with a dedicated navigation section on a new left hand sidebar.

When selecting a scene the right hand sidebar contains a Navigation section listing all actors and triggers in the scene allowing you to quickly select any entities within the scene. This worked okay in GB Studio v1 where the maximum number of actors and triggers per scene was quite low but as this had grown the list is now quite difficult to use and can take up a large amount of space pushing the scripting section (which is more often the section that's wanted) quite far down. It is also difficult to navigate between actors within a scene as after clicking on any the full sidebar switches to the editor of the selected entity.

A new optional left hand sidebar has been added dedicated to navigating your project.

Screenshot 2020-11-07 at 20 34 26

The new navigator has the following features

- Tree navigation list containing all scenes which can be individually opened to show the actors/triggers within each scene
- Accessible with Tab key and full keyboard navigation for navigating the list
- Clicking a scene in the list will focus the world view on that scene
- List is now sorted alphabetically
- List includes color circles showing label color for scene if selected using scene editor sidebar
- Custom Events are now listed in this sidebar so are always accessible rather than needing to click on the background to access the project editor sidebar
- The Global Variables list is also displayed in the navigator, clicking any variable opens a new right hand variable editor sidebar (screenshot below) allowing you to rename the variable (by clicking the header in the sidebar) and provides a list of all the uses of that variable in your project. This can be useful for debugging puzzles to quickly find scenes that are sharing variables.
- Alt clicking the variable select field in any event (if it is currently set to a global variable) will open the variable editor for the selected variable.
- Each section of the navigator is collapsible by clicking the section header
- If you'd prefer not to show the navigator sidebar at all you can drag it completely to the left to hide it, to reenable it use the menu item "View / Show Navigator"

Screenshot 2020-11-07 at 20 51 24

Not that I'm aware of.

* **Other information**:

There is currently an issue with the Variable Uses list showing false positives for the the first few variables, I believe this is because Custom Events store internal variables as the same value as a global variable "0", "1", "2" etc and the walkEvents function is picking up values from within the custom events. @pau-tomas I'm wondering how difficult it would be to get custom events storing Variable A-J as V0-V9 maybe?


pau-tomas:
I pushed a [patch](https://github.com/pau-tomas/gb-studio/commit/efd2969b01df674ed9471828c2760d3c415846a1) to my fork to use `V0-9` as the Custom Event variables index, it still needs some more testing (but it seems to work for both variable dropdown and variables in the text field). It'll also need a migration script. I can open a PR once that's done instead of adding the change to this PR.

One thing I noticed, the default variable names in the navigator start with `Variable 1` instead of `Variable 000`, which makes them inconsistent to the name they have when selected.

chrismaltby:
Oops, good spot on the off by one error and inconsistent left pad on the default variable names! That's fixed now.

Other than those false positives I'm fairly happy with this feature so I'm going to do a bit more testing but I might push this today and we can handle the custom events id rename separately.

Thanks for putting that patch together, looks to work well from quick test :-) Do you want me to have a look at putting a migration script together?

pau-tomas:
As you prefer, if you think the custom events fix can wait and it doesn't block this patch, I'll probably have time to look at the migration later today and add submit a PR with it.

pau-tomas:
Started with the migration script last night, couldn't finish but will continue with it later today.

I decided to rename the variable fields in the Custom Event call events from `$variable[0-9]$` to `__parameter_V0-9` to match the naming convention of Custom Events 2. It makes more sense and the old one was awful naming and I don't know why I went that way. And also, will save some complexity from the Custom Events 2 PR.

PR #706, opened 2021-01-26
Detect itch.io parent, add class and attempt to fix layout style issues

entozoon:
Bug fix for itch.io hosted games

* **What is the current behaviour?**
Rendering is borked on itch.io, if you set your game as `640x576` or larger, because the existing styles force it to 480px wide so it stretches vertically.
Moreover, itch.io sets the iFrame to 640px wide, but an ancestor element obnoxiously enforces a max width of 600px under a breakpoint with the rule:
```css
@media (max-width: 1000px)
.responsive .inner_column {
    max-width: 600px;
}
```
.. which cuts off the right 40px when it's `width: 100%`

Fair enough if you don't want to merge, because it's partly a problem with itch.io, but here's my workaround attempt anyway as I know lots of games end up on that platform and this behaviour was quite unexpected. 

Allows for a 600px wide game on itch.io (the page's max-width for smaller browsers) - which you can [check out here](https://entozoon.itch.io/rumboy) .

Nope, it's pretty safe.

* **Other information**:
* I absolutely love gb-studio. Incredible piece of kit, and have enjoyed using it massively.

### DIFF
--- a/appData/js-emulator/css/style.css
+++ b/appData/js-emulator/css/style.css
@@ -315,3 +315,12 @@ body {
   }
 }
 
+/* itch.io fix for game size 640 x 576 */
+@media only screen and (min-width: 640px) {
+  .detected-itch #game canvas {
+    display: block;
+    width: 100% !important;
+    height: auto !important;
+    max-width: 600px;
+  }
+}
--- a/appData/js-emulator/index.html
+++ b/appData/js-emulator/index.html
@@ -40,5 +40,6 @@
     <script src="js/other/resize.js"></script>    
     <script src="js/GameBoyCore.js"></script>
     <script src="js/GameBoyIO.js"></script>
+    <script src="js/other/itchDetection.js"></script>
   </body>
 </html>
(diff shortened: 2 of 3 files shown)


entozoon:
To be honest though - it does harm to fullscreen mode, and I can see no way of targeting the fact that the iframe has gone fullscreen from within it.. so we're a bit boned if anyone has it enabled in itch.

maxoakland:
> To be honest though - it does harm to fullscreen mode, and I can see no way of targeting the fact that the iframe has gone fullscreen from within it.. so we're a bit boned if anyone has it enabled in itch.

How does it harm Full screen mode?

entozoon:
Stops it from working properly. There'll be a way to get these styles right I'm sure. 
That said, you can contact itch.io and persuade them to let you add custom CSS, which I've done successfully

maxoakland:
> That said, you can contact itch.io and persuade them to let you add custom CSS, which I've done successfully

Can you go into more detail about that? How'd you persuade them and what did you do with the custom CSS?

entozoon:
Yeah I don't see why not @maxoakland 
Sent them an email entitled "Request CSS edit access" explaining that I'm a web developer and certain itch.io layout bugs are messing up my game's rendering. 

Specifically
```css
 .responsive .inner_column {
      max-width: 600px;
  }
```
Because, say you have a game with width 640px (as would be ideal for us), the last 40px of the iframe get cut off by this parent style. Super annoying.

In fact I'll close the issue on this because it is a problem with itch, and this is a fair workaround

Thanks for help, @maxoakland

Issue #662, opened 2020-11-16
Dialogue box disappears between texts


Animacreep:
**Describe the bug**
I have noticed in recent versions that the dialogue box disappears between text boxes.

**To Reproduce**
1. Open GB Studio and make/open a project.
2. Add a dialogue event.
3. Make a lot of text boxes.
4. Run the project.

**Expected behavior**
Dialogue boxes should not disappear.

**Screenshots**
![box](https://user-images.githubusercontent.com/34885906/99240788-48694600-27fd-11eb-8d54-45e2b3656db8.gif)

**Platform (please complete the following information):**
 - OS: Windows 7 Ultimate x64 SP1
 - App version: 2.0.0 Beta 5 (dev builds)

Issue #636, opened 2020-10-28
Faster camera speed


Rebusmind:
**Is your feature request related to a problem? Please describe.**
I want to make a Link's Awakening style screen transition, where the camera moves to the new area. Right now, even the fastest movement option moves only 1px per frame, which makes the transition too long. Options where the camera moves 2px or even 4px would be great. 

**Describe the solution you'd like**
New options for the camera speed would be nice. In order to not break other projects, the other speed options would have to stay as they are, but the caption could be edited to mark the new options as the 'fastest'. 

**Describe alternatives you've considered**
A new system where you could use a slider to decide how many pixels the camera should move per frame would also be cool, but would need every project to be merged properly.

Issue #629, opened 2020-10-20
[V2 Beta] Actor Flies Off-Screen When Changing Movement Speed While Off Tile


patrickmollohan:
**Describe the bug**
Similar to issue #449, an actor will fly off-screen if you change its movement speed while it is not on a tile.

**To Reproduce**
Change movement speed of actor while off tile. A simple way of doing so is either putting the movement speed change in a timer script and have an actor move around randomly or tie it to a button press and press the button in between tiles.

**Expected behavior**
Actor shouldn't fly off screen.

**Screenshots**
Pretty much the same ones from #449 apply here.

**Platform (please complete the following information):**
 - OS: Linux
 - App version 2.0.0-beta5

**Additional context**
I'm sure a similar fix to #455 could apply. However, as @coasterspaul mentioned in #449, the fix creates a new issue in that you can only change the direction if the actor is standing still. The same would be true here (could only change speed if actor is standing still). I tried thinking of ways to have the change direction script wait for the actor to reach a tile:
```
while (!ACTOR_ON_TILE(active_script_ctx.script_actor)) {
  ;
}
actors[active_script_ctx.script_actor].dir.x = script_cmd_args[0] == 2 ? -1 : script_cmd_args[0] == 4 ? 1 : 0;
actors[active_script_ctx.script_actor].dir.y = script_cmd_args[0] == 8 ? -1 : script_cmd_args[0] == 1 ? 1 : 0;
actors[active_script_ctx.script_actor].rerender = TRUE;
```
but this locks the actor permanently in position. Maybe a queue of actions could fix this? As in, if the player is on tile, perform the action immediately (change direction/speed/etc), else enqueue it. Whenever the player reaches a tile, dequeue the actions.

FranticPedantic:
Thanks for clarifying exactly what's causing this. I was experiencing the same issue and was not sure what was causing it.

PR #709, opened 2021-02-02
Fixing beta5

### LINKED ISSUE #711: Can't interact with actors in first 2 columns of the scene
**Describe the bug**
I can interact with actors as long as they are not in the first 2 columns of the map, if the player X coordinate is less than 2 (0 or 1) I can't interact with actors. I'm not sure if it's a problem with the player or the actor, but it happens even if I change scene.

**To Reproduce**
1. Put an actor at position (0, 0) (or 0, 1)
2. Set actor interaction to display a dialogue
3. Interaction from up or down does not work

**Expected behavior**
I should not be able to walk through the actor and I should be able to interact with it.

**Screenshots**
Dg7VI55hFK

**Platform (please complete the following information):**
 - OS: Windows 10
 - GB Studio 2.0

**Additional context**
Had no problem in 1.2 IIRC

RichardULZ:

Bug Fix

Multichoice always picks option 2 when B pressed, some users mistake A and B.
Changing actor speed on attach or mid frame caused issues #629
Emote on simulscript like update not cleared on scene switch

Multichoice no longer uses b for option 2 by default, menu unaffected.
Actor speed events clamp the actor position by 4 pixels, seems to fix the issue for any speed.
Emote sprite 0-1 is cleared on scene start

* **Other information**:
Todo:
- [ ] Update engine version, 
- [x] update changelog.
- Hard fix
- [ ] Ui blocking for simul scripts does not block while closing.
- [x] Textbox Flash of nothing #662 (perhaps need a frame counter if speed instant?)
- [x] Set Palette event with "Don't Modify" breaks Per Scene Palette selector
- [x] Fix #711 
- [x] Better random seeding
- Event changes?
- [ ] Faster camera speeds #636 move 1-4 pixels per frame, on top of 1 pixel every x frames.
- [ ] Switch Scene BG (I promised this, should be easy?)
- [ ] Engine field idea to test, toggle ui_block/Await close, May make textboxes on a timer and textboxes for location while mid game possible, and not interfere with any existing events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,18 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add live preview of how dialogue will appear in game when hovering over "Display Dialogue" events using ascii.png, frame.png and avatar sprites
 - Add character count to dialogue event text input (52 characters max, or 48 if using an avatar)
 - Add autocomplete and syntax highlighting for variables and speed codes in dialogue
-
 - Add Engine Property Fields for platfomer physics, 2d grid size, and fade style in settings
 - Add Event: Engine: Field Update for changing platformer physics mid game
 - Add 16bit variables for engine property fields only
 - Create Engine.json with ejected engines, to add your own custom engine properties
-
 - Add Navigation side bar to left of the editor, Accessible with Tab key and full keyboard navigation
 - Navigation sidebar lists: All scenes, actors/triggers within each scene, Custom Events, Global Variables
 - Navigation Clicking a variable opens a variable editor, to rename the variable, and list all uses of that variable in your project
 - Alt click a variable field in any event (if currently a global variable) to open the variable editor
 - Drag navigator sidebar completely to the left to hide it, to reenable it use the menu item "View / Show Navigator"
-
 - Add GBDK 2020 v4.0.1
 
 ### Changed
@@ -34,10 +31,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Custom Events are now listed in navigator sidebar only
 - Updated to GBDK2020 v4.0.1 for massive performance increase, new ejected engine mandatory
 - Engine bank push pop functions replaced with __banked for performance increase
+- Multichoice no longer uses B for option 2, new users mistook A and B buttons as both chose an option, with little indication of result, menu is unaffected.
 
 ### Fixed
 
 - Fix issue where sidebar tabs could become hidden if translation wasn't able to fit in space available
+- Fix Actor speed event while moving caused the player to walk forever, now clamps to 4 pixels
+- Fix Emote sprites no longer stuck on screen after scene change
+- Fix Textbox flash of nothing between instant transitions
+- Fix Bg palette event "Don't modify" no longer breaks per scene palettes
 
 ## [2.0.0-beta4]
 
--- a/appData/src/gb/engine.json
+++ b/appData/src/gb/engine.json
@@ -1,5 +1,5 @@
 {
-  "version": "2.0.0-e17",
+  "version": "2.0.0-e18",
   "fields": [
     {
       "key": "plat_min_vel",
--- a/appData/src/gb/include/Math.h
+++ b/appData/src/gb/include/Math.h
@@ -8,7 +8,6 @@
 INT16 DespRight(INT16 a, INT16 b);
 UBYTE Lt16(UINT16 a, UINT16 b);
 UBYTE Gt16(UINT16 a, UINT16 b);
-void SeedRand();
 
 #define IS_NEG(a) ((UBYTE)(a)&0x80)
 
--- a/appData/src/gb/src/core/Actor_b.c
+++ b/appData/src/gb/src/core/Actor_b.c
@@ -93,9 +93,9 @@ UBYTE ActorInFrontOfPlayer_b(UBYTE grid_size, UBYTE inc_noclip) __banked {
     }      
   } else {
     if (player.dir.y == -1) {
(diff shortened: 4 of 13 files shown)


patrickmollohan:
Great work! The speed fix seems to work for my projects; no more actors flying off screen. Thanks for fixing this!

RichardULZ:
@patrickmollohan did you try a larger clamp of 8 pixels to fix direction mid tile?
 0xFFFC is 4px,  0xFFF8 would be 8 pixels, don't know if it could get funky still by clamping left/up.
 
 As for the 8 dir top down code, I don't know, that would probably also need an 8px clamp, which is quite a downgrade.
 The Adventure mode which is still missing actor collisions (gbdk2020 update changed actor check maybe?) would probably be better. Or an overhaul of per tile movement, which tries to get to the grid from any pixel, not just wait until it lines up.

chrismaltby:
Hey @RichardULZ Thanks for looking at these. I've been quite preoccupied with the new engine rewrite at the moment over at https://github.com/chrismaltby/gbvm so I appreciate you looking at fixing up some of the 2.0 issues. Just let me know when you think this is ready and I'm happy to merge it.

The more I work on the new version the more I think it will become a 3.0 release as there's quite a lot different. I think it's worth fixing up some of these 2.0 issues though as there will still be plenty of people using this version (I've just today been fixing issues in the 1.X releases for Deadeus even!)

RichardULZ:
Awesome, totally understandable as the gbvm engine really is the future of gbstudio, and it might not have been a reality if 2.0 was completed and progress slowed, it's as big a jump as 1.2 to 2.0.
Calling it 3.0 sounds like the right move, numbers are just numbers anyway.

Apparently the issue on #711 stems from out of date SDCC, where `0xFF == 0x0 - 1` returns false, was recommended to update to GBDK 4.0.2

As far as this pr goes, it just needs an engine version up and engine changelog.

@chrismaltby I'd like your position on things that didn't get finished for 2.0, events and features that may have to be re-replicated for 3.0. (hope it's not too long)
- New events like Switch Scene BG,
- Sound Effect Advanced options from private beta, it was pretty much complete for square waves, but not crash/beep, and needed migration scripts.
- Some speed changes around camera move to, and contemplating perhaps a 3/4 actor speed, would need a migration script, but these could be lived without.
- (making the list longer, set player palette event...)

But Some tricky changes.
### Cutscene/Simultaneous script switch
The concept of Cutscene scripts and Simultaneous scripts are not well documented,
Attach script to button events are now simultaneous instead of cutscenes, which causes problems for dialogue (you can interact/move between boxes), and 1.2 migration.

In 2.0's current design, changing the active script runner to be script 0 should force it to become a cutscene, but would this work in gbvm, or be necessary there?
A checkbox for attached/timer scripts to choose might help, but Events to switch a script between cutscene or simultaneous on the fly seem like a good idea too, it could stop other hit events from occurring if you died by touching an actor, and wanted your player to fall off screen without hitting other actors/running physics, along with safer dialogue, or smoother triggers/interacts.
Since it could cause trickle down issues to 3.0 and gbvm, I'm not sure about it.

### UI Interactable engine field
Another area i've been testing on discord, disables UI blocking, and "script waiting" for a text box, when "ui interactable" field is off.
In a pinch, this open up a lot, timed dialogue, a textbox that opens while you keep walking (town name), a crude hud for in game or debug, camera shake/motion while text is on, but in the grand scheme of things, it would either be best integrated into Set Text Speed event, or tackled in a different way, like a new dedicated text box tool.

It may be safest to keep Ui interactable field as an engine mod?

This pr could be applied as is (plus engine changelog), and new events done separately, or scrapped and moved to 3.0, i'm not sure.

deepx77:
> The more I work on the new version the more I think it will become a 3.0 release as there's quite a lot different.

Will version 3.0 work with project files from for example Beta 4? I just started a week ago with GB Studio.

RichardULZ:
@chrismaltby 
I haven't added bg scene switching, advanced SFX, or the idea of switching between a cutscene/simultaneous script by event.

But the rest of these engine changes have been helping a few people, and should probably be added sooner rather than later. (Snuck in a little palette hex preview change too)

The cli builds are down again, so pushing this should sort it at least for another month.

Note this does not update gbdk, but that should be looked at for minor math bugs (Adventure mode).

chrismaltby:
Thanks for this @RichardULZ Everything you've fixed here looks good and nothing is clashing with intended functionality in 3.0/GBVM.

I'll check back on this PR in a few days, if no issues have been reported I'll push this up to Itch.io as an official release :-)

The new features you're thinking about probably should wait for the GBVM release and there's a few things already in place that should make them easier
- **Cutscene/Simultaneous script** Internally in GBVM all scripts are what you're calling simulatenous, but we've got a command `VM_LOCK` and `VM_UNLOCK` which allows any script to become the "cutscene" script pausing any background scripts until the foreground one is unlocked. By default scene start and interact scripts start with an automatic VM_LOCK but we could allow manual events for doing this which should allow you to do what you're suggesting quite easily.
- **UI Interactable** We've got a flag in the scripting language UI_MODAL which is used when moving the overlay or when waiting for text to be drawn, it's not exposed anywhere yet but I think changing this value should allow doing things like moving around while the text box is open
- **Sound Effects** I've not started integrating this feature yet but we have the ability to play small wav files, can probably do something quite interesting here
- **Player set palette** I've got to make a few changes to how sprite colouring works which will make this a little different. Basically sprite palettes will now be set per scene (like they are for backgrounds already) and each indidual sprite can use multiple/all of the scene colours by using a different palette number per tile. Going to be a bit awkward to get the UI right for this but it should allow potentially 32 colours in a single sprite

maxoakland:
> * **Player set palette** sprite palettes will now be set per scene (like they are for backgrounds already) and each indidual sprite can use multiple/all of the scene colours by using a different palette number per tile

This is awesome. I was hoping for something like this

PR #749, opened 2021-05-16
Add a mapping for Japanese encoding Shift_JIS and font file

gmarty:

This adds a mapping for the Shift_JIS encoding used for the Japanese language and an example of a font file in PNG format.
I understand this branch is actively being worked on and an alpha release is months away, but this PR doesn't contain code change so can be considered low risk to land.

The IDE doesn't currently allow typing in Japanese. Users usually resort to 3rd party tools, which complicates the workflow.

By choosing `Shift_JIS` in the UI Settings and selecting a compatible font image (e.g. the `k8x12S` provided), the users can type in Katakana directly from the editor.

No, it doesn't contain any code change.

* **Other information**:
There are few caveats:
  * Only katakana are supported. This is also true for the [Shift_JIS encoding](https://en.wikipedia.org/wiki/Shift_JIS#As_defined_in_JIS_X_0208:1997).
  * Katakana "mutators" (dakuten and handakuten) have to be placed on top of non-voiced characters, ideally on the previous line. This makes it possible to drastically reduce the memory needed.
  * The preview in the editor doesn't currently support previewing alternative encodings, but the generated ROMs are correct.

The mapping contains half-width and full-width alphanumeric characters and katakana. This gives a more user-friendly experience. It means, for example, that `ニホンゴ` and `ﾆﾎﾝｺﾞ` will result in the same output.

The Japanese font used in the sample PNG file is k8x12S, a free font available from https://littlelimit.net/font.htm.


untoxa:
can you please also add proportional version?

gmarty:
> can you please also add proportional version?

There's no variable width version for Japanese fonts (outside of alphanumeric and punctuation signs). All the characters are proportioned to more or less fit in a box, so they all have similar widths and heights. If you look at the text samples in https://littlelimit.net/k8x12.htm, you'll see that it is similar to a monospaced font.
Plus that would not work here with the dakuten and handakuten that need to be placed exactly at the same position on the previous line.

In the future, I want to work on a more complete font that will contains hiragana and katakana with integrated dakuten/handakuten as well as some kanjis, but that's a huge piece of work. I will need help to understand how to build variable width fonts too.

chrismaltby:
That's great thanks :-)

That reminds me, I need to work out a good way to make it clear the specific encoding a font should use.

I'm considering doing something like having an optional .json file named the same as the .png in `$PROJECT/assets/fonts/` which contains the encoding per font even (so the same data you've got in `shift_jis.json` would be moved to `k8x12S.json`, that way we can even have multiple encodings per project without the user needing to understand the mapping (if someone provides them with a png/json pair), potentially even multiple encodings in a single text box if you use the new font switching text codes.

gmarty:
> I'm considering doing something like having an optional .json file named the same as the .png in `$PROJECT/assets/fonts/` which contains the encoding per font even (so the same data you've got in `shift_jis.json` would be moved to `k8x12S.json`, that way we can even have multiple encodings per project without the user needing to understand the mapping (if someone provides them with a png/json pair), potentially even multiple encodings in a single text box if you use the new font switching text codes.

YES!! I experimented and played a lot with this yesterday and my conclusion is to support complex languages like Japanese or Chinese GB Studio will need to:
* Allow arbitrary mapping per project (doesn't have to be standard, more like a user builds a font using the characters they want to use)
* Allow using several fonts in the same dialogue box (to overcome the 256 characters limit)

I hope this last point is possible, although there is a UX challenge associated with it to make it user friendly and readable.

cubicstyle:
Position of the "Dakuten" and "Han-dakuten" in the font file is incorrect.

ja_font_mistake

untoxa:
They are supposed to be in the bottom of the upper line, i guess.

gmarty:
> They are supposed to be in the bottom of the upper line, i guess.

Correct. The font used, k8x12S, is a 8x12 pixel font. The kanji are 12 pixel height, but the kana are 8 pixel height + the dakuten and handakuten to be placed on the line above like this:
```
゛
カ
```

Then it will be rendered as intended by the font authors.

cubicstyle:
When add Dakuten, use two characters in a row. So Dakuten should be at the upper.

ガ ＝ カ゛＝ ｶ˝

PR #760, opened 2021-06-01
[v3] Editor Sidebar TS migration

crutchcorn:

This PR refactors the `CustomEventEditor` and `EditorSidebar` components to use TS & functional hooks

N/A

No

* **Other information**:

Let me know if there's anything I need to change. Looking through other TS components, I suspect one of the requests might be to make these functions into `useCallback` instances. I'm also not sure how we want to handle the `might be undefined` type checks. Right now I'm silencing them with `!`, but I know that's generally not a great practice.


chrismaltby:
Thanks @crutchcorn

I removed those non-null assertions, eslint should have been throwing warnings with those (`npm run lint` though there's some things Pau is working on that's also showing warnings at the moment) but other than that this is all good!

PR #761, opened 2021-06-01
[v3] window zoom

crutchcorn:

I admittedly don't know what tests to add (if any) or what I should do for docs on this PR

Users may have difficulties interacting with the UI due to element's sizes. They may prefer to zoom in or out of the application from their OS' default scaling rules. This PR adds the ability to zoom in or out of the application using a new item in the menu bar

Before window scaling:

After window scaling:

This feature will also be incredibly helpful when doing livestreaming (or other content creation) of GBS development. I, for example, have a hi-res monitor but want to be able to zoom in on the app for mobile users to be able to see the contents of the engine. This PR makes this possible

You're unable to change the application's scale without also modifying OS level scaling

Adds ability to zoom in the app

No breaking changes

* **Other information**:

It would be ideal to store this data and load the app with this level of scaling applied at launch. Would love some guidance on how y'all might think this could be done - I've never used Electron prior to this (although I have built apps with React Native)


chrismaltby:
Nice idea, I'll check this branch out later, thanks!

> It would be ideal to store this data and load the app with this level of scaling applied at launch. Would love some guidance on how y'all might think this could be done

Check out how electron-settings [0] is used in `initProject.js`, basically it's a key/value store (similar to localStorage) that's persistent between launches, so where you're setting the zoom level you could also write to electron-settings, and on page load you could read the value (if present) to set the initial zoom to the previous value.

[0] https://www.npmjs.com/package/electron-settings

chrismaltby:
I've just had a play around, seem to work well but I think it's confusing to have this on the Window menu with the exact same wording as the View menu, would be quite easy to confuse the two.

I think the feature probably belongs on the Preferences window rather than the menu, and probably doesn't need a keyboard shortcut either.

Aseprite, had a "UI Elements Scaling" setting in their preferences window, maybe we could use that same wording here

Screenshot 2021-06-01 at 18 23 20

Could probably just have a dropdown with `100%`, `200%`, `300%` and `400%`.

crutchcorn:
Works for me. Where do we want that dropdown? Maybe in the UI elements?

UI elements

chrismaltby:
I was actually thinking the application preferences window, rather than the project settings section. I think it's under the Edit menu on Windows

preferences

maxoakland:
> I was actually thinking the application preferences window, rather than the project settings section. I think it's under the Edit menu on Windows
> 
> preferences
> 
> check out `src/components/app/Preferences.tsx`

I agree that it should be in the application preferences rather than the Window menu. It’s more consistent with other applications

RichardULZ:
This would be a great addition!
also a plus for smaller low resolution screens too, where you want a bit more space at the cost of squinting.

crutchcorn:
I admittedly didn't even know about the preferences window 😅😅

I'm fairly busy the next few days, but when I get the chance I'll update the PR to include persistence (thanks for the guidance there btw) and move the UI to the preferences window

crutchcorn:
Just migrated the zoom functionality to the preferences screen

However, I'm noticing that the preferences screen seems to be too short to display the select

image

I don't know what would cause this - would anyone be able to help guide me on this? I'm not sure how window sizing works in Electron

chrismaltby:
Perfect, thanks!

To resize the window you'd want to change the values here https://github.com/chrismaltby/gb-studio/blob/f728db72e49e6a0b099c59eef45e6defba835dd7/src/main.ts#L88-L90

But actually I've just moved the two default editor dropdowns to be side by side and I think that works nicely

Screenshot 2021-06-12 at 12 54 35

This all looks good now so I'm happy to merge :-)

maxoakland:
> I think the feature probably belongs on the Preferences window rather than the menu, and probably doesn't need a keyboard shortcut either.
>

I agree with this and was going to say the same thing

PR #764, opened 2021-06-03
[v3] Fix clipboard on Windows

crutchcorn:

While testing GBS 3, I noticed that none of the copy+paste behavior worked as-intended

After doing a bit of debugging, I saw that `readBuffer` wasn't giving a proper output after the data was being stored 

image

It turns out this was an upstream bug in the version of Electron we were using:

https://github.com/electron/electron/issues/24253

And a fix was published in version `8.4.1`:

https://www.electronjs.org/releases/stable?version=8&page=2#8.4.1

No copy+paste action worked on Windows

Copy+paste now works in Windows

No breaking changes

* **Other information**:

Side note: I noticed that we're a fair few stable releases away from the most recent stable Electron version. I'd be happy to do some testing to see how far we're able to upgrade our dependencies - is that something that we'd want to do, or is there something I'm not aware of when it comes to our dep versions?

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "@typescript-eslint/parser": "^4.25.0",
     "babel-eslint": "^10.1.0",
     "css-loader": "^3.0.0",
-    "electron": "8.2.5",
+    "electron": "8.5.5",
     "eslint": "^7.21.0",
     "eslint-config-airbnb": "^18.1.0",
     "eslint-config-prettier": "^6.11.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4289,10 +4289,10 @@ electron-winstaller@^5.0.0:
     lodash.template "^4.2.2"
     temp "^0.9.0"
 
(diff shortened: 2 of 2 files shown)


chrismaltby:
Everything seems to be working with this Electron release so happy to bump up to this one if it fixes some issues on Windows.

On the topic of the clipboard, I need to go through and migrate all the really old copy/paste functionality (scenes/actors/sprites/events) to work the same way as the new clipboard functionality (metasprites/tiles). The main difference is the older functionality ends up with JSON data in the txt clipboard and you can end up accidentally pasting a huge blob of data into a text field if you're not careful.

crutchcorn:
@chrismaltby if you'd like, I can actually take up the migration from old copy/paste functionality to the new one. During debugging, I dived pretty deep into how the new method works and I am pretty comfortable with migrating things. 

If that'd help, would it be possible to make a GH issue with a checklist of things to migrate? Or even just drop a checklist here?

PR #766, opened 2021-06-03
[v3] Upgrade major Electron deps

crutchcorn:

This PR upgrades all Electron deps to their latest major version. This includes:

- `electron@8` to `electron@13`
- `electron-settings@3` to `electron-settings@4`

Looking at the upcoming breaking changes in v14, I think we can safely upgrade without codebase changes when that's released

No breaking changes

* **Other information**:

While I've tested my game, opening dialogs, building the ROM, and a few other behaviors - I would strongly encourage some manual QA work to double-check that no functionality broke


crutchcorn:
Actually, something I want to highlight in this upgrade is that Electron 11 added support for M1 Macs!

https://www.electronjs.org/blog/electron-11-0

I admittedly don't have an M1 Mac to test this, but it appears that we can build GBS 3 for that platform in the future

https://www.electronjs.org/blog/apple-silicon

crutchcorn:
I think I misunderstood what `remote` was doing. I strongly suspect that I've broken a fair bit of functionality. Revisiting now

crutchcorn:
OK this PR should be fixed now. Again, though, let's make sure to do some thorough manual QA testing

chrismaltby:
Definitely would be good to get on the latest Electron, especially since I'm using an M1 Mac myself!

Had a look around and fixed a few issues on the Preferences window.

Not sure what it's like on Windows with this release but it's noticeably much slower running from this branch, even when running a production build. e.g. If I click any React Select elements there's around a 1 second delay before the menu opens (!!) and loading a project seems to take at least a few seconds longer than it would normally. Might need to wait for another point release or could just need playing around with the `webPreferences` values in main.ts.

I'll hold off merging this but lets keep this PR open for now.

PR #777, opened 2021-06-27
[v3] SceneInfo TS migration

crutchcorn:

This PR refactors the `SceneInfo` component to use TS & functional hooks

N/A

No

* **Other information**:

- [ ] Scene
	- [x] SceneInfo (this PR)
	- [x] EventHelper (#778)
	- [x] Actor (#779)
		- [x] SpriteSheetCanvas
	- [ ] Trigger
	- [ ] SceneCursor
	- [x] SceneCollisions (#780)
	- [ ] ColorizedImage

### DIFF
--- a/src/components/world/SceneInfo.tsx
+++ b/src/components/world/SceneInfo.tsx
@@ -1,7 +1,6 @@
-import React, { Component } from "react";
+import React, { MouseEventHandler, MouseEvent } from "react";
 import ReactDOM from "react-dom";
-import PropTypes from "prop-types";
-import { connect, useSelector } from "react-redux";
+import { useSelector } from "react-redux";
 import cx from "classnames";
 import debounce from "lodash/debounce";
 import {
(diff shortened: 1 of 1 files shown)


src/components/world/SceneInfo.tsx
```diff
@@ -0,0 +1,456 @@
...
+        // (1) component has been unmounted (see isMounted in snippet)
+        // (2) delay has changed
+        if (inputsRef.current.delay === delay && isMounted())
+          inputsRef.current.cb(...args);
+      },
+      delay
+    ),
+    [delay, debounce]
+  );
+}
```
crutchcorn, line 65: This code _was_ originally pulled from SO but I am confident in its abilities. Pretty familiar with how `useRef` is doing things here, as [I wrote a fairly in-depth article about how `useRef` functions w/ dep arrays](https://unicorn-utterances.com/posts/react-refs-complete-story/)

That said, I'm not _super_ happy with this code living here. Would love to move to a helpers file of some kind. Not familiar enough yet with the codebase structure to decide where that should go. Any thoughts @chrismaltby ?

crutchcorn: One more reason it would be nice to keep these hooks elsewhere would be because we'll likely be reusing this logic for debouncing the function when we migrate `ColorizedImage`'s `draw` function:

https://github.com/chrismaltby/gb-studio/blob/0e5f99ca4b3935f994ce56d24a5b3613892d37f6/src/components/world/ColorizedImage.js#L35

crutchcorn:
Oh! After speaking with @pautomas in the Discord, it seems like this component may have already been migrated to TS:

https://github.com/chrismaltby/gb-studio/blob/wip/script_editor_v2/src/components/world/SceneInfo2.tsx

In particular, it seems like there's an existing `useDebouncedCallback` created in this branch

https://github.com/chrismaltby/gb-studio/blob/wip/script_editor_v2/src/components/world/SceneInfo2.tsx#L31

There's a pretty different implementation between how things were done in this PR and how they were done in that branch. I'm wondering if we wouldn't want to use an implementation similar to the one found in _this_ PR instead?

In particular, here's how you would use the `useDebounceCallback` in the `wip/scene_editor_v2` branch:

https://github.com/chrismaltby/gb-studio/commit/0c3d956ad7dafa670be15273efcba72a4623ae9f#diff-13c6f0fa2e1acd09a1e69d8773c7b6f023a954ef1f96c3163f43d62f0816c25aR117-R259

In order to keep an empty dep array (in order to avoid the reassignment of the function to `debounce`), we have to pass in _everything_ to the denounced function's usage. On the flip side, this PR's implementation allows us to track local variables in our function and not have any issues with `debounce` being assigned repeatedly:

https://github.com/chrismaltby/gb-studio/pull/777/files#diff-3e07ebc5dafc3ecbfd8bdd40ca13137cddab823bea21eff541a09282b709ccedR268

This is because we're relying on `useRef` to not re-render when the variable is assigned with `useEffect`.

That said, the implementation in the `wip` branch is significantly smaller and requires less arcane knowledge of `useRef` to maintain:

https://github.com/chrismaltby/gb-studio/commit/0c3d956ad7dafa670be15273efcba72a4623ae9f#diff-c5361c11f69cf547da2fa6cf52480558af29780b4b18563cd5ab15dc08a34da9R6-R12

vs.

https://github.com/chrismaltby/gb-studio/pull/777/files#diff-3e07ebc5dafc3ecbfd8bdd40ca13137cddab823bea21eff541a09282b709ccedR47-R65

All in all, I would argue that this PR's implementation of `useDebounce` reduces the potential to improperly use `useDebouncedCallback` by forgetting its implementation details and adding a dep array. However, totally understand if we don't want to move forward with how things are as I implemented - just figured I'd mention a different way of implementing that I personally prefer.

chrismaltby:
Hey @crutchcorn yeah I think you're right about the useDebounce hook being a little nicer in this PR, I'll update my version to use this instead, thanks :-)

PR #778, opened 2021-06-27
[v3] Event Helper TS migration

crutchcorn:

This PR refactors the `EventHelper` component to use TS & functional hooks

N/A

No

* **Other information**:

- [ ] Scene
	- [x] SceneInfo (#777)
	- [x] EventHelper (This PR)
	- [x] Actor (#779)
		- [x] SpriteSheetCanvas
	- [ ] Trigger
	- [ ] SceneCursor
	- [x] SceneCollisions (#780)
	- [ ] ColorizedImage


src/components/world/EventHelper.tsx
```diff
@@ -0,0 +1,118 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const isShapedArg = (v: any): v is Arg<any> => v?.value;
+  if (isShapedArg(arg)) {
+    if (arg.type === "variable" || arg.type === "property") {
+      return undefined as never;
+    }
+    return arg.value;
+  }
+  return arg as never;
+};
```
crutchcorn, line 32: This code I wrote is frankly disgusting. I sort [nerd sniped](https://xkcd.com/356/) myself on this one. However, it's also _technically_ correct ([the best kind of correct](https://www.youtube.com/watch?v=hou0lU8WMgo)) and is the most type-strict variation of this logic present.

In particular, this line:

```
T['type'] extends 'variable' | 'property' ? undefined : R
```

Will literally not run unless an `arg` is defined with `as const` (or if the root typing is strict and not just assigned to `string`).

For this reason, no problem if we want to cull the typings down to something more reasonably maintainable. Apologies for even committing this grossness

PR #779, opened 2021-06-27
[v3] Actor TS migration

crutchcorn:

This PR refactors the `Actor` component to use TS & functional hooks

N/A

No

* **Other information**:

- [ ] Scene
	- [x] SceneInfo (#777)
	- [x] EventHelper (#778)
	- [x] Actor (this PR)
		- [x] SpriteSheetCanvas
	- [ ] Trigger
	- [ ] SceneCursor
	- [x] SceneCollisions (#780)
	- [ ] ColorizedImage


src/components/world/Actor.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, [dragActorStop])
+
+  const onMouseDown: MouseEventHandler = (e) => {
+    if (props.editable && e.nativeEvent.which !== MIDDLE_MOUSE && actor) {
+      dragActorStart({ sceneId, actorId: actor.id });
+      setTool({ tool: "select" });
+      window.addEventListener("mouseup", onMouseUp.current);
+      isMouseListening.current = true;
+    }
+  };
```
crutchcorn, line 97: This is a particularly tricky bit of code. Ideally, we should probably look into refactoring how this is done long-term to be a bit more clean. However, this code is as direct of a port of the class-based functionality as I could manage. 

I did a tiny bit of extra legwork to make sure that we're doing cleanups and re-assigning `onMouseUp` when `dragActorStop` is changed. This should at least make this code stable enough to use in production until we can rethink how to handle this logic

PR #780, opened 2021-06-27
[v3] SceneCollisions TS migration

crutchcorn:

This PR refactors the `SceneCollisions` component to use TS & functional hooks

N/A

No

* **Other information**:

- [ ] Scene
	- [x] SceneInfo (#777)
	- [x] EventHelper (#778)
	- [x] Actor (#779)
		- [x] SpriteSheetCanvas
	- [ ] Trigger
	- [ ] SceneCursor
	- [x] SceneCollisions (this PR)
	- [ ] ColorizedImage

### DIFF
--- a/src/components/world/SceneCollisions.tsx
+++ b/src/components/world/SceneCollisions.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import {
   COLLISION_TOP,
   COLLISION_ALL,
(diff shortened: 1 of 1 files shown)


chrismaltby:
Great 👍 I've made a few minor style changes but other than that all good!

I've been trying to remove the CSS/BEM classes as I rewrite components replacing them with Styled Components instead (eventually I'm planning to delete most of the CSS files) so I was going to migrate the wrapper `div` but it turns out it served no purpose and didn't have any styles targeted to it anyway so I've just removed the div instead.

PR #784, opened 2021-07-04
Added note about Windows cloning with symlinks

crutchcorn:

While working on fixing our tests, I noticed that I had cloned the project without symlinks (as it requires admin mode + special command to clone that way on Windows). As such, our `l10n` tests were failing.

This PR adds a documentation note of this to our `CONTRIBUTING.md` document so others don't end up reporting errant test failures on Windows dev environments.

N/A

N/A

No

* **Other information**:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,14 @@ Fork, then clone the repo:
 git clone git@github.com:your-username/gb-studio.git
 ```
 
+> If you're using Windows, you should clone using Symlink support, otherwise your tests will fail.
+>
+> To do this, you must've enabled Symlink support when installing Git for Windows.
+>
+> With it enabled, you must run your `git clone` command like this, with admin privileges:
+>
+> `git clone -c core.symlinks=true git@github.com:your-username/gb-studio.git`
+
 Install it:
 
 ```bash


chrismaltby:
Ah I didn't realise that symlinks in git required special permissions for Windows, scanning through the codebase I only found two symlinks, both just used for l10n
```
./src/lang/it.json -> it-IT.json
./src/lang/fr.json -> fr-FR.json
```
I'm not actually sure what conditions would cause `it.json` to be used rather than `it-IT.json` and same for FR. It's possible we only really need one. Alternatively we can just duplicate these files in the codebase. Just thinking, given it's only two files it might just be better to disallow using symbolic links if it makes it easier for Windows users to contribute/run tests.

What do you think? Happy just to merge this copy change if you think it's enough but I'm definitely open to just making the change so that no extra permissions are even needed

crutchcorn:
Yeah, to be honest - I think that's the right move as well. At the time of opening the PR, I wasn't sure what the purpose was around the namings of these two files, so I figured I'd open a PR and we could go from there :) 

If you end up renaming, feel free to close this PR instead

chrismaltby:
Cool, I've switched them from being symlinks to just being copies of the files now as of 6899c014e39c1e6f226ff2f2a97d253139fa28d0 I'll close this for now but let me know if the v3alpha branch still has issues when running Windows

PR #786, opened 2021-07-10
WIP trigger on leave script

pau-tomas:

Feature

Triggers have one script that execute when the player steps into them

Added a new script for triggers that executes when the player steps out of them

Yes. A migration script needs to be added for old triggers to have an empty "on leave" script. Otherwise the app crashes when selecting the trigger. 

* **Other information**:

I submitted a PR with the changes for gbvm and a sample project https://github.com/chrismaltby/gbvm/pull/107


src/store/features/entities/entitiesTypes.ts
```diff
@@ -99,7 +99,7 @@ export type Trigger = {
   width: number;
   height: number;
   script: string[];
-  leaveScript: string[];
+  leaveScript?: string[]; // Optional: Old versions of GBS don't have this script
```
pau-tomas, line 102: @chrismaltby let me know if you prefer a different way to deal with this case. Maybe a migration script?

appData/src/gb/include/gbs_types.h
```diff
@@ -1,9 +1,7 @@
 #ifndef GBS_TYPES_H
 #define GBS_TYPES_H
 
-#ifdef CGB
-    #include <gb/cgb.h>
-#endif
+#include <gb/cgb.h>
```
chrismaltby, line 4: Is this ifdef removal needed for the feature?

src/store/features/entities/entitiesTypes.ts
chrismaltby: Yeah I think a migration script so we can expect all triggers to at least have an empty array would be preferable. With the latest v3beta branch you no longer need an "EVENT_END" event in the array for it work so a migration to just add `leaveScript: []` to every trigger should be enough :-)

chrismaltby: v3alpha even, you know what I mean :-)

appData/src/gb/include/gbs_types.h
pau-tomas: Ooops not sure why it's here

PR #801, opened 2021-08-28
Add preview view to menu event

pau-tomas:

Feature

Menu events don't have a preview like the dialogue events do. Also, menu items are limited to 6 characters.

Menu events now have a preview when selected. The character limit has been removed as the preview displays (more or less) when the menu will look broken in-game.

"more or less" because the preview doesn't take into account if the game runs on DMG or not and the tile limit for text is different based on that so, in some cases, the preview and the game will not match as the text will overflow in-game but not on the preview. Like the `8` in the following example.

image

@chrismaltby not sure what's the best course of action here, we could keep the character limit per item but extend it to a text line in monospaced font (16 characters IIRC) but this won't be very useful once non-monospaced fonts are used (see `Other Information` below).

No

* **Other information**:

I also added support for font changes in menus, but not the autocomplete functionality that dialogues have as it seems like something that will require some substantial refactor to be able to be reused. 

image

### DIFF
--- a/src/components/script/MenuPreview.tsx
+++ b/src/components/script/MenuPreview.tsx
@@ -3,28 +3,18 @@ import uniq from "lodash/uniq";
 import React, { FC, useEffect, useLayoutEffect, useRef, useState } from "react";
 import { useSelector } from "react-redux";
 import { assetFilename } from "lib/helpers/gbstudio";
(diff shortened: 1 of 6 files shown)


maxoakland:
This is really nice progress to see

PR #864, opened 2021-12-12
3.0 Fix uge playback, .ARG10-16

RichardULZ:

Bug fixes

Clicking outside of gbstudio while in the music editor will close the music engine, 
previewing uge songs in events would cause multiple listeners to be added and corrupt playing sound.
Using large custom events may ask for arguments over 9 and cause an error,
missing \ key and -+ from openmpt uge keymap

Clicking outside a window will not stop audio preview, only pressing pause on an event or closing the music editor tab. 
Previewing in events no longer corrupts after previewing lots of songs
large custom events can have up to 15-16 arguments (actors + variable holders) and not crash. (it seems global var reference was not added... please look into it some time? 🥺 )

toxa noted nesting events may cause issues with the 64 arg stack, so limiting them to 16 was advised (letting larger ones still have a compile error), This would be better with hard global var references, so less stuff needs passing.

Music from event previews will keep playing until pause is pressed, perhaps would be better if event tab was changed?

* **Other information**:
.ARG0 from math clamp to 0-255 is not fixed as part of this pull, pautomas said something?
Double check if .ARG10-16 are ever included in another patch.


maxoakland:
Being able to reference global variables from within custom events would be a game changer

pau-tomas:
We should merge this change as it'll help in most of the cases, however Custom Scripts allow 10 variables and 10 actors as parameters by default, meaning it'll get to `.ARG21` if I understand this well which will still crash. 

What about expanding to `.ARG21` and add a warning at compile time if any script uses anything above `.ARG16`?

PR #910, opened 2021-12-19
Add "On Move" script to Triggers

pau-tomas:
Please check the Other Information section below as I'm posting this more as an open discussion rather than something I think it should be merged. 

New feature

Trigger scripts run when the player enters or leaves the trigger, once the player is moving inside the trigger no scripts are executed. In 2.0 (when there was only one `On Trigger` script) this wasn't the case, as the `On Trigger` script was running on each step of the player inside the trigger. This was used to add features like random encounters.

Add a new `On Move` script to the Trigger that executes when the player moves around the trigger.

image

No

* **Other information**:

I'm not 100% sure this is needed anymore. The "random encounter" use cases can be achieved with an `On Update` script in an invisible actor pinned to the screen. Not sure if there's other use cases that this script type can fulfill. Open to discussion, really.


maxoakland:
I think this is a good idea and it should be added unless it causes noticeable performance problems

I used it for several unconventional reasons with my game, including making a waterfall that splashes water on the player’s head. The trigger detects where the player is and updates the position

It also allows a trigger to be used for more than one trigger, by checking the location of the player when triggered by a move and acting based on that location. I used that because I ran out of triggers on my map and so several doors are actually one trigger

Anima-SodoDev:
Although `On Update` can be used, I think the old method should be available, since it's better optimized for DMG devices (as it only runs when you take a step instead of constantly). Some 2.0 projects also use it.

I don't see any problem with adding this back to be honest.

notadeadmoth:
I'm making a script that hides sections of my player sprite when it's behind background objects and this would make it so much smoother than "On Update"!

Anima-SodoDev:
I ran into problems regarding the collisions and copy/pasting between tabs when testing.

I had an actor with a 16x16 collision box and the grid mode set to 16px. For some reason, I had to move one block up for the trigger to start working correctly, and even then, I remember I ran into some strange issues.

As for the tabs, I can't seem to copy entire scripts from On Move to On Enter.

IRIS-Kane:
I really wish I could have a trigger activate each time the player takes a step. It would save me needing to use 4 variables and keeping track of player's current and new position to detect when they move.

maxoakland:
> I really wish I could have a trigger activate each time the player takes a step. It would save me needing to use 4 variables and keeping track of player's current and new position to detect when they move.

I would like it too. I use it pretty often in GBStudio 2

chrismaltby:
Closing this one as it's quite out of date now. I think since this PR the Threads functionality would allow support for the feature without needing any IDE changes too

- On Enter, start a thread which runs a loop (make sure to include an Idle event to prevent an infinite loop freeze)
- In the loop run your "On Move logic" if needed you can add extra logic to confirm if the player has moved, or if you just want to run every frame you're inside the trigger just don't check at all
- On Leave, stop the thread

I'm sure this could be wrapped in a plugin if needed too

PR #938, opened 2021-12-25
Global variables in Custom Scripts

pau-tomas:
**DRAFT** need to do some further checks and fix the tests and add migration.

Feature

Global variables can't be used inside custom scripts, limiting their use cases. 

Global variables now appear in the variable selector inside Custom Scripts.

custom_script_global_vars

Hopefully not. But it requires a migration script, as now the "parameter" variables use `Vn` as the id as opposed to just `n` to differentiate them from the global variables. 

* **Other information**:

This draft is built on top of #818 which should be much safer to merge before this one.


chrismaltby:
Closing since we've now merged this in #1143

PR #939, opened 2021-12-25
[Tracker] Set default start playback position 

pau-tomas:

Feature

- Start playback position for song preview is always the start of the song. 
- When Play is pressed the song will start from the last played position. There's no option to restart the song from the beginning. 

Allows to set the start playback position by clicking on the top bar on piano roll or first column on the tracker view.

playback_position

Also added a Stop button that resets the current position to the set one, and modified the play button so if `Alt` is pressed it'll restart from the set playback position.

No

* **Other information**:

@chrismaltby I'm struggling to have the position bar in the piano roll to be "sticky" when the grid scrolls vertically. Any suggestion on how to do that? Not a blocker IMHO but it'd be a nice to have.


maxoakland:
I’m so glad you implemented this!

chrismaltby:
Very nice! I've just pushed an update to make that header sticky too 👍

PR #940, opened 2021-12-26
[Tracker] Selectable keyboard layouts for tracker

pau-tomas:

Feature

Key bindings for the tracker note/instrument/effect input are limited to the OpenMPT / Linear style 

Add a setting to change the layout between Linear and Piano (like Milkytracker)

No

* **Other information**:

- I have some doubts about the naming of the options: Piano / Linear might not be very well-known but also don't know if using the OpenMPT / Milkytracker directly as the options labels is the best so I opted for a mid-level solution image
- The keybinding system is suppose to be flexible and could be used in other places and, eventually, offer a way to customize the keybdindings by modifying a json file and adding it somewhere in the system.
- I don't think the current keybindings will work in non-QWERTY code layouts, I need to investigate using `.code` instead of `.key` on the keyboard events as it seems like that might be a better option. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code


maxoakland:
With the naming: Would it be possible to show the images you made above as the way you select it. Like this: 
image

chrismaltby:
Nice! I agree with @maxoakland could do with explaining what the shortcuts are, but rather than cluttering the preferences window for now, how about we add those two images to the docs site and add a small help button next to the dropdown in Preferences to link to the correct portion of the site maybe?

If the preferences gets much bigger might need to revisit the layout though and at that point it might have a bit more room to add things like the images

PR #981, opened 2022-01-10
Support for Apple Silicone

SileemFarghaly:

Update Electron to v11, this is the first version that natively supports Apple Silicone (m1, m1pro, m1max)

GB Studio currently runs using Apple's Rosetta emulator

GB Studio runs natively on Apple Silicone

* **Other information**:
This is most of the work to get GBS running natively on M1, including native binaries for the build tools.
I have been running into an issue with the GBS build system. 
`- Error: ENOENT: no such file or directory, open '/var/folders/p8/n5vkb7qn7fn7hp0z3bp3846r0000gn/T/_gbsbuild/obj/absolute.o'`
I've been trying to figure this one out for a few days, but I'm lost, it's like its not running make in the tmp directory?


SileemFarghaly:
@chrismaltby Just wondering if you had any idea what could be going on with the GBS Builds? If I take what gets generated into the tmp folder and run make, it seems to generate that missing `absolute.o` file.

chrismaltby:
Hi @SileemFarghaly Thanks for looking at this, I use an M1 Mac myself so am definitely interested. My dev environment is running completely in Rosetta2 so I can't really debug this right now though.

The things I'd check are logging that it's using the right build tools path to get the gbdk you placed in `darwin-arm64` at:
https://github.com/chrismaltby/gb-studio/blob/406ba3848fcc68411676c8c3a991ba9c2ac81602/src/lib/compiler/makeBuild.js#L32

and add some logging around here where it uses GBDK to compile the .c files to .o files
https://github.com/chrismaltby/gb-studio/blob/406ba3848fcc68411676c8c3a991ba9c2ac81602/src/lib/compiler/makeBuild.js#L95-L99

You might want to try clicking the "Empty Build Cache" button too (or just delete the tmp/_gbsbuild folder) as I'm wondering if you've got some cached data interfering.

Where did you get the ARM build of GBDK-2020 by the way? There's been quite a few changes to it recently, I'm wondering if your version maybe is missing a feature preventing it from compiling `absolute.c`, though actually you said you've managed to build using the Makefile so it's probably not that. Either way if we want to support darwin-arm we should probably see if the GBDK-2020 can add official builds.

On another note I'm surprised how few changes were needed to switch to Electron 11, we tried updating to a newer version a while back and ended up with quite a few broken things and the UI weirdly running noticeably slower.
The only issue I've spotted so far in the Mac Intel version of this build is the "About GB Studio" window is broken (probably needs an update for the module we're using).

SileemFarghaly:
Thanks @chrismaltby add a bit of debugging around there later today just to see if I can suss out whats going on.

Good catch on the about window, updating that package fixed it.

As for GBDK, I built it from source and hoped for the best honestly. It seems to work fine in my environment, but I can't  be sure that its not actually running in Rosetta mode silently.

SileemFarghaly:
Just a quick update - I know its been more than a week. I narrowed down the issue to 
https://github.com/chrismaltby/gb-studio/blob/406ba3848fcc68411676c8c3a991ba9c2ac81602/src/lib/compiler/makeBuild.js#L16
always being 0 on m1 macs. Looks like it was a known issue with Electron 11.

chrismaltby:
A bit late but closing this as Apple Silicon is now supported as of the latest GB Studio betas.

PR #995, opened 2022-01-15
Fix keyboard accessibility for add/remove buttons in form fields

rik-smeets:

It allows users to use the keyboard to select the '+' and '-' buttons in form fields, for example in the `Display dialogue` event text areas.

The buttons are inaccessible by keyboard.

The button are accessible by keyboard.

No.

* **Other information**:
The minor CSS changes ensure the buttons match the styling in the application.


chrismaltby:
Ah good point, thanks! I made some slight changes to the buttons (and switched them to use the same Plus/Minus icons used in the rest of the app now) as the text alignment put the labels a little too low but other than that this works great!

PR #996, opened 2022-01-15
Enable 16bit inputs in events that set or check a variable value

Rebusmind:
Feature Change

Events that set or check variables allow only inputs between 0 and 255.

These events have been updated to allow numbers between -32768 and 32767. 

Clamping is unchanged, so older projects that use this should still work the same. Projects that use variable wrapping in creative ways could have unexpected behavior (but that should already have been the case before this PR).

* **Other information**:
There seems to be a bug in the codegen of the math event that is unrelated to this PR that affects the wrapping after the value exceeds the limit.


maxoakland:
I'm really excited about this one

PR #1008, opened 2022-01-19
[Events] Add missing actor label in eventActorShow.js

ReyAnthony:
I was looking at the code to get some understanding of GBVM and noticed eventActorShow was missing a label as opposed to eventActorHide. I just added it back.


chrismaltby:
Good spot, thanks!

PR #1009, opened 2022-01-19
Add vertical scrolling in last parallax viewport

um3k:
Makes it possible to have vertical scrolling in the last parallax viewport, allowing the use of a fixed position parallax viewport heads up display with more freedom of movement. The scene height must be no more than 256 pixels (32 tiles).

Enables vertical scrolling in the last (either second or third) parallax viewport. While this may act strangely with standard parallax backgrounds, it will enable top-of-the-screen HUD elements in a fixed position parallax viewport. 

While a scene taller than 144px will have vertical scrolling, it will erroneously overwrite tiles in the fixed viewport, thereby corrupting the HUD. Also, in order to have vertical scrolling, one must enable three parallax layers, which means the HUD can be no smaller than 2 tiles, made up of two layers.

The fixed viewport is not updated when scrolling, so it can be used for a tile-based HUD. Vertical scrolling is now enabled on the bottommost parallax viewport whether 2 or 3 are enabled.

I don't believe so.

* **Other information**:
This is only really helpful for a very specific use case, but I don't think it harms normal functionality, so I don't see any reason not to add it.


untoxa:
does this works correctly when you move camera instantly? i mean this part:
```c
            // If row differs by more than 1 render entire screen
            scroll_render_rows(draw_scroll_x, draw_scroll_y, ((scene_LCD_type == LCD_parallax) ? port->start_tile : -SCREEN_PAD_TOP), SCREEN_TILE_REFRES_H);
```

untoxa:
also pr changes to [GBVM](https://github.com/chrismaltby/gbvm) because we develop core there and only then merge it into the main GBStudio repo.

um3k:
> does this works correctly when you move camera instantly? i mean this part:
> 
> ```c
>             // If row differs by more than 1 render entire screen
>             scroll_render_rows(draw_scroll_x, draw_scroll_y, ((scene_LCD_type == LCD_parallax) ? port->start_tile : -SCREEN_PAD_TOP), SCREEN_TILE_REFRES_H);
> ```

It seems to work correctly in my testing.

> also pr changes to [GBVM](https://github.com/chrismaltby/gbvm) because we develop core there and only then merge it into the main GBStudio repo.

Ok, I'll close this one and move it there.

untoxa:
no. please split into two, because there is no `compileData2.ts` in that repo

um3k:
Oh ok, I'll do that.

um3k:
Apologies for my trainwreck of commits, I'm still pretty dumb when it comes to github

chrismaltby:
Thanks for this @um3k 

For a cleaner git history I just added this change as a single commit here https://github.com/chrismaltby/gb-studio/commit/b8f7e8811e5316eb0450838352eb7f77982fb28a I had to change it a little though as your version would prevent the final layer from being parallax in the 1 layer and 2 layer modes. I've made it so the final layer doesn't fill the screen anymore (allowing vertical scroll) if the Speed=1

um3k:
Thanks! I definitely made a mess of those commits 😆 

For future reference, is there a straightforward way to work on changes that need to be coordinated between the gb-studio and gbvm repositories?

Issue #1006, opened 2022-01-18
Converted .uge tracks return error message, cannot compile or edit


Anima-SodoDev:
**Describe the bug**
I have converted my .mod tracks to .uge using hUGETracker (importing the original file and saving as the new format), but they seem to return an error message when trying to compile the game or modify the tracks in the embedded tracker, `Cannot read property '0' of undefined`.

**To Reproduce**
1. Download [this track](https://cdn.discordapp.com/attachments/925126645491302563/933107821522874398/An_Age_Long_Gone_Short.uge).
2. Make a play music event for it.
3. Try to edit or compile.

**Expected behavior**
I should be able to edit the tracks or at least compile the game.

**Screenshots**
![imagen](https://user-images.githubusercontent.com/34885906/149939974-cf71a544-7030-49cf-8d28-7af910744fca.png)
![imagen](https://user-images.githubusercontent.com/34885906/149940166-b3c7ede9-e656-4496-b7ca-c623265051f3.png)

**Platform (please complete the following information):**
 - OS: Windows 11 Pro
 - App version: 3.0.2, built from GitHub, so I have the latest commits.

pau-tomas:
Sorry but I can't reproduce this issue on Mac, the song loads in the tracker and compiles correctly.

Anima-SodoDev:
I noticed that track specifically does work after making this issue, I updated the link to a track that I know doesn't work. My bad.

pau-tomas:
I think I found the issue but it might take a bit to solve in a proper way. Thanks!

PR #1011, opened 2022-01-20
Use pattern key when loading/saving uge files

pau-tomas:

Bug fix

Version 5 of the uge file format uses unique keys to reference each pattern, those keys aren't consecutive like in previous versions. 

GB Studio is skipping those patterns and using a consecutive id instead meaning that some uge files created with hUGETracker can't be opened with GB Studio because the patterns referenced in the sequence order *by key* have a different id when generated in GBS. #1006  is an example of this.

To make it worst. uge songs saved with GBS have a, related but different bug, where the pattern key is always set to `0`, `1, `2` and `3`... meaning that those files can't be open with hUGETracker because they contain duplicate pattern keys.

In short... a mess. Sorry.

On load. If the file is version 5 the pattern key is read from the file and used as the pattern id. 

On save. A unique id is assigned to the pattern key of each pattern [read other information below to some more context about why not using the original pattern keys].

The fix for on load, breaks compatibility with the uge songs saved with the old (broken) save method. So, on load, detect if the song has duplicate patterns and is version 5, and in that case, default to the previous method (ignoring pattern key) so the song can be loaded properly. The song is fixed the moment it's saved again.

As far as I can tell, no. 

* **Other information**:

So... why are the uge songs not saved using the pattern keys provided in the original uge file? 

All is due to a design decision from the early days on the tracker, that could be hard to revert as most of the UI is based around it.

In hUGETracker (an in the .uge format) a pattern is a set of 64 notes that can be used in any of the 4 channels. The sequence is made by orders that are a combination of one pattern for each channel. 

So an example of a sequence in uge is as follows

```
Order 0 [0, 1, 2, 3]
Order 1 [4, 5, 6, 7]
Order 2 [0, 1, 2, 3]
Order 3 [18, 5, 22, 3]
```

And the uge format stores all the patterns from 0 to 22, even if they're not used in the song orders.

In GB Studio a pattern is a combination of 4 set of 64 notes (one for each channel). And the song sequence is a list of Patterns. At load time we check for unique combinations of hUGETracker-style patterns and assign them to a GBS-style pattern.

So the example above becomes

```
Pattern 0 = [0, 1, 2, 3]
Pattern 1 = [4, 5, 6, 7]
Pattern 2 = [18, 5, 22, 3]
```

and the sequence

```
[Pattern 0, Pattern 1, Pattern 0, Pattern 3]
```

and any unused hUGETracker-style pattern is discarded (because it doesn't belong to any of the GBS-style patterns).

On save, we loop thru the sequence, and separate each GBS-style pattern into 4 hUGETracker-style patterns and generate the list of orders based on those newly generated patterns and their ids (which aren't related at all with the original ones they have when created in hUGETracker).

The result is a trimmed down .uge file that sounds the same and has the same structure for patterns, orders and sequence but they have different ids and only contains the used patterns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix rendering of garbage when no scene has loaded yet [untoxa](https://github.com/untoxa)
 - Fix overlay hide [untoxa](https://github.com/untoxa)
 - Fix issue where walking events was incorrectly replacing actorIds with $self$
+- Fix issue with saving/loading patterns from UGE files [@pau-tomas](https://github.com/pau-tomas)
 
 ## [3.0.2]
 
--- a/src/lib/helpers/uge/ugeHelper.ts
+++ b/src/lib/helpers/uge/ugeHelper.ts
@@ -22,7 +22,7 @@ export const loadUGESong = (data: ArrayBuffer): Song | null => {
   // TODO: Use `DataView` object instead of loads of Uint32Arrays
   let offset = 0;
   const version = new Uint32Array(data.slice(offset, offset + 4))[0];
-  console.log(`uge version: ${version}`);
+  console.log(`! uge version: ${version}`);
   if (version < 0 || version > 5) {
     throw new Error(`UGE version ${version} is not supported by GB Studio`);
   }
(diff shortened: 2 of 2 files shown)


chrismaltby:
Thanks @pau-tomas Can't say I fully understand the change without delving into the UGE file format but I managed to reproduce the bug and this PR fixed it! 😅 How stable is the UGE v5 format? Might be time to but some test cases in for these helpers

pau-tomas:
Yeah, it's a bit cumbersome to explain. v5 is the current version used by hUGETracker so pretty stable I'd say (I can confirm with @superdisk). Definitely some tests will be helpful.

PR #1013, opened 2022-01-21
Improve palette select for sprites by using the "real name"

ReyAnthony:
This commit adds a convenience feature to the sprite palette select for sprites by using the "real name" of the palette instead of just "Palette N"

When listing the available palettes for coloring a sprite, it shows a list of "palette N" 

The list now uses the name of the palette (eg : default sprites) 

* **Breaking changes ?**
No, afaik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added ability to see where automatic Fade In event will appear in Scene "On Init" script with option to disable or change speed
 - Added missing label in Actor Show event [@ReyAnthony](https://github.com/ReyAnthony)
 - Add vertical scrolling in last parallax viewport is Speed=1 [@um3k](https://github.com/um3k)
+- Add palette name to tile palette select based on current preview scene [@ReyAnthony](https://github.com/ReyAnthony)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ReyAnthony:
@chrismaltby Maybe it would be better to also show the number of the palette too, anyway, I believe it's an improvement from the current behavior because I had to do back and forths with the settings pane.

pau-tomas:
The dropdown in the sprite editor allows to pick a palette index not a specific palette for that sprite. That palette index could be different per scene, since each scene can have its own set of palettes.

To put it another way, if there's two scenes with different palettes in in index 1, a sprite with palette 1 will be using different palettes when it appears on each scene.

ReyAnthony:
@pau-tomas Oh, I indeed did not knew of this behavior. (Been using the tool for 3 days tops 😛)
I've checked what you suggested and when previewing as scene is used, the right name is used as expected with my code !

An image is worth more than words : 

default : 
image

scene preview :
image

As for the indices, would this be satisfactory ? 
image

ReyAnthony:
reyanthony/gb-studio@7e1ccfe6

image

ReyAnthony:
reyanthony/gb-studio@3c7852a 
(reyanthony/gb-studio@7e1cfe6 was a mistype)

image

chrismaltby:
That's a nice change, thanks @ReyAnthony

PR #1017, opened 2022-01-22
Fallback to current song waveforms in song preview

pau-tomas:

Bug fix

Wave preview fails because no waveforms are passed to the function

If no waveform are passed to the function use the current song waveforms. The parameter waveforms are only used for previewing from the instrument editor as they might be changed there.

No

* **Other information**:


chrismaltby:
Thanks @pau-tomas that's much better.

Small issue I've noticed is if you edit the waveform, the "Test Instrument" button immediately reflects the change but if you then click to add a note it seems like it uses the original waveform. If you play the song, then add a new note though it then correctly uses the new waveform.

PR #1025, opened 2022-01-25
Update pl.json

ReptiIe:
Update for the file, so again it will be up to date.
Added the mod. from en.json (a2ae888) 2022-01-24

Update of translation

Missing translation.

Update of course of the missing part.

No.

* **Other information**:
-

### DIFF
--- a/src/lang/pl.json
+++ b/src/lang/pl.json
@@ -2,7 +2,7 @@
   "RTL": false,
 
   "GBSTUDIO_DESCRIPTION": "Visual retro game maker\nPolish translation by: Reptile (reptile@o2.pl)",
-  "GBSTUDIO_COPYRIGHT": "Distributed under MIT license.\n Translation: 981220124",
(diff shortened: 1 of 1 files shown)


ReptiIe:
@chrismaltby another update.
By the way I have a question if you do not mind. When/How the translations are added to software?
I believe it is not like the GBStudio will "update" it self every day, is that correct?
It is like -this translation- that is right now, will be available with next GBStudio update (3.0.3)
Just simply wondering.

chrismaltby:
Thanks @ReptiIe

To answer your question, the translation files are stored within the application package (compressed in the file `resources/app.asar`), and the application doesn't auto update (though it will notify you if a newer version is available) so changes you make here will only appear in official releases the next time I make a snapshot and upload to Itch.io.

I'm trying to keep pushing regular releases at the moment so it shouldn't be too long before this change will be in the live version though.

PR #1034, opened 2022-01-28
Fix ActorMoveRelative and ActorSetPositionRelative moving in the wrong direction near top or left side of screen

um3k:
Bugfix

ActorMoveRelative and ActorSetPositionRelative calculate a negative destination when they would move left or up past 0. This would overflow and become a large positive number, resulting in the actor moving right or down, or simply disappearing from the screen.

The calculation results are now clamped to 0, so the actor will stop at the edge of the screen as expected.

I don't believe so.


um3k:
Are there issues with this that are preventing it from being merged? Please let me know so I can fix them.

chrismaltby:
Thanks @um3k this is all good and fixes the issue! Just been a bit busy recently and not had much time for testing PRs

Issue #1019, opened 2022-01-22
Ability to interrupt a moveto on actors


blurymind:
**Is your feature request related to a problem? Please describe.**
Lets say I have an AI with a patrooling behavior

The AI is using a switch to jump between states. The patrooling state is using a chain of movetos to define the route
Lets call that state == 1

Now lets say we have the player shoot at the ai, while the ai is patrooling and instead of hiding the enemy actor, we want to put the enemy actor into a temporary stun state.
So when the player bullet hits the enemy actor, we set the ai state to == 2 (stunned). When set to stunned, we expect the enemy to stop moving instantly upon being hit by the bullet. Since the state is now not one, we expect the movetos to instantly stop.

So what appears to happen however is that the ai switches to another state (2), but the moveto continues to execute - after being hit by a bullet the enemy continues to walk, until executing all the remaining movetos - it walks for a while on its route and then finally stops.

**Describe the solution you'd like**
This looks like a bug to me. If the wrapping if statement stops being true, the movetos inside it should instantly be interupted, but they arent.
If there is a plausible technical or design reason for that, then I would like to request an action, which clears the movetos and instantly stops the ais movetos from executing.

**Describe alternatives you've considered**
I "wait" action, when reached should also be capable of stopping/interrupting the movetos- but it seems completely helpless too. 

**Additional context**
Please make a more capable example of enemy ai as part of the project. Gb-studio appears to have a serious flaw preventing it from being able to do a capable state machine using the move actions predictably

maxoakland:
Maybe there could be a “Stop Move” and “Stop Wait” event like the one to stop an actor’s On Update script

blurymind:
the problem with the action to stop an actor’s On Update script is that we dont have one to start it again.

I tried to use that and it did stop the movetos, but I couldnt get the ai to get out of the stun status. We do not have an action to start the on update script after it was stopped

maxoakland:
> the problem with the action to stop an actor’s On Update script is that we dont have one to start it again.
> 
> I tried to use that and it did stop the movetos, but I couldnt get the ai to get out of the stun status. We do not have an action to start the on update script after it was stopped
> 
> it could potentially be also used for pausing the game?

I’m not saying you should use Stop On Update, I’m saying a similar script could be a way to implement this request

blurymind:
yes of course, I like the idea of a "stop move to actions" script, just not sure how to implement it

maxoakland:
[...]

PR #1036, opened 2022-01-31
Actor stop movement

um3k:

Feature

Currently the only way to stop a running Move To event is to terminate the script that is running it, however, this means the script cannot be easily restarted.

A new event name Actor Interrupt Movement, which causes the actor's current move event to stop as soon as the actor reaches an integer tile coordinate.

I do not believe so.

* **Other information**:
Inspired by this issue: https://github.com/chrismaltby/gb-studio/issues/1019
GBVM counterpart: https://github.com/chrismaltby/gbvm/pull/138


untoxa:
i renamed the event to VM_ACTOR_MOVE_CANCEL which makes more sense. please update the codegen part.

um3k:
Codegen is updated!

As far as the checkboxes up top, are they strictly necessary? I have no idea how to write or run tests, but I would be happy to implement them if someone can point me toward some learning resources. And should I go ahead and write an update for the docs?

pau-tomas:
Works well on my testing. Pretty useful in combination with the new Hide and Deactivate events. As it'll allow to reuse enemies once they're "dead".

For tests, check any `event*.test.js` file, they give an input to the compile function and checks that the methods on scriptBuilder are called properly and with the right parameters. You can run tests with `yarn test`. Not all events have tests, so not 100% required :)

PR #1037, opened 2022-01-31
Experimental: Platformer actor collision

um3k:

Feature

Platform mode does not support solid collision between actors and the player

Actors have a 'solid' property, when it is enabled, the player will collide with them like a wall, and will move along with one that it is standing on.

I don't believe so.

* **Other information**:
Still need to investigate optimization. It works OK in color mode, but sprites get jittery in DMG mode. There are also issues with how the player moves with an actor it is standing on, and collision events currently do not trigger on solid actors.


darkhog:
Once you are done with this, could you please also do the same for the adventure mode? Code shouldn't be that much different tho in adventure the moving along with actor would have to go.

The reason for this is that in adventure mode currently there is no possibility of creating secret entrances that only become passable once certain conditions are met as map collision is static and you can pass through actors that would normally block players from entering before they met the necessary conditions.

um3k:
Yes, I intend to port the changes to that mode.

darkhog:
That's awesome. Thank you!

maxoakland:
This is so cool. I hope you can get this implemented because we’ve needed this feature for a loooong time

godai78:
Only now have I found it. This one sparks joy.

godai78:
Will it make it into 3.1?

maxoakland:
> Will it make it into 3.1?

Looks like there are still some conflicts that must be resolved

chrismaltby:
Hi @um3k Just to let you know, after a very long time I have finally been testing this and definitely interested in merging.

Not sure it's going to be 3.1 though, I'm starting to feature freeze on what makes it in to 3.1 and I'm still having a few issues with it.

- In `Platform` scenes if stood on a moving actor, that is solid in all directions, while it is moving upwards and you're either at the left edge jumping right, or the right edge jumping left you snap back to the side of the platform (think this is probably the horizontal collision detection kicking in when it shouldn't). From what I can tell as long as the vertical moving platform just has "Top" collisions turns on, or your fully solid actor is either stationary or only moving horizontally then it all works great.

https://user-images.githubusercontent.com/16776042/180873013-5aa43bb1-68c2-4a1b-bcef-786b9693b440.mp4

- Possibly related vertical moving actors set to fully solid in `Adventure` also do some odd snapping when colliding with top / bottom edges (left/right work correctly)

https://user-images.githubusercontent.com/16776042/180874469-f46fc157-b609-4904-9b9c-ee8c3bd0d226.mp4

- I think the quick check you do pre running the full collision check might be too small a range for some larger actors, if I'm understanding what you're doing correctly increasing to `+ 3` and `> 6` help a bit
```
             ((UBYTE)((BYTE)(PLAYER.pos.x >> 8) - (BYTE)(sol_actor->pos.x >> 8) + 3) > 6) ||
            ((UBYTE)((BYTE)(PLAYER.pos.y >> 8) - (BYTE)(sol_actor->pos.y >> 8) + 3) > 6)) {
```

By the way the persistent actors feature ("Keep Running While Offscreen" checkbox on "On Update") and recent optimisations to actors_update() and vm_actor_move_to() + some better offscreen detection we wrote recently were to prepare for merging this feature.

Thanks for putting this PR together, definitely is a great feature that people will make lots of use of!

um3k:
Hi Chris, thanks for checking in! I guess it's time to dust off the ol'
keyboard and find some time to reacquaint myself with the code.

I definitely experienced those collision bugs at some point during my
testing, it's possible I have more recent code in my local directory that
fixes them. I can't remember exactly where I left off in my endless
tinkering. So I'll look into that soon.

I agree, those numbers for the quick check are probably better. I was
hesitant to set them too high and potentially defeat the purpose of the
check. I wonder if it would be feasible to set them at compile time based
on the largest sprite in the project? Might need to look into that.

I'm humbled that that work was done in anticipation of this, now I'd better
pull it together and get it done!

jessemillar:
I fixed the two merge conflicts locally and built an adventure scene. It works pretty well for 16x16 actors (I didn't try any larger actors). The one "bug" I found is that solid actors pushed with the "Push Actor Away From Player" event do not always collide with other actors that are set to solid on all sides. I was able to consistently push a solid box to the right and have it phase through another solid box. A solid box pushed vertically DOES stop properly when encountering another solid box. 
image

godai78:
That's impressive work, looks like it's closer than ever before :)

jessemillar:
I just made my own PR branching off the content of this PR. Attempted to fix the merge conflicts and issues found by me and Chris mentioned in the comments above. https://github.com/chrismaltby/gb-studio/pull/1232

PR #1059, opened 2022-02-19
Update pl.json

ReptiIe:
Another fresh update for pl.json file // added missing lines, so it has again the same number as en.json (998)

Update

Missing translation for actor move cancel

Added translation for it.

No

* **Other information**:
I think I will drop a comment, since there are few things.

### DIFF
--- a/src/lang/pl.json
+++ b/src/lang/pl.json
@@ -2,7 +2,7 @@
   "RTL": false,
 
   "GBSTUDIO_DESCRIPTION": "Visual retro game maker\nPolish translation by: Reptile (reptile@o2.pl)",
-  "GBSTUDIO_COPYRIGHT": "Distributed under MIT license.\n Translation: 996220213",
+  "GBSTUDIO_COPYRIGHT": "Distributed under MIT license.\n Translation: 998220219",
  
  "// 1": "UI -------------------------------------------------------",
   "ACTOR": "Aktor",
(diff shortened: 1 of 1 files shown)


ReptiIe:
@chrismaltby 
There will be a bigger update of PL.JSON file soon, few label (translations) I have to correct them, so they will be more accurate, and be the same as DOCS for PL version (yeah, I am slowly working on the DOCS as well - so they will get an update as well).

But I see DOCS for EN version are missing some options (events commands like: group events, link: connect/host/transfer/close,  GBVM script, set music, now actor move cancel)
who I suppose to contact, so the DOCS for EN will be complete, will get complete? With who I can work, with the docs, so the work will be fluent like this that we have right now (so I do not need to wait too long for updates and rewrite the whole DOCS again and again :D )

Other question: 
I notice there was some PR, or something (some project) for the GB Studio to run on Raspberry Pi.
Just would like to know, is this still on?
I mean there will be some works, that Raspberry will receive it's own GBS build?
(maybe even in the repository that people could write: "sudo apt get instal gbstudio" and they have GBStudio already?) 
- not to jump into any high hopes, or making a big news out of it,
Just would like to know if there will be any progress in the future, is it possible for RPI to receive it's version;
and "maybe" when - like at the end of year, or maybe within 3.0.10 or something like that.

Issue #1058, opened 2022-02-18
Condition to check distance from player


blurymind:
**Is your feature request related to a problem? Please describe.**
Right now every time we want to make an ai, we have to nest several ifs, just to check if a distance threshold from the player has been met. Example:
https://www.youtube.com/watch?v=5e_n8Znid8w&t=320s

**Describe the solution you'd like**
Instead of all these ifs making the code hard to maintain, I came up with a simple expression for that example
abs(playerX - catX) + abs(playerY - catY)  - attackDistance < 0 // attack player when met for when player is aligned  with cat on y

What if instead  of formulas or nested ifs, gb-studio offered a built in solution that just gives you two actors to select (player/cat),  some extra parameters maybe and then you nest under it what happen - or just give it a script to select as a parameter

**Describe alternatives you've considered**
I havent tested this expression myself, just giving it as example for a way to do it cheaply perhaps. Maybe there is more to this - especially if the game could also be a platformer

**Additional context**
The goal here really is to get rid of more common boilerplate - for building ais and other game mechanics. Very very often we need to check for distance from  player/npc. Having an easy shortcut  method will really improve the ease of use of the engine imo

juliusl:
Not sure if this is helpful but I wrote a range check script,

![Range check](https://cdn.discordapp.com/attachments/692239309926957116/943431324105076776/unknown.png)

Basically, I just take the distance formula and instead of taking the sqrt I just square the distance I want to check. Example, usage ends up looking like this: 

![image](https://user-images.githubusercontent.com/5741075/154822893-eb74e1a9-c21e-4404-8dd0-ee51bb98c739.png)
![image](https://user-images.githubusercontent.com/5741075/154822907-30194895-34e3-42f7-a828-5cbfa6e0e18f.png)
![image](https://user-images.githubusercontent.com/5741075/154822940-6c3b8aa0-1f08-4b44-ab2b-2534469b1aa4.png)
![image](https://user-images.githubusercontent.com/5741075/154822946-63111252-7c47-4c6e-bdf0-253d5d6b1122.png)

I tinkered with it a bit more and was able to bring it down to using only 4 variables however, I didn't save that version as the reason I made it was to show someone an example and I wanted show all of the math step by step. But I think using the gbvm script event, I could make it more user friendly (and faster), because all I really need is a reference to the other actor, the range that needs to be checked, and the output variable to set.

juliusl:
I added a PR that adds an event for this, #1060

PR #606, opened 2020-10-11
Add event "If Actor In Boundary"

patrickmollohan:
* **Please check if the PR fulfills these requirements**
- [X] The commit message follows our [guidelines](/chrismaltby/gb-studio/blob/develop/.github/COMMIT_MESSAGE_GUIDELINES.md)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Adds the event "If Actor In Boundary". This event checks whether the specified actor is in the boundary. The boundary is rectangular, from top left (X1, Y1) to bottom right (X2, Y2).
If Actor In Boundary

* **What is the current behavior?** (You can also link to an open issue here)
You cannot check whether an actor is in a boundary without using several events (slower) and quite a few variables. You can also attempt to fake this functionality using triggers, but that has its own drawbacks.

* **What is the new behavior (if this is a feature change)?**
You can now easily check if an actor is in a specified boundary.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
Not that I'm aware of.

* **Other information**:
As with my previous pull request, not sure if I'm adding events correctly, but it works. It would be nice to add some visual indication in the GUI on the Game World page for where the boundary is, much like what can be seen with triggers or events like "Actor: Move To".

PR #1060, opened 2022-02-20
Adding a conditional actor to actor range check event

juliusl:
Signed-off-by: Julius Liu <juliusl@microsoft.com>

- [X] Tests for the changes have been added (for bug fixes / features)

Feature

N/A

N/A

N/A

* **Other information**:
Addresses #1058

Screeshots:

# Event Configuration
image

# Event Preview
distance_preview_2

# Menu
image

# In Action
tinker


juliusl:
Some things I was thinking about after working on this, and I'd like to open it up for discussion:

1) I'm not sure if pixel units is accurate or if it should be tile units
2) I think the calculations I'm using could be optimized but I opted for correctness 
3) If this is used in OnUpdate, it can slow perf down, but adding a wait immediately after the event mitigates that, I'm wondering a polling rate field should be added to the event?

blurymind:
i think tile units would be better
optimized is probably better than super accurate
yes some sort of polling rate would also be helpful

It not like we can have tons of actors at once in gb-studio. It already has a limit on that number of actors. Thing is when that limit is almost reached, using this doesnt affect performance much

juliusl:
>  I dont think it needs to be super accurate - just work with minimal input.
(instead of asking you to store player/actor x and y, just ask you to select actor a and select actor b - extract their x/y under the hood and not ask the user to do that) The less manual steps it takes to set up, the better!
optimized is probably better than super accurate

Yep, since I'm adding a new event, I was able to flesh out the UX alot more, this is what it looks like now: 

image

juliusl:
@blurymind 
> yes some sort of polling rate would also be helpful
> 
> 
> 
> Its not like we can have tons of actors at once in gb-studio. It already has a limit on that number of actors. Thing is when that limit is almost reached, using this doesnt affect performance much - and of course the minimum input required to work. The less mandatory input/steps to set up- the less noise in the gui to read the logic and the faster we can set the ai.

Yeah so since I know gbstudio is geared more towards artists, I figured polling rate might be too technical of something to consider up front, and also the interval you want your ai to check could also be considered a narrative characteristic. 

What do you think if we called it "Detection Speed", with options like: (slower, slow, normal, fast, faster, custom)? That way it's consistent with other event options.

blurymind:
Maybe Detection rate? Detect every x ms?

Detection speed works too :) Great job btw

blurymind:
should it have a ><= parameter?

What if we want if distance is greater or equal?

pau-tomas:
Nice work! It reminds me of #606 

A couple of comments:
- Since tiles is the default unit of measure across the app you can leave the field as `Distance` no need to mention `tile units`.
- Would be great to be able to preview the range in the scene (similar to how `Move To` shows a box of the destination position (check `eventsHelpers.ts`).
- Add an option to use a variable for Distance.
- Detection speed is probably not needed as it will overcomplicate the event. A lot of existing events used inside `On Update` will have a hit in performance if there's no wait or vm_idle right after them.

juliusl:
Just letting everyone know that this is still on my todo list, I've just been on vacation will get to back to it this week ~ 

> @pau-tomas Would be great to be able to preview the range in the scene (similar to how Move To shows a box of the destination position (check eventsHelpers.ts). 

Preview would be awesome, I'll take a look at that. I think I should be able to add a variable for distance. 

> @blurymind should it have a ><= parameter?
> 
> What if we want if distance is greater or equal?

That should be doable, I'll try it out and show what it looks like with that option

juliusl:
distance_preview

@pau-tomas  :D

juliusl:
distance_preview_2

@pau-tomas @blurymind  Ok I think I've about covered all of the comments so far, take a look.

juliusl:
@chrismaltby Hey there, I was wondering where do I update documents?

blurymind:
very nice :) I do hope this gets merged

pau-tomas:
Hey @juliusl, thanks for the updates! I tested it again and found a crash when using the event in a non-actor script and trying to preview the distance in the scene. The crash was due to the `$self$` actor being `undefined` in that case. I fixed that bug and took a bit of a different approach to using `actorsLookup` here https://github.com/pau-tomas/gb-studio/commits/fix/pr_1060 if you like that approach, please, feel free to use it.

In that branch also I ran `yarn prettier` on the changed files to ensure they're consistent with (most) of the codebase. (note for myself: document that in the how to contribute list)

juliusl:
@pau-tomas Looks good to me, I'll update my branch

juliusl:
@pau-tomas gentle ping, is there anything else I can do for this PR? :D

chrismaltby:
Hi @juliusl I finally got around to reviewing this, awesome work! Really like the preview :-) 

I made a few minor tweaks to the UI and it seems like you'd missed handling the case when distance was set to a variable rather than a number so I've implemented that, other than that it's all good, happy to merge!

juliusl:
@chrismaltby woohoo! thanks for the fixes and merge 🥳🥳

PR #1065, opened 2022-02-21
Update pl.json

ReptiIe:
Update pl.json; missing fields for SE from 21.02.2022 update.

Update

Missing fields for some translation

Update missing fields - so it will be the same as current en.json (1005 lines)

no

* **Other information**:
Again the comment from previous PR 😄


ReptiIe:
- the same comment as the previous PR

@chrismaltby
There will be a bigger update of PL.JSON file soon, few label (translations) I have to correct them, so they will be more accurate, and be the same as DOCS for PL version (yeah, I am slowly working on the DOCS as well - so they will get an update as well).

But I see DOCS for EN version are missing some options (events commands like: group events, link: connect/host/transfer/close, GBVM script, set music, now actor move cancel)
who I suppose to contact, so the DOCS for EN will be complete, will get complete? With who I can work, with the docs, so the work will be fluent like this that we have right now (so I do not need to wait too long for updates and rewrite the whole DOCS again and again :D )

Other question:
I notice there was some PR, or something (some project) for the GB Studio to run on Raspberry Pi.
Just would like to know, is this still on?
I mean there will be some works, that Raspberry will receive it's own GBS build?
(maybe even in the repository that people could write: "sudo apt get instal gbstudio" and they have GBStudio already?)

not to jump into any high hopes, or making a big news out of it,
Just would like to know if there will be any progress in the future, is it possible for RPI to receive it's version;
and "maybe" when - like at the end of year, or maybe within 3.0.10 or something like that.

chrismaltby:
Thanks @ReptiIe 

On the Raspberry PI build, I bought a PI4 a while back specifically for this but was unable to get gbstudio running on it correctly at the time, if I remember rightly it didn't seem to have enough RAM to build. We'd also need to get an ARM version of [GBDK-2020](https://github.com/gbdk-2020/gbdk-2020) which is not an official build at least yet (though it may work, not sure). I'd like to see a PI version, just not got time to commit to doing this myself just yet.

On the docs I'll be coming back to these once again once the next few smaller releases are finished to start documenting the things that were added recently, my day job is quite busy at the moment and only get a small amount of time to work on this.

PR #1070, opened 2022-02-23
Add pixels option to actor/camera position selection

Y0UR-U5ERNAME:

Feature

The events where it is possible to choose a actor/camera position (Store Actor Position In Variables, Actor Move To, Actor Move To Variables, Actor Move Relative, Actor Set Position, Actor Set Position To Variables, Actor Set Relative Position, If Actor At Position, Camera Move To) only allow for positions in terms of tiles and not pixels. Additionally, it is only possible to get the tile position X and Y as a property.

- Each of the events listed above has an additional dropdown to choose whether to use tiles or pixels.
- It is possible to get an actor's pixel position X and Y through properties.
- The camera and actor position previews in the World are updated to reflect the pixel position if needed.

No

* **Other information:**

This does not affect overlay position selection.


chrismaltby:
Thanks @Y0UR-U5ERNAME , this is really nice. Sorry it's taken me so long to review though!

I'm working through merging this now, I'm just going to simplify the UI a little. Rather than adding the extra dropdown, which takes up quite a lot of space, I've made it so that the Number and Variables inputs can have a units selector which appears within the element like this:

Screenshot 2022-07-16 at 19 29 23

and clicking on the units opens a menu to switch them

Screenshot 2022-07-16 at 19 28 59

other than that I'm mostly able to use what you've written :-)

PR #1071, opened 2022-02-23
Fix: undo on music tracker also undoing global project history

pau-tomas:

bugfix

If undoing in the tracker or piano roll editor, undo will also be triggered for the general project history meaning that actions done in other editors (World, Sprite, etc) will also be undone. Same applies to redo.

Added a guard to the undo and redo actions to only trigger for the project history if there's no past or future history on the tracker document.

No

* **Other information**:

@chrismaltby not sure if this is the best solution for the issue. You might have a different/more elegant one, maybe.


chrismaltby:
Thanks @pau-tomas I tried to figure out a more elegant solution but I think that would require combining all the `project` reducers and the `trackerDocument` reducer undo a single `undoable` which would require probably quite a lot of refactoring and probably introduce some tricky bugs. This fix works well for me so lets use it :-)

PR #1073, opened 2022-02-26
Add Emulicious debug support

RichardULZ:

Feature

bgb profiling doesn't work, can not debug with c easily without setting up gbdk and external build tools,

Changes the bgb profile button to instead export c debugging files.
Profile button exposed to regular builds.

Requires Emulicious, VSCode, and Emulicious debugger extension
https://github.com/Calindro/emulicious-debugger 

Regular builds are not changed.

* **Other information**:
Includes an example vscode\launch.json to point emulicious at the rom when you press F5 in vscode.
The folder should be named .vscode, but is ignored by github, have to rename it. Having this json file written could work?

A Readme is included with win64 build tools, only tested on windows 64bit, copy and amend this for macos if it works.
Debug compiler flags should work everywhere, unsure on debugging link for emulicious on other systems.

Extreme thanks to @Calindro for creating Emulicious and putting up with a long delay in publishing this simple pr. 

This should work as is, but any testing, improvements to the workflow, or suggestions on how to set flags would be appreciated.

### DIFF
--- a/buildTools/win32-x64/vscode/Readme Emulicious debugging.txt
+++ b/buildTools/win32-x64/vscode/Readme Emulicious debugging.txt
@@ -1,0 +1,17 @@
+"Enable Emulicious C debugging" will generate debug files for your rom, which you can use to step through code.
+
+Requires emulicious https://emulicious.net/
+vscode https://code.visualstudio.com/
+and the Emulicious Debugger extension https://marketplace.visualstudio.com/items?itemName=emulicious.emulicious-debugger  
+
+in VS Code, Configure emulicious debugger's "extension settings" and set the path, eg "C:\Emulicious\"
+
+Set your temporary folder in gbstudio to a convenient location in the prefrences (or here, since you found this file)
+Rename this folder from vscode to .vscode (you may need to close this file)
+Open this gbstudio temporary folder in vscode, 
+
+Generate a rom via the build or run buttons in gbstudio with "Enable Emulicious C debugging" checked
+
+you can press F5 to launch emulicious with your rom and begin debugging
+Or see other methods of running with emulicious in the extension's decription.
+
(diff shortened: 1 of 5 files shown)


Calindro:
It's all platform-independent so it should work on all platforms as-is. :)

I'm also watching this PR so I can answer questions or help with issues that may arise.

chrismaltby:
Thanks @RichardULZ looks good, made a few small changes since it seems like you had a bug preventing the flags being passed to the linker correctly. Also changed the .gitignore so we can just add the .vscode folder without requiring a manual rename.

Only problem now is does this actually work for you? Even before I changed anything when I build with this setting enabled I get an overflow in Bank 2, think we might need to move some things around in [GBVM]( https://github.com/chrismaltby/gbvm)  to free up some space in the core banks before this can work.

Screenshot 2022-03-06 at 11 54 33

Calindro:
Thanks for looking into this PR. :)

> Only problem now is does this actually work for you? Even before I changed anything when I build with this setting enabled I get an overflow in Bank 2, think we might need to move some things around in [GBVM](https://github.com/chrismaltby/gbvm) to free up some space in the core banks before this can work.

I did see it working for him. 

By `this setting` do you mean the debug flags (`-Wf--debug`and  `-Wl-y`) to lcc? Afaik, they should not affect the rom generation. They should only cause additional files (containing debug symbols) to be generated.

Calindro:
Are there any news on this?

Calindro:
@chrismaltby thank you! :)

chrismaltby:
Finally got this PR to work :-) looks like the CFLAGS change wasn't being applied I think an old/unused function for generating the flags had been updated rather than a newer function I'd switched to (I need to double check it's not used anymore and remove) 

From some quick testing the VSCode extension is stepping through the C source as expected!!

Calindro:
Awesome! Also thanks a lot to @RichardULZ for implementing it! <3 :)

PR #1092, opened 2022-03-30
Rename constants & variables for consistency

patrickmollohan:

Several gametype constants and variables were named inconsistently, making the code more difficult to read through or modify. I have renamed these variables and constants to be consistent with the file names of the states (i.e., "shooter" changed to "shmup" to be consistent with "appData/src/gb/src/states/shmup.c").

Several variables and constants are inconsistently named, for instance, "pl_", "plat_", "platform_", "platformer_".

The variables and constants are now consistent with the file name associated with the gametype ("platform_").

None that I am aware of.

* **Other information**:

### DIFF
--- a/appData/src/gb/engine.json
+++ b/appData/src/gb/engine.json
@@ -4,7 +4,7 @@
     {
       "key": "INPUT_PLATFORM_JUMP",
       "label": "FIELD_INPUT_JUMP",
-      "group": "GAMETYPE_PLATFORMER",
+      "group": "GAMETYPE_PLATFORM",
       "type": "select",
       "options": [
         ["INPUT_A", "A"],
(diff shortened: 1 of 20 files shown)


DeerTears:
Won't this break a lot of plugins or engine mods by changing many of these property names? I agree the consistency is for the best, but I wonder if there's a way to keep some parity with popular gamemodes like platformer and 2d top down.

chrismaltby:
Hi @patrickmollohan While I'm not opposed to this PR I think it will be a bit tricky to change without breaking a lot of projects. At the very least we'd need to add a data migration so that projects made with the previous variable names would have their `.gbsproj` files updated to use the new names otherwise any uses of the "Engine Field" events would be broken and any modified values in Settings would have reverted to their defaults.

I'll have a think about this, would like the names to be more consistent just don't want to break anything

patrickmollohan:
Very good points. I think it's best for now to close this PR and revisit the idea later on, as this probably breaks more than it fixes.

PR #1110, opened 2022-04-30
Projectiles from 224 to 360 degrees now look up

john-lay:
Bug fix to the `projectiles.c` file. The previous logic missed a scenario where projectiles greater than 224 degrees fell into the `(angle > 96)` condition.

Tweak the projectile logic when setting the correct frame based on angle.

Projectiles between 224 and 360 degrees point down.

Projectiles between 224 and 360 degrees point up.

No

* **Other information**:


chrismaltby:
Ah of course, good spot! Thanks for fixing this

PR #1133, opened 2022-06-22
Add AppImage support for Linux

patrickmollohan:

Adds AppImage support when building GB Studio for Linux.

Currently, electron-forge builds .deb and .rpm packages for Linux, which is good for Debian and Red Hat-based distros of Linux, but not so much for other distros such as those based on Arch. While the .deb/.rpm packages can be converted or extracted to work on Arch, it isn't recommended to do so. Additionally, all Linux distros are subject to "dependency hell" due to Linux's insistence on using shared libraries.

Most, if not all, Linux distros are capable of running AppImages, and thus, it would be useful to a broader group of Linux users. Additionally, AppImage bundles all required dependencies into one convenient package, and thus shouldn't suffer from "dependency hell"-related issues in the future. It also isn't as controversial as other software packaging formats such as snap.

None that I am aware of.

* **Other information**:
[bTw I UsE ARcH](https://knowyourmeme.com/memes/btw-i-use-arch) ;)

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -15,6 +15,10 @@ module.exports = {
       platforms: ["darwin", "win32"],
     },
     {
+      name: "@reforged/maker-appimage",
+      config: {},
+    },
+    {
       name: "@electron-forge/maker-deb",
       config: {},
     },
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
   },
   "devDependencies": {
     "@electron-forge/cli": "6.0.0-beta.57",
+    "@reforged/maker-appimage": "^1.1.3",
     "@electron-forge/maker-deb": "6.0.0-beta.57",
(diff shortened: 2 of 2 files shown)


patrickmollohan:
I feel like the AppImage only route would be sufficient to cover Linux these days. In my testing on my Arch desktop and Ubuntu VM, I haven't noticed any issues, and I've been using that since a few months before submitting the PR. I've also noticed a trend with Linux apps in general that the .deb/.rpm versions are slowly being phased out in favour of AppImages/Flatpaks/snap packages, and while I'm sure you'll have a few naysayers, it's honestly better in the long run unless Linux gets away from shared dependencies.

pcavalcanti2005:
Agreed. I've actually made an AppImage very quickly from the rpm package on Fedora Silverblue and it runs the same.

PR #1136, opened 2022-06-28
Add "Start Actor's 'On Update' Script" event

patrickmollohan:

Adds the ability to start an actor's "On Update" script.

Although you can stop an actor's "On Update" script, you cannot (intuitively) restart it once stopped.

Provides a more intuitive way to restart an actor's "On Update" script.

None that I am aware of.

* **Other information**:
1. This can technically be done with two existing events, a "Deactivate Actor" event followed by an "Activate Actor" event. As a matter of fact, that is what this code does, hiding it as a single event. This may not be a necessity to add, but it does make it a lot more intuitive to restart an actor's "On Update" script than running the other two events.
2. This does not add or change anything in GBVM, but instead relies on existing GBVM commands. One downside is it toggles the actor's hide property as a result of the deactivate and activate functions, but I haven't noticed any flickering of the sprite so it should be fine. It may be a better idea to add a new GBVM command for the "script_execute" function, much like the stop "on update" has for "script_terminate". Wouldn't be difficult to implement this way, just wasn't sure if a slot should be allocated for this in the array of script commands or if it should be reserved for more useful commands. Let me know either way and I can get it done! For now, though, I went with the more conservative route and spared adding a new GBVM command.


maxoakland:
Great feature! Has been much needed

chrismaltby:
Very nice, I agree we should probably have a new command in GBVM for this but for now this works great. Was also expecting a flicker but not seen any issues yet.

patrickmollohan:
Awesome! Glad to hear it is working on your end as well. I will hopefully have a proper GBVM script for it in the coming days!

PR #1143, opened 2022-07-11
Pass values to custom scripts + Global variables in custom scripts

chrismaltby:

PR Merge + Features

Implements functionality from #938 and #818 which had conflicts due to previous changes to ScriptBuilder:
- Add ability to pass number values as parameters when calling scripts [@pau-tomas](https://github.com/pau-tomas)
- Add ability to access global variables from within scripts [@pau-tomas](https://github.com/pau-tomas)

Additionally makes the following changes:
- Add ability to set script parameters to be passed by reference or by value (previously was always by reference)
- List of "Variable Uses" now calculated in background thread to reduce pause while reading large projects
- Timer, input and music scripts set from within a custom script no longer allow the script parameters to be used as they were not working, global variables can now be used instead.

Adding support for global variables from with custom scripts required a migration to replace parameter variables "0" to "9" with "V0" to "V9" so that they can be distinguished from the globals, opening your project will perform this migration automatically but after saving your project will no longer be compatible with previous releases.


src/components/forms/ActorSelect.tsx
```diff
@@ -90,7 +91,7 @@ export const ActorSelect = ({
           };
         }),
       ]);
-    } else if (sceneActorIds) {
+    } else if (context === "entity" && sceneActorIds) {
```
pau-tomas, line 94: I guess not a big issue because it's already not working, but it's unfortunate it has to be this way as one of the main uses of custom scripts is to configure actor behavior on button press (mostly the player tho...). Prefabs will be a good replacement for this anyway,

pau-tomas (commented):
Great job! The migration seems to be working fine and my existing scripts run well. 

From the tests I ran I found a couple of issues so far:

- For parameters set to `By Reference` when passing more than one parameter as number, all the parameters take the value of the last number parameter. It works as expected if the parameters are set to `By Value`. Not sure what would be the expected behavior here as passing a number by reference isn't correct... maybe the input field on the call script shouldn't allow to pass only variables in that case?

- The `By Ref / By Val` dropdown resets to `By Ref` when there's changes in the script.

- SUPER NIT: The `Global` header doesn't appear in the variable dropdown when inside a timer/attach/music script inside a custom script.

Will run more tests (I want to try how recursive scripts behave :) )

chrismaltby:
> Great job! The migration seems to be working fine and my existing scripts run well.
> 
> From the tests I ran I found a couple of issues so far:
> 
> * For parameters set to `By Reference` when passing more than one parameter as number, all the parameters take the value of the last number parameter. It works as expected if the parameters are set to `By Value`. Not sure what would be the expected behavior here as passing a number by reference isn't correct... maybe the input field on the call script shouldn't allow to pass only variables in that case?
> * The `By Ref / By Val` dropdown resets to `By Ref` when there's changes in the script.
> * SUPER NIT: The `Global` header doesn't appear in the variable dropdown when inside a timer/attach/music script inside a custom script.
> 
> Will run more tests (I want to try how recursive scripts behave :) )

Thanks for looking into this @pau-tomas I've fixed all these issues now, from my tests it seems good to go :-) Thanks for all the initial work get these features in place too!

src/components/forms/ActorSelect.tsx
chrismaltby: Yeah it is a shame, I wonder if in the future maybe we can have the timer/input/music scripts have a small amount of memory for storing args, that would allow all the script parameters (both variables and actors) to be used as normal

Something like this
```
        VM_CONTEXT_PREPARE      1, ___bank_script_input_0, _script_input_0
        VM_PUSH_CONST           5 ; Actor Param
        VM_PUSH_CONST           VAR_HEALTH ; Variable Param
        VM_INPUT_ATTACH         128, ^/(1 | .OVERRIDE_DEFAULT)/, 2 ; <-- Take 2 params off stack
```
then within the input script itself we could read the values the same way we do in custom scripts today

PR #1148, opened 2022-07-18
[Tracker/Bugfix] Set order count on song preview

pau-tomas:

Bug fix

When adding a new pattern to a song it was not taken into account on the preview. So if the song had 2 patterns and a new one was added, the preview will only play the first 2, ignoring the 3rd one.

Pattern (Order) count is properly set on play. 

Nope

* **Other information**:

🤦  pretty silly issue 🤦


pau-tomas:
LOL I thought you went to sleep already, Chris!

Issue #1098, opened 2022-04-12
Set position to another actor's position in custom script


john-lay:
**Describe the bug**
`Set Actor Position` using the property of a second actor's position inside a **custom script** doesn't change the actor's position.

**To Reproduce**
1. Create a Custom Script
2. Add the `Set Actor Position` event (this will add an `Actor A` parameter to the script)
3. Change the type of position X to property
4. Set the X position to Actor B's X Position (this will add a `Actor B` parameter to the script)
5. Do the same for position Y
6. Create a scene with 2 actors spaced apart
7. Attach the custom script to one Actor 1
8. Set the Actor A and Actor B parameters to each actor respectively

**Expected behavior**
Actor B's position changes to Actor A's

**Platform (please complete the following information):**
 - OS: [e.g. Win 10]
 - App version [e.g. 3.0.3]

Pandelume:
I'm convinced custom scripts can't affect actors other than the player. My custom scripts set to any other actor instead affect the player. Try with straightforward ones like Hide Actor, Set Actor Sprite Sheet, or Actor Move To and see. I expect your issues may be related to mine.

PR #1149, opened 2022-07-19
Fix setting actor's position to another actor's position

patrickmollohan:

Fixes issue where setting an actor's position to another actor's X/Y coordinates causes the second actor to move instead of the first.

Setting an actor's position to another actor's X/Y coordinates causes the second actor to move instead of the first. I believe variableSetToProperty is overwriting which actor is active, so calling actorSetActive before variableFromUnion was ineffective.

Setting an actor's position to another actor's X/Y coordinates now works as intended.

None that I am aware of.

* **Other information**:
~~This issue might affect other events as well? Will have to test.~~ It doesn't appear any other events are affected by this. Hurray! :D
Although it initially appeared to fix issue #1098, there is still something wrong with setting an actor's position to another actor's position via custom scripts. However, at least part of the issue should be resolved now.

### DIFF
--- a/src/lib/events/eventActorSetPosition.js
+++ b/src/lib/events/eventActorSetPosition.js
@@ -71,14 +71,15 @@ const compile = (input, helpers) => {
     variableFromUnion,
     temporaryEntityVariable,
   } = helpers;
-  actorSetActive(input.actorId);
   if (input.x.type === "number" && input.y.type === "number") {
     // If all inputs are numbers use fixed implementation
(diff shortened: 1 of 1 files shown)


chrismaltby:
Thanks @patrickmollohan must have missed this one, but just confirmed this is fixing the setPosition issue. I've also made this change 7718c8073156481df2698b66bdf09880b81b7944 which should fix the issue seen when using properties within custom scripts (it was always reading the player's properties)

PR #1154, opened 2022-07-21
Prevent user from creating projects ending with a period

patrickmollohan:

This PR fixes an issue with project names.

Currently, if a Windows user creates a project with a project name ending with a period, the project folder is inaccessible and cannot be deleted by normal means.

This now prevents a period from being entered as the last character in the project name and warns the user.

Users now can't add a period to the end of their project names. I can't imagine this is a deal-breaker.

* **Other information**:
There might be other invalid characters that can be entered this way, resulting in folders that certain operating systems can't read. Might be good to eliminate them all at some point. This was just one I noticed when trying to create a project ending with a period myself. Whoops! Haha.

### DIFF
--- a/src/components/app/Splash.tsx
+++ b/src/components/app/Splash.tsx
@@ -166,9 +166,14 @@ export default () => {
   };
 
   const onChangeName = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const newName = e.currentTarget.value;
+    var newName = e.currentTarget.value;
+    if (newName.endsWith(".")) {
+      newName = newName.replace(/.$/,"");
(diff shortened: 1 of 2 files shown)


chrismaltby:
Thanks for spotting this @patrickmollohan I've just merged a slightly different fix as we already had a function for stripping out characters that were invalid filenames across different systems https://github.com/chrismaltby/gb-studio/commit/fe3ffe611340cb7e6817cb8c0fc8162e27ad3ee2 Just added your regex to the rest. Still credited you with the change though.

This version doesn't show the error message but as it's quite a rare case I think it's not too important.

patrickmollohan:
@chrismaltby whoops, not sure why my search for "strip" didn't return that function in VS Code, but yeah that is much better!

PR #1155, opened 2022-07-22
Enhancements to "Loop For" event

patrickmollohan:

Adds the ability to specify the comparison and operation performed by the for loop.

Currently, the for loop only checks if "From" is >= "To", and increments the variable by the specified amount. This limits the usefulness of the for loop, as you can either increase the variable to a given value, or create an infinite loop (for var = 0; var <= 5; var += -1).

Now you can specify the comparison and the math operation. Comparisons include "==", "!=", "<", "<=", ">", and ">=". Math operations include "+=", "-=", "*=", "/=", and "%=".

None that I am aware of.

* **Other information**:
Does this need a migration script?


pau-tomas:
While I see the usefulness of this changes I also feel it makes the default loop event more complicated than it needs to be, with `Loop While` already being the complex version of loop events that allows to do what this PR add. If not merged it could be released as an `Advance Loop` plugin.

patrickmollohan:
@pau-tomas I see your point, but I respectfully disagree with that assessment. Adding in two dropdowns doesn't really increase the difficulty of using it. As a matter of fact, I found using the for loop as is to be frustrating in that it only works if "From" is less than "To". Let's say you have an animation state that you want to play forwards and backwards. You can use the "for" loop for the forwards part (for variable = 0; variable <= 4; variable++), but it does not work in reverse (for variable = 4; variable > 0; variable--). I would have to create a new animation state that reverses the animation and use the forwards for loop, which is kinda hacky. And while you can use a "while" loop as a "for" loop or vice versa, again, I feel that's kinda hacky. A "for" loop should be more useful than what it currently is. Having said that, if the consensus is to rebrand this as an "Advanced Loop", I can get it done, but I don't think that's the right approach; to me, that would just be unnecessary clutter (from a GB Studio user's perspective).

chrismaltby:
Thanks @patrickmollohan I've made a few small changes to your PR so it looks like this now
Screenshot 2022-07-25 at 23 23 13
I've changed the operator dropdowns to show mathematical notation, but it still shows the full text labels in the menu.


Was originally trying to do this simpler, this was my inspiration
Screenshot 2022-07-25 at 23 30 38

First thought was to switch to a `>=` comparison when the incBy value was negative but since the increment can be a variable at run time, don't want to have to do an extra check on the variable from a GBVM script. I think you're right we just need to give people the choice.

Will need a migration too but when I get a min I can write that and update this PR before merging

patrickmollohan:
@chrismaltby looks good! Yeah I can see the appeal for something simpler like that for sure. I did think of one other alternative an hour ago: a loop with no variable/from/to/comparison/operation, just a simple "Loop For X Times" with only a number field and the body. That way if someone wants to do something X number of times, they don't have to deal with variables and such and would be even simpler to use. I can submit a PR for that if you think that'd be useful!
Sorry about the migration script; I will eventually learn to write those!

chrismaltby:
Thanks @patrickmollohan I've just added the migration and merged.

"Loop N Times" could be a useful simpler event too, I'm currently starting to limit what I'm merging to prepare for a 3.1 release, trying to do mostly bug fixes for now but we could add an event like that after launching the next update

patrickmollohan:
@chrismaltby awesome!
Alright, sounds like a plan! I'll get something prepped in the mean time so it'll be ready for whenever you want it!

PR #1156, opened 2022-07-23
Add "Loop Do While" event

patrickmollohan:

Adds a do while loop.
The event has the "true" path above the expression field intentionally to indicate to the user that the true path gets executed before the expression is evaluated.

Currently, there is no "do while" loop, only "for" and "while".

Adds "do while" loop.

None that I am aware of.

* **Other information**:
For that one programmer in the world that uses "do while" loops, you're welcome! Haha.

It might be important to note in the documentation that if you put a "do while" loop within an actor's "On Update" script, it will run the body every time the "On Update" loops, so it may seem like it's not checking the condition, even though it is. "Do While" is most effective in areas that only run once, such as in "On Init" scripts or if you attach a script to a button. A similar warning should also be included about the "for" loop, since it modifies the value of the variable every time the "On Update" runs.

### DIFF
--- a/src/lib/events/eventLoopDoWhile.js
+++ b/src/lib/events/eventLoopDoWhile.js
@@ -1,0 +1,42 @@
+const l10n = require("../helpers/l10n").default;
+
+const id = "EVENT_LOOP_DO_WHILE";
+const groups = ["EVENT_GROUP_CONTROL_FLOW"];
+
+const autoLabel = (fetchArg, args) => {
+  if (args.expression) {
+    return l10n("EVENT_LOOP_DO_WHILE_LABEL", {
+      expression: fetchArg("expression"),
+    });
+  } else {
+    return l10n("EVENT_LOOP_DO_WHILE");
+  }
+};
+
+const fields = [
+  {
+    key: "true",
+    type: "events",
+  },
+  {
+    key: "expression",
+    type: "matharea",
+    rows: 5,
+    placeholder: "e.g. $health >= 0...",
+    defaultValue: "",
+  },
+];
+
+const compile = (input, helpers) => {
+  const { doWhileExpression } = helpers;
+  const truePath = input.true;
+  doWhileExpression(input.expression || "0", truePath);
+};
+
+module.exports = {
+  id,
+  autoLabel,
+  groups,
+  fields,
+  compile,
+};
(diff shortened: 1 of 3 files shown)


maxoakland:
What’s the difference between while and do while?

Btw, great job on all these new features. Very useful!

patrickmollohan:
@maxoakland a "while loop" checks the condition first, then if true, executes its body, whereas the "do while loop" does the opposite: executes the body first, then checks the condition. The "do while loop" ensures that the events within it gets executed at least once, as opposed to the "while loop" where the body may never execute at all depending on the conditions. I very rarely see "do loops" being used in programs, but I figured while I was in that area of code anyway for the "for loop" enhancements, I might as well implement it.

Thanks! Between participating in the recent GMTK Game Jam and being inspired to GameBoy-ify a certain 40-year old game, I've been implementing things I find to be helpful. I hope others can make use of these things too!

maxoakland:
> @maxoakland a "while loop" checks the condition first, then if true, executes its body and loops back afterwards, whereas the "do while loop" does the opposite: executes the body first, then checks the condition and loops back if true. The "do while loop" ensures that the events within it gets executed at least once, as opposed to the "while loop" where the body may never execute at all depending on the conditions. I very rarely see "do loops" being used in programs, but I figured while I was in that area of code anyway for the "for loop" enhancements, I might as well implement it.
> 
> Thanks! Between participating in the recent GMTK Game Jam and being inspired to GameBoy-ify a certain 40-year old game, I've been implementing things I find to be helpful. I hope others can make use of these things too!

Oh, OK I can see how that would be quite useful. What game are you inspired to Gameboy-if?

patrickmollohan:
Well, not that I'm quite ready to announce anything yet (need to figure out how to tell if the player is overlapping a specific pixel value of the background in order to make one of the core features of the game work properly), but as a wee teaser, this is the error screen for debugging purposes:
image

src/lib/events/eventLoopDoWhile.js
```diff
@@ -0,0 +1,42 @@
...
+    rows: 5,
+    placeholder: "e.g. $health >= 0...",
+    defaultValue: "",
+  },
+];
+
+const compile = (input, helpers) => {
+  const { doWhileExpression } = helpers;
+  const truePath = input.true;
+  doWhileExpression(input.expression || "0", truePath);
```
pau-tomas, line 33: The scriptBuilder.ts changes that include `doWhileExpression` are missing from the commit

patrickmollohan:
Closing as it is out of date. May consider resubmitting in the future.

PR #1157, opened 2022-07-23
Add "If Current Scene Is" event

patrickmollohan:

Adds "If Current Scene Is" event, which checks if the current scene equals the selected scene.

No way to check if the current scene is a specific scene aside from using variables and assigning each scene a value.

Can now check if the current scene is a specific scene.

None that I am aware of.

* **Other information**:


maxoakland:
This is great! I’ve been wanting this

patrickmollohan:
Thanks, I'm glad to hear so! Hope you get to use it soon!

chrismaltby:
Hi @patrickmollohan unfortunately because you've got the condition checking happening at compile time it doesn't work if you use it within a custom script as all the same compiled script is used for all scenes. It does work if you add the script directly into a scene but I image most use cases would probably want to have the check in a shared script.

Instead I think you'd need to read the value from `far_ptr_t current_scene` from `data_manager.c` and compare the `current_scene.bank` and `current_scene.ptr` match the corresponding values for the selected scene.

I'll have a go at making this change and see if it works.

chrismaltby:
Hi @patrickmollohan I've just modified this to use a runtime check rather than a compile time one. Seems like it should work. 

Do you mind testing at your end too and if this is also working for you then I can merge.

patrickmollohan:
@chrismaltby oh, right! That makes sense. Sorry about that; I guess I never tried using it in a custom script before, but that makes a lot more sense now. Based on my testing, everything appears to be working perfectly now!

PR #1166, opened 2022-07-27
Add "Store Digit In Variable" event

patrickmollohan:

Adds the ability to store a digit of a number/variable into a variable. Works up to the ten thousands digit, since the 16-bit limit on numbers/variables is up to five digits. Example of a use could be for creating scoreboards.

You can perform the mathematics behind this using the built-in math functions, but for some, it might not be straightforward or obvious.

Now, you can quickly and easily get the specified digit of a number/variable.

None that I am aware of.

* **Other information**:
~~Would there be a need to add properties to the union? I can't think of a reason to need a digit from, say, the player's X coordinate, but maybe someone would? It would be easy enough to add in.~~ Went ahead and added it anyways. Maybe someone could find a use for it.


patrickmollohan:
Closing as it is out of date. May consider resubmitting in the future.

PR #1169, opened 2022-07-29
Tracker fixes and improvements for effects and pasting

RichardULZ:

Improvement

Could not see the new effect editor in tracker view, 
Placing a new effect in piano roll would default to 000.
New
Copying a selection was not ModPlug compatible
Pasting a partial selection overwrote other parameters

Selecting any effect column in tracker will open the effect editor
Placing an effect in piano roll will keep the last selected effect data for quick placement (or 0)
New
Copying a selection includes spaces for non overwriting selection and is ModPlug compatible
```
GBStudio hUGETracker paste format compatible with...
ModPlug Tracker  XM
|   01......
|   ........
|   .....022
|   .....201
```
Pasting merges without overwriting where there are spaces
Mouse release on tracker grid will select a region (cheap drag select)
Remake tracker "activeField" when changing patterns for fast past in place

It uses a bit of a hack to preserve effect parameter data, as it seemed any effect other than arpeggio 0 would clear the parameters. Dispatch changes twice works.
Tracker selection rectangle still redraws the entire tracker, which remains slow when shift arrow keys.

* **Other information**:
Want to add!
Increment / decrement all values in a selected region column (ctrl & ctrl Shift with + - or scroll wheel)
(this feature is over 50% of my music workflow)
Edit 2: Added! 
Does not set field to use when moving with just arrow keys
 but clicking or any selection works flawlessly.

### DIFF
--- a/src/components/music/SongTracker.tsx
+++ b/src/components/music/SongTracker.tsx
@@ -21,7 +21,7 @@ import { getKeys, KeyWhen } from "lib/keybindings/keyBindings";
 import trackerActions from "store/features/tracker/trackerActions";
 import clipboardActions from "store/features/clipboard/clipboardActions";
 import { clipboard } from "store/features/clipboard/clipboardHelpers";
-import { cloneDeep, mergeWith } from "lodash";
+import { clamp, cloneDeep, mergeWith } from "lodash";
 
 interface SongTrackerProps {
   sequenceId: number;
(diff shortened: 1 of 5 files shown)


src/components/music/SongTracker.tsx
```diff
@@ -497,6 +544,23 @@ export const SongTracker = ({
...
+          if (e.deltaY < 0) return transposeSelectedTrackerFields(1,true);
+          if (e.deltaY > 0) return transposeSelectedTrackerFields(-1,true);
+        } else {
+          if (e.deltaY < 0) return transposeSelectedTrackerFields(1,false);
+          if (e.deltaY > 0) return transposeSelectedTrackerFields(-1,false);
+        }
+        return;
+      }
+    },
+    [activeField, selectionRect, pattern]
```
pau-tomas, line 561: You don't need this dependencies for this callback but just `transposeSelectedTrackerFields`. 

I use an eslint rule to warn me about those locally https://github.com/facebook/react/issues/14920

pau-tomas: Just noticed that once you add `transposeSelectedTrackerFields` as a dependency it'll mean that this functions will be updating on every render. To solve that you'd have to wrap  `transposeSelectedTrackerFields` on its own useCallback with its own dependencies.

RichardULZ:
Not so sick anymore! 
Merged these commits from pautomas (without prettify and callback solved twice)

https://github.com/pau-tomas/gb-studio/tree/rulz_tracker_improvements_followup
https://github.com/chrismaltby/gb-studio/commit/69d24afcd28b6a14ca5ee6b01dfdb39fa6b15cb5

Selection preview changes on mouse move, and a fair bit faster! 

Notes:
Holding shift and the arrow key can still bog down lower end systems, but it is faster. 
Arrow key navigation does **not** let you copy, or ctrl scroll a single value unless you press shift (selection mode).

### Grander scheme
Trominal's template .uge instruments must be added for release!
Note step size needs a button! Pautomas worked on one.
Note step is how far the cursor should advance when you press a note, current 1, i use 0 in openMPT, some like larger spacing for timing.

pau-tomas:
I think I got the ctrl+a to select a channel from OpenMPT. Is it not common? I wonder if we could have ctrl+A once to select a channel and if you do it again it selects the whole pattern.

RichardULZ:
OpenMPT uses Shift + 5 apparently 🤷 
Ctrl Q and A are transpose note (specifically note)... I've always preferred Ctrl+- or scroll.
Milky uses Ctrl A to select the whole pattern

I like the Double tap Ctrl A idea, to select whole pattern, we will need to remove the transpose bindings for that.

pau-tomas:
Ugh, I forgot about transpose (and don't know where I got the idea for ctrl+a to select a channel, then 🤣). Tracker keybindings are nuts.

Issue #1188, opened 2022-08-16
Timer event broken in multiple ways in 3.1 alpha0


Diablohead:
**Describe the bug**
The Timer event only works if the timer is set to 0.3 seconds or below
Using the timer event twice in a scene only allows one to work
Time set (0.1, 0.2 or 0.3) all update at the same rate

**To Reproduce**
create actor, create an onupdate timer event to position relative to itself every 0.4

**Expected behavior**
Timer event should repeat based on given time
Multiple timer events should function together

**Screenshots**

**Platform (please complete the following information):**
 - GBstudio 3.1 alpha0 build x86 64bit version

**Additional context**
Attached sample project: 
[Timer event bug v3_1_a0.zip](https://github.com/chrismaltby/gb-studio/files/9352990/Timer.event.bug.v3_1_a0.zip)

pau-tomas:
>  Using the timer event twice in a scene only allows one to work

That's the behavior of timer since they were introduced. Since 3.0 GBVM supports multiple timers, but the IDE hasn't been updated with that functionlity just yet.

Diablohead:
> > Using the timer event twice in a scene only allows one to work
> 
> That's the behavior of timer since they were introduced. Since 3.0 GBVM supports multiple timers, but the IDE hasn't been updated with that functionlity just yet.

Thanks for the reply, good to know about the multiple use limit.

pau-tomas:
@patrickmollohan is quick #1189 ;)

patrickmollohan:
@pau-tomas doing what I can to help haha. I'll probably replace the dropdown with something like the button selector for "Attach Script To Button" for aesthetic reasons, but it should work for now

pau-tomas:
ooh, I see what's happening. Since you're setting the timer inside an on update script, it means that each frame(ish) the timer is overrode by the new one and hence restarting the countdown. So, if the time is too low it's very possible that the timer will be replaced and restarted before it even executes.

You can use a `Wait` event instead of a timer to achieve what you want.

Diablohead:
> ooh, I see what's happening. Since you're setting the timer inside an on update script, it means that each frame(ish) the timer is overrode by the new one and hence restarting the countdown. So, if the time is too low it's very possible that the timer will be replaced and restarted before it even executes.
> 
> You can use a `Wait` event instead of a timer to achieve what you want.

[...]

PR #1189, opened 2022-08-16
Add timer select to all timer events

patrickmollohan:

Adds a timer select to all timer events. Enhancement to help with issue #1188 

Currently, only timer 1 can be used, even though there are four timers available.

You can now select which timer to use.

None that I am aware of.

* **Other information**:


pau-tomas:
(Following up from the other thread) You can use the same field type as the save slot selector on the save/load data events.

patrickmollohan:
@pau-tomas thanks for that! Saved me quite a bit of time there!

chrismaltby:
Very nice! Thanks @patrickmollohan I've just fixed a few merge conflicts that happened with some of the late 3.1.0 release changes and added a migration script so that existing projects added the first timer by default (compiling was working but the UI element was acting as if no timer was selected when opening old projects) but other than that looks good.

I'm going start working through reviewing/merging your PRs now, but might take me a little while :-)

PR #1201, opened 2022-09-04
Fix for build error, cannot read property matchAll of undefined

barrosfilipe:

Bug fix

Project doesn't build
```
- TypeError: Cannot read property 'matchAll' of undefined
```

- Fixed the `generateGameGlobalsLookup` function checking for it's param before doing the `matchAll` check.

No

### DIFF
--- a/src/lib/compiler/objCache.js
+++ b/src/lib/compiler/objCache.js
@@ -70,6 +70,7 @@ const generateIncludesLookup = async (buildIncludeRoot) => {
 
 const generateGameGlobalsLookup = (gameGlobalsContents) => {
   const lookup = {};
+  if (gameGlobalsContents === undefined || !gameGlobalsContents) return lookup;
   const globalMatches = [
     ...gameGlobalsContents.matchAll(/([A-Za-z_0-9]+)[\s]*=[\s]*([0-9]+)/g),
   ];


chrismaltby:
Hi @barrosfilipe I think this will remove the error but not actually fix the issue, by the point that function is called game_globals.i should definitely exist and without it being available we'll likely have issues with caches being invalid.

Do you have any steps to reproduce this issue (I'm not seeing it myself)? Are you using the application or the command line tool? Does it affect all projects or just specific ones?

One thing that could help is adding some logging to `fetchCachedObjData` and `cacheObjData`

around here
```
  const includesLookup = await generateIncludesLookup(buildIncludeRoot);
  const gameGlobalsLookup = generateGameGlobalsLookup(
    includesLookup[GAME_GLOBALS_FILE]?.contents
  );
  ```
 add a `console.log(includesLookup)` that should show something like this:

Screenshot 2022-09-04 at 12 47 31

with `data/game_globals.i` being present 

If not will need to trace through why the file isn't there yet

The file contents gets generated here
https://github.com/chrismaltby/gb-studio/blob/8be58707647a2eb363b051218062f386dfe2cec3/src/lib/compiler/compileData.js#L1734

This function passes the contents to ejectBuild where it gets written to disk then calls makeBuild which is doing that cache read
https://github.com/chrismaltby/gb-studio/blob/8be58707647a2eb363b051218062f386dfe2cec3/src/lib/compiler/buildProject.js#L29-L47

File gets written to disk here
https://github.com/chrismaltby/gb-studio/blob/8be58707647a2eb363b051218062f386dfe2cec3/src/lib/compiler/ejectBuild.js#L148

Cache read happens here
https://github.com/chrismaltby/gb-studio/blob/8be58707647a2eb363b051218062f386dfe2cec3/src/lib/compiler/makeBuild.js#L66

**Comment by chrismaltby** [comment]
Just had a thought, I'm going to guess your `includesLookup` has `data\game_globals.i` with a `\` rather than `/`. Mac/Windows differences :-)

**Comment by barrosfilipe** [comment]
Hey @chrismaltby! Thanks for all the info around the issue!

I thought the fix would be enough due to the following line.
```
includesLookup[GAME_GLOBALS_FILE]**?**.contents
```
Although I was wondering if that would actually fix the issue rather than just hiding it.

When it comes to reproduce the issue, someone from Discord posted the issue there and I pulled the branch, tested myself and got the same result.

> - Building through the GUI application
> - Windows 10

Where you not able to reproduce the issue?

chrismaltby:
Hi @barrosfilipe I think I've got a fix now with 0036fe517d380c7ebb052465581087a764f9c59e which has been merged to develop. Managed to get access to a Windows machine and was able to reproduce the issue. Think this PR isn't needed anymore, thanks for looking into the issue though!

PR #1230, opened 2022-10-21
Rename Obj Palette selector to Monochrome Palette

pau-tomas:

Small silly improvement, I guess

The tile palette selector in the sprite editor is called `Obj Palette` and `OBP0` and `OBP1` which are terms that don't mean much without good understanding of how the GB works.

Rename `Obj Palette` to `Monochrome Palette` and `OBP0` and `OBP1` to `Palette 1: OBP0` and `Palette 2: OBP1` also added a palette preview to both options in the dropdown. 

No

* **Other information**:

Not sure how needed is this, but it was annoying me a bit :)


maxoakland:
This is good because I originally thought you could use this to make sprites choose the darker color on GBC. I think this clears up any possible confusion

PR #1266, opened 2023-02-16
Fix muting bug with FXH parser

datguywitha3ds:
It adds a two line bug fix to an error Beatscribe and I found with the FX Hammer parser, the first NR51 write ignores the used/unused channels, thus muting the unused channel.

If a sound effect only uses one channel, there's the chance the opposite channel will get muted via a null NR51 write.

A failsafe check for CH2Pan and CH4pan that ORs the necessary bits if CH2Pan or CH4Pan are 0.

No.


chrismaltby:
Thanks @coffeevalenbat, merged now. Sorry it took me 10 months to get around to reviewing this!!

PR #1312, opened 2023-12-03
Some QOL for logo scenes

pau-tomas:
**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature/improvements

**What is the current behavior?** (You can also link to an open issue here)
Logo scenes have some quirks that confuse users from time to time. 

- Added a warning to the scene when the background of a logo scene is used in a non-logo scene (see notes below)
image

- Added counter for sprites in scene info (max 12 sprite tiles)
- Removed the limit of zero actors for Logo scenes (actors are allowed but limited by number of sprites)
image

- Logo scenes don't need a default player sprite. Removed from the option from settings.
image

**Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

**Other information**:

I'm not super confident about the implementation for the background already used in logo scene warning, as it will add a slow down to scene selection (due to having to check all other scenes for type and backgrounds).

Something I'd like to add in the future is a warning when text/dialogues are used in scripts inside a logo scene and, maybe, filter them out from the add event list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add ability to set background tile priority for Color games using Priorty tool in colorize section. Priority tiles appear above sprites
 - Add support for UGE v6 to music editor [@pau-tomas](https://github.com/pau-tomas)
 - Subpattern editor added to Instrument Editor [@pau-tomas](https://github.com/pau-tomas)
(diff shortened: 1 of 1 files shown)


pau-tomas:
Right, the more information the better. Maybe just grey them out and disable with a ℹ️ icon and a tooltip or something like that.

maxoakland:
> Right, the more information the better. Maybe just grey them out and disable with a ℹ️ icon and a tooltip or something like that.

I think that's a great idea

PR #1182, opened 2022-08-12
Add the ability to launch a projectile at another actor

patrickmollohan:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our [guidelines](/chrismaltby/gb-studio/blob/develop/.github/COMMIT_MESSAGE_GUIDELINES.md)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Adds the ability to launch a projectile at another actor.

* **What is the current behavior?** (You can also link to an open issue here)
Currently, there's no easy way to launch a projectile at another actor.

* **What is the new behavior (if this is a feature change)?**
You can now select "Actor Position" under the "Direction" field of the projectile event to launch a projectile at another actor.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
None that I am aware of.

* **Other information**:
Requires [pull request 155 of GBVM](https://github.com/chrismaltby/gbvm/pull/155) for the atan2 function.

PR #1356, opened 2024-02-10
Add support for launching projectile towards target actor

pau-tomas:

Feature

The only way to launch a projectile towards the position of an actor is by using ATAN2 in a GBVM event (introduced in 3.2)

A new `Target` option in the Direction field for the Launch projectile event 

launch_projectile_target

Also added a new event field type for angles that displays the degrees (it's a common question in Discord about how angles work) and support for `atan2` to the Math expression events.

image

No


patrickmollohan:
Already opened a PR for this a while ago: #1182. Love the angle field and atan2 math expression though!

pau-tomas:
Oh! I didn't notice! I'll rebase on top of yours and we can merge this one first

patrickmollohan:
Haha no worries, even I had to search for it! Sounds like a plan!

pau-tomas:
Thanks! I rebased into this one. All should work without conflicts if your PR gets merged first. 

I fixed an issue for custom events. On the event if other actor id key isn't unique it won't be detected as an actor parameter inside the custom events. I renamed the key from `otherActorId` to `targetActorId`.

Other than that there's only some minor differences in naming (Actor Pos vs Target) which I don't feel super strong so I'll let @chrismaltby decide on that one.

patrickmollohan:
Awesome work! Thanks for fixing up the issue as well; I think I ran into that before and couldn't figure out why. Makes sense now haha. I really need to go through my old PRs and clean them up a touch.

Actually, I like "Target"; I think that might be better than my attempts at naming things. I'd be totally good with renaming it to "Target" if Chris is!

chrismaltby:
I went with this PR in the end but updated the changelog to credit you both :-)

PR #1357, opened 2024-02-11
Add loop animation checkbox to set state event

pau-tomas:

Feature

You can't pick if the animation will loop or not when changing state. The only way to stop the animation from looping is to use a timer or wait and set animation speed to None.

A checkbox in the Set Animation State event allows to check if the animation should loop or not.

image

No. The event should default to looping for older versions without need of migration.

* **Other information**:

Might be interesting to expose the other actor flags with an event too.


maxoakland:
> Might be interesting to expose the other actor flags with an event too.

What are the other actor flags?

pau-tomas:
> What are the other actor flags?

From https://www.gbstudio.dev/docs/scripting/gbvm/gbvm-operations/#vm_actor_set_flags

.ACTOR_FLAG_PINNED - pin/unpin the actor (which you seen the plugin already)
.ACTOR_FLAG_HIDDEN - hide/show actor 
.ACTOR_FLAG_COLLISION - disable/enable collision
.ACTOR_FLAG_PERSISTENT - set persistent actor flag (this is the one that is set when the run on update when not visible checkbox is true on the On Update script)

maxoakland:
> > What are the other actor flags?
> 
> From https://www.gbstudio.dev/docs/scripting/gbvm/gbvm-operations/#vm_actor_set_flags
> 
> .ACTOR_FLAG_PINNED - pin/unpin the actor (which you seen the plugin already) .ACTOR_FLAG_HIDDEN - hide/show actor .ACTOR_FLAG_COLLISION - disable/enable collision .ACTOR_FLAG_PERSISTENT - set persistent actor flag (this is the one that is set when the run on update when not visible checkbox is true on the On Update script)

Oh yes, that would be very useful

PR #1359, opened 2024-02-14
[Feature] Add magnitude field to camera shake event

patrickmollohan:

Adds a magnitude field to the camera shake event so that you can control how strong the shake is.

Currently, the camera shake is locked to a range of -5 to +5.

You can now specify the magnitude, resulting in a range of -magnitude to +magnitude, defaulting to 5 to preserve the original shake.

Not that I'm aware of. I think I finally figured out the migration script?

* **Other information**:
I set a maximum magnitude of 255 arbitrarily; realistically, I don't think anyone would want to go higher than 30, as the shake is very intense. The value can be lowered, if desired, to a more focused range.


chrismaltby:
Perfect, thanks @patrickmollohan I just fixed up the tests as `npm test` was failing and added a few more to cover this new functionality but other than that all good 👍

PR #1362, opened 2024-02-22
Update/improve German translation

gonzoMD:

 Sync strings and partially translate them.

The current German translation is partially wrong (seems machine generated).
Many strings are missing and/or untranslated.

a better and more complete translation

Not that I knew.

* **Other information**:

the problem of missing strings in other languages makes it slightly harder for translators to do their work. Maybe the should be put in sync and in the future taken into account when adding new strings.
I did not finish the whole translation and I am forced with other projects, but offer this PR as is, because it would be a pity if the changes would go lost.


chrismaltby:
Thanks for your contribution @gonzoMD not sure I've documented this anywhere easy to find but there's actually a way of determining all the untranslated strings for any language file with a script
```npm run missing-translations de``` this will list out the untranslated content and merge the english values into `src/lang/de.json` I'll add something to the README file about this to help people working on translations in the future

PR #1368, opened 2024-03-14
Update maker-appimage

patrickmollohan:

Updates maker-appimage to be compatible with the latest changes.

As reported on Discord, building the Linux AppImage fails due to using an older version of maker-appimage.

maker-appimage is now working again with no issues.

None that I'm aware of.

* **Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "@electron-forge/plugin-webpack": "^7.3.0",
     "@electron/fuses": "^1.7.0",
     "@electron/notarize": "^2.3.0",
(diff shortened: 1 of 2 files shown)


chrismaltby:
I wish I'd seen this before spending the morning debugging this issue :-) I ended up using an even newer version of the module which ended up requiring switching to Node 21. It seems to build okay but I have no way of testing it currently. If there's still issues with this version let me know and I'll roll back to this one you've used here. Thanks for looking into this!

PR #1399, opened 2024-05-06
Allow to rename variable flags

pau-tomas:

Feature

Flags can't be renamed in the Variable Flags events.

Allows to rename flags for global and local variables that are named. 

gbs_flag_rename_wip

No. Flag events should continue working as expected.

* **Other information**:

This PR only allows to rename flags for variables that are named. Creating a new entity for flags felt like too much just for this so I reused the existing variable entity and added a `flags` field. A variable entity only exists when it's named. 

I think it'd be possible to either make `variable.name` and `variable.symbol` optional or create a named variable with default names (ie: `Variable 100`) when flags are renamed for non-named variables. The former feels the correct way but the latter feels less risky.

It's also possible that allowing only to rename flags for named variables makes sense as it's "good practice". But in this case maybe there needs to be some tip about why flags can't be renamed in that case.


marco-afonso:
Awesome work! Just here to say that I need this exact feature very badly, and very glad to see it here! haha Would love to also have a toggle flag event

maxoakland:
This is amazingly useful

maxoakland:
> Awesome work! Just here to say that I need this exact feature very badly, and very glad to see it here! haha Would love to also have a toggle flag event

Toggle is a good idea. Please suggest it here on Github

PR #1412, opened 2024-05-19
Update Japanese JSON file with fully localized strings

tomo666:
Updates Japanese language file for the resource strings available as of 4.0.0 beta. Previous translations are now fully reviewed and fixed.

More than >90% of strings found to be English bleeding, and the existing translations contains term inconsistencies and contextual issues.

Now the app supports Japanese language, ready for non-English speakers to dive into dev using the app.

No

* **Other information**:
None


chrismaltby:
Amazing, thanks for all the effort put into this update!

PR #1433, opened 2024-05-24
Fix Hardcoded strings for L10N ready

tomo666:
Fixed the hard-coded UI strings to be localize-ready.
Please see attached report for the fixed sections and tests.
[GBStudio_L10n_ENBleed_Report.pptx](https://github.com/chrismaltby/gb-studio/files/15435549/GBStudio_L10n_ENBleed_Report.pptx)

There are still quite a few hard-coded strings that are limiting the translators to fully localize the app. Some are very noticeable and can impact the target language user’s usability.

Previously strings showing up in English (when switching languages) are now showing up in its local language.

No. New keys added to the localization json file that are not localized will remain in English as usual.

* **Other information**:
This fix PR only addresses the most noticeable issues during the app usage. I’m aware of few other hard-coded strings throughout the code but those are either low priority to fix or not reproducible on screen so I haven’t touched any of the code for it. Only relevant ones are suggested to be fixed for this beta.
The updated jaJP json file is also committed in this pull request, that was also used in testing.


chrismaltby:
Really appreciate the effort you've put into this PR with a full QA document and everything! 

I've tidied up a few things but I'm going to have a think how we can  handle the palette names without needing a deepcopy and try to fix that before merging

tomo666:
Thanks for looking into this and tidying up (I'm still learning how the GBS codebase works so apologies for any mess-arounds). Had a look into all the changes so far and looks darn elegant! Looking forward to how we can avoid the deep copying issue.

chrismaltby:
I managed to remove the deep copying by moving the localisation to happen earlier, at the point where palettes are fetched from the Redux store https://github.com/chrismaltby/gb-studio/pull/1433/files#diff-496f45e398e749283729a37ee84aead71fcd6053991aa5bba13830bf899e5e91R3808-R3836 using [createSelector](https://redux-toolkit.js.org/api/createSelector). From all my tests this seems to give the same results as your implementation (though please let me know if I broke anything!)

Thanks again for the effort you put into this!

Issue #1452, opened 2024-06-14
VM_IF and VM_IF_CONST dont do what vm.i says they do


DavidMeagher1:
vm.i says for VM_IF and VM_IF_CONST that @param N is the "Number of values to be removed from the stack if the result is true."
but looking at the code you can see that the stack pointer is reduced regardless of if the statement was true or not.

```c
void vm_if_const(SCRIPT_CTX * THIS, UBYTE condition, INT16 idxA, INT16 B, UBYTE * pc, UBYTE n) OLDCALL BANKED {
    INT16 A;
    if (idxA < 0) A = *(THIS->stack_ptr + idxA); else A = script_memory[idxA];
    UBYTE res = FALSE;
    switch (condition) {
        case VM_OP_EQ: res = (A == B); break;
        case VM_OP_LT: res = (A <  B); break;
        case VM_OP_LE: res = (A <= B); break;
        case VM_OP_GT: res = (A >  B); break;
        case VM_OP_GE: res = (A >= B); break;
        case VM_OP_NE: res = (A != B); break;
    }
    if (res) THIS->PC = pc;      //// THESE TWO LINES ARE THE ISSUE
    if (n) THIS->stack_ptr -= n; ////
}
```
putting the second if statement `if (n)` into the first one would fix this.

chrismaltby:
Hi @DavidMeagher1 thanks for spotting this, I've checked with the dev who wrote that code and it's just a case of the documentation being incorrect. The engine code is stored in a different repo, so I'll update the docs there to correct this. Thanks

chrismaltby:
I've fixed this now in
GBVM: https://github.com/chrismaltby/gbvm/commit/4dc5117660d100f92b4ede0a386e774a38d46e3e
which I have scripts to auto update
GB Studio: https://github.com/chrismaltby/gb-studio/commit/483074e5d5099678e6ef8ea34d1481618c068438
GB Studio Docs: https://github.com/chrismaltby/gb-studio-docs/commit/f3a4032fb92e1591bc55d6283217bc0f789c1e26
The docs site is on a new branch which won't go live until GB Studio 4 launches officially

PR #1453, opened 2024-06-14
Fix issue #1452

DavidMeagher1:
This is a bug fix for https://github.com/chrismaltby/gb-studio/issues/1452 making sure that VM_IF and VM_IF_CONST align with documentation

Users who rely on the fact that VM_IF and VM_IF_CONST pop values from the stack regardless of the result will have to add ways to clean the stack if the result is false.


untoxa:
the doc is wrong! it must always pop if you ask instruction to pop explicitly, no matter of the result.

PR #1466, opened 2024-06-25
add "linux" to maker-zip target for building generic linux binaries

Hellscaped:
- [x] Tests for the changes have been added (for bug fixes / features) No official tests needed, this just allows for building for generic linux.

Adds "linux" to the maker-zip target so it will allow building on generic Linux.

It refuses to build maker-zip on Linux, hence my hacky fix in the [gb-studio-git](https://aur.archlinux.org/packages/gb-studio-git) package.

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -17,7 +17,7 @@ module.exports = async () => {
       },
       {
         name: "@electron-forge/maker-zip",
-        platforms: ["darwin", "win32"],
+        platforms: ["darwin", "win32", "linux"],
       },
       new MakerAppImage({}),
       {


chrismaltby:
Happy to include this :-) Are you maintaining a copy of GB Studio in the Arch Linux repository then? That's great, thanks for looking after this!

Hellscaped:
> Happy to include this :-) Are you maintaining a copy of GB Studio in the Arch Linux repository then? That's great, thanks for looking after this!

oops, forgot to reply when this was closed.

I'm maintaining the gb-studio-git PKGBUILD, which is basically a glorified shell script that says how to assemble a package. the -git packages take the source code, and compile them into an arch package. The old -git PKGBUILD took the debian package and squashed it through dpkg in order to convert it, which is bad practice. This just allows building generic linux binaries, for repackaging for other platforms.

PR #1482, opened 2024-07-12
Fix IME confirmation issue for non-English text

tomo666:

Fixes IME confirmation issues while typing in non-English text in input fields.

GBS_IME_issue1
GBS_IME_issue2

When we input text using IME (i.e. Writing out Japanese, Chinese, Korean characters), we tend to hit the Enter key to accept the character conversion whilst typing a sentence. However, GBS will treats the Enter key presses as form input acceptance, which cancels out the IME input and looses focus from the input field. As a consequence, we cannot continue typing until the end of sentence.

This issue seems like it is present on Chrome running on a Mac (confirmed on both Intel based and M1). Windows 11 (confirmed on a Surface Laptop) does not seem to have this issue.
Hence, on a macOS, we need a way to distinguish the Enter key pressed specifically for IME input and other Enter key presses.

A workaround to this issue is to detect if user is composing text using IME or not and dismiss the Enter key press event when typing something with IME.

This PR fix will distinguish between the Enter key pressed specifically for IME input and other Enter key presses.

As far as I know, this issue occurs during renaming and inserting search text with:

[RenameInput.tsx]
- Custom script renaming
- Scene name renaming
- Global variable renaming
- Background image name renaming
- Tileset name renaming
- Music name renaming
- Instrument name renaming
- Sound effects name renaming

[AnimationStateSelect.tsx]
- Sprite state renaming

[PalettePage.tsx]
- Palette name renaming (both in navigation pane and in the editor)

[FlagField.tsx]
- Flag name renaming

[ScriptEditorEvent.tsx]
- Renaming event labels

[VariableSelect.tsx]
- Variable name renaming in event editor components

[AddScriptEventMenu.tsx]
- When searching event components with IME ON

[AddReferenceMenu.tsx]
- When searching references with IME ON

[ReferencesSelect.tsx]
- When renaming references to both asset & variables

(Basically, any input field that accepts the Enter key as form-input acceptance has the issue)

No.

* **Other information**:

This fix should work with Japanese, Chinese, Korean, or any other languages that uses IME.

### DIFF
--- a/src/components/forms/AddReferenceMenu.tsx
+++ b/src/components/forms/AddReferenceMenu.tsx
@@ -359,6 +359,11 @@ const AddReferenceMenu = ({ onBlur, onAdd }: AddReferenceMenuProps) => {
     (state) => state.project.present.settings.musicDriver
   );
 
+  const [isComposing, setComposition] = useState(false);
+  const onRenameCompositionStart = () => setComposition(true);
+  const onRenameCompositionEnd = () => setComposition(false);
+  var isFocusOut = false;
+
   useEffect(() => {
     const allOptions = ([] as (EventOptGroup | EventOption)[]).concat([
       {
@@ -550,7 +555,14 @@ const AddReferenceMenu = ({ onBlur, onAdd }: AddReferenceMenuProps) => {
         setSelectedIndex(Math.max(selectedIndex - 1, 0));
         scrollIntoViewIfNeeded(selectedIndex - 1);
       } else if (e.key === "Enter") {
-        onSelectOption(selectedIndex);
+        if (!isComposing || isFocusOut) {
+          onSelectOption(selectedIndex);
+          isFocusOut = false;
+        } else if (isComposing) {
+          // We cannot set isComposing to false as state here as it will trigger back to true again when the Enter key is pressed
+          // So instead, we set a flag to immediately focus out when user is not composing in IME mode and Enter key is entered
+          isFocusOut = true;
+        }
       }
     },
     [
@@ -605,6 +617,8 @@ const AddReferenceMenu = ({ onBlur, onAdd }: AddReferenceMenuProps) => {
             placeholder={l10n("TOOLBAR_SEARCH")}
             onKeyDown={onKeyDown}
             onChange={onChangeSearchTerm}
+            onCompositionStart={onRenameCompositionStart}
+            onCompositionEnd={onRenameCompositionEnd}
           />
         </SelectMenuSearchWrapper>
         <SelectMenuOptionsWrapper
--- a/src/components/forms/AnimationStateSelect.tsx
+++ b/src/components/forms/AnimationStateSelect.tsx
@@ -126,6 +126,11 @@ const AnimationStateSelect = ({
     spriteStateSelectors.selectAll(state)
   );
 
+  const [isComposing, setComposition] = useState(false);
+  const onRenameCompositionStart = () => setComposition(true);
+  const onRenameCompositionEnd = () => setComposition(false);
+  var isFocusOut = false;
+
   const onRenameStart = () => {
     if (currentValue) {
       setEditValue(currentValue.label);
@@ -144,7 +149,14 @@ const AnimationStateSelect = ({
 
   const onRenameKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
     if (e.key === "Enter") {
-      onRenameFinish();
+      if (!isComposing || isFocusOut) {
+        onRenameFinish();
+        isFocusOut = false;
+      } else if (isComposing) {
+        // We cannot set isComposing to false as state here as it will trigger back to true again when the Enter key is pressed
+        // So instead, we set a flag to immediately focus out when user is not composing in IME mode and Enter key is entered
+        isFocusOut = true;
+      }
     } else if (e.key === "Escape") {
       setRenameVisible(false);
     }
@@ -198,6 +210,8 @@ const AnimationStateSelect = ({
           onKeyDown={onRenameKeyDown}
           onFocus={onRenameFocus}
           onBlur={onRenameFinish}
+          onCompositionStart={onRenameCompositionStart}
+          onCompositionEnd={onRenameCompositionEnd}
           autoFocus
         />
       ) : (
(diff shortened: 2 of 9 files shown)


chrismaltby:
Hi @tomo666 oh this is an interesting problem 😅 I've tried enabling Japanese IME input and can reproduce the problem and your PR does seem to fix it BUT I want to tidy up a few things before merging

1. I'm not sure `isFocusOut` is actually doing anything, with a React functional component it would never have got set to true, you'd have to use a setState hook if you really needed it

```
      } else if (e.key === "Enter") {
        if (!isComposing || isFocusOut) {
          onSelectOption(selectedIndex);
          isFocusOut = false;
        } else if (isComposing) {
          // We cannot set isComposing to false as state here as it will trigger back to true again when the Enter key is pressed
          // So instead, we set a flag to immediately focus out when user is not composing in IME mode and Enter key is entered
          isFocusOut = true; // <--- This doesn't change the value higher up because of how React functional components work
```
changing this to just
```
      } else if (e.key === "Enter") {
        if (!isComposing) {
          onSelectOption(selectedIndex);
        } 
```
seems to work fine though. Is there a reason `isFocusOut` is needed that I'm missing?

2. The amount of places where the fix was needed is making me think I need I to combine some of this shared logic into one or two new components. My worry is that with this code across multiple components it would be easy for me to forget it when adding a new similar component later, having it only in one place will be a lot easier going forwards. Maybe something like `<SubmittableInput onSubmit={myFn}/>` where onSubmit is handling the IME/Enter logic, then all the places that are using inputs with Enter handling would switch to using this instead. 

I'm happy making both of these changes myself (especially 2. as I'll need to refactor a few things) but I'd appreciate your input on 1. as I don't want to remove `isFocusOut` if it's needed for something

tomo666:
@chrismaltby Thanks for your inputs.
I'm pretty sure I had to use a flag variable (isFocusOut) because the isComposing state was triggered again before the Enter key in one place. I might be wrong. Let me check again and try to reproduce that issue.
I strongly agree with having a single component to handle this sort of thing, but didn't want to do this myself since I didn't know to what extent you prefer other people messing with your code :)
Anyways, I will check about the flag this weekend and provide you with some findings/outputs.

chrismaltby:
I've started trying to refactor this in the branch `wip/ime-support` if you want to have a look.

I've moved most of the logic into a `useIMEIsComposing()` hook and then made an `<IMEInput />` component using the hook that's a drop in replacement for <input> but blocks keydown events while composing. I had to make it block ALL keydown events as I found a bug in your implementation on AddScriptEventMenu where if you started composing then pressed escape keyboard navigation using Enter would stop working (the UI got stuck thinking it was composing), it seemed safer to just not try to do anything custom while composing.

So far I've only updated <AddScriptEventMenu /> and <RenameInput /> (currently only used when renaming items in left hand sidebar but this probably should be used everywhere there's a renaming input) but I'll update the others later. I'm away now for the weekend but I can look at this again start of next week.

tomo666:
@chrismaltby Thanks for sharing the wip branch.

I just figured out that not all scripts require isFocusOut flag, but some do.
When I started fixing, I tackled with the RenameInput.tsx file first, which required this flag. Without isFocusOut, the user cannot enter out from the current editing mode, after completing IME input and pressing the Enter key.

To reproduce the issue, simply omit the isFocusOut logic part in either "RenameInput.tsx" or "AddReferenceMenu.tsx":

`else if (e.key === "Enter") {
        if (!isComposing) {
          onSelectOption(selectedIndex);
        } `

Then, type in something using IME, and after confirming input, hit "Enter" to try exiting out from the edit mode.
The result is, you can never exit out from edit mode.

So from my findings, these are the files that require "isFocusOut" flag:

**Requires "isFocusOut":**

[RenameInput.tsx]

Custom script renaming
Scene name renaming
Global variable renaming
Background image name renaming
Tileset name renaming
Music name renaming
Instrument name renaming
Sound effects name renaming

[AddReferenceMenu.tsx]

When searching references with IME ON


Other files did not require. I'm sorry I thought they were all the same and blindly used the same logic that includes the isFocusOut flag.

tomo666:
Just in case, I've updated the code in 239241c which eliminates the unwanted usage of isFocusOut flag.

chrismaltby:
Thanks @tomo666 following your steps I see what was happening, I can see why adding isFocusOut to those inputs is helping but it's actually introducing a few very subtle opportunities for bugs in the future. I'm not sure if you have ESLint configured (you can also run `npm run lint` to see any errors/warnings but you had warnings listed about missing dependency in the second argument to `useCallback` (https://react.dev/reference/react/useCallback#parameters)

Screenshot 2024-07-22 at 10 58 17

fixing these for example in RenameInput.tsx
```
  const onRenameKeyDown = useCallback(
    (e: React.KeyboardEvent<HTMLInputElement>) => {
      if (e.key === "Escape") {
        onRenameCancel();
        setName(value ?? "");
      } else if (e.key === "Enter") {
        if (!isComposing) {
          onRenameComplete(name);
        }
      }
    },
    [name, onRenameCancel, onRenameComplete, value, isComposing] // <-- added isComposing state to dependencies causing this function to always have the latest value of this state when called
  );
  ```
prevents the issue you were seeing but without having to add `isFocusOut`.

Don't worry about updating anything in your branch though I can take these learnings to complete the refactored branch I was working on.

Thanks again for your help with this

chrismaltby:
I've implemented this fix based of your PR here https://github.com/chrismaltby/gb-studio/commit/2156d4b58938cc9d112682b435ec37ab453ea1ad and credited you with the fix :-)

If any other fields need fixing in the future we'll now be able to replace `<Input />` with `<IMEInput />` and `<input />` with `<IMEUnstyledInput />` with no other changes needed (in theory)

I'll close this PR as it shouldn't be required now :-) Thanks again

tomo666:
@chrismaltby Wow, did not know that command. Never used React before I started looking into GBS, so that's something I learnt today :)
After installing ESLint to VSCode and examining the errors and warnings, I see that I was missing the dependency hook (as you said) that was causing the need to introduce the flag var. Got it working flawlessly by fixing this in my code too.

Many thanks on going through refactoring and fixing this issue (and for being such a great tutor!). Much appreciated!

PR #1494, opened 2024-07-31
Add compiler optimisation settings

patrickmollohan:

Adds two options to the "Preferences" screen to control compiler settings. The first, "Compiler Preset", changes the value of "max-allocs-per-node", allowing for the user to select a tradeoff between compiling the ROM faster and compiling the ROM more optimally. The second option, "Optimise ROM For", allows the user to select whether to optimise for faster code ("opt-code-speed"), smaller code ("opt-code-size"), or none.

No current way to set these compiler optimisations.

Compiler optimisations are now selectable if desired.

None that I'm aware of.

* **Other information**:
The code should default to the same settings GB Studio uses right now, so there shouldn't be any change to the ROM output out of the box.

### DIFF
--- a/src/components/app/Preferences.tsx
+++ b/src/components/app/Preferences.tsx
@@ -38,6 +38,8 @@ const Preferences = () => {
   const [musicEditorPath, setMusicEditorPath] = useState<string>("");
   const [zoomLevel, setZoomLevel] = useState<number>(0);
   const [trackerKeyBindings, setTrackerKeyBindings] = useState<number>(0);
+  const [compilerPreset, setCompilerPreset] = useState<number>(3000);
+  const [compilerOptimisation, setCompilerOptimisation] = useState<number>(0);
 
   const currentZoomValue = zoomOptions.find((o) => o.value === zoomLevel);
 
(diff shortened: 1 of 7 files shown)


chrismaltby:
Very nice, thanks @patrickmollohan I'm going to make a few changes before merging, but overall this works well :-) 

Not changing a huge amount but I'm going to move this to be a per project setting (rather than per app) so probably in the "Build Options" section of Settings, also you'd used "electron-settings"  in buildMakeScript.ts which would I think break the CLI tool that doesn't use Electron, finally I think I just need to make sure it rebuilds the project when you correctly when you change the setting, think that's just adding the setting to env in makeBuild.ts so it won't use cached files with the old setting.

I get the feeling that making this changes will conflict with the branch [feat-splitproject](https://github.com/chrismaltby/gb-studio/tree/feat-splitproject) that I'm currently working on to split the .gbsproj file into smaller resource files, so I'm going to wait until I've finished/merged that first before updating this PR

chrismaltby:
Hi @patrickmollohan I've adapted this PR into this change https://github.com/chrismaltby/gb-studio/commit/0d8a3a95065d558ea931bb97ee7fbc6610684971 instead. Much of the code is the same and I've credited you in the changelog. I've just moved the UI to the settings page instead.

In testing I found a few combinations of settings that didn't work, it would fail to compile some of the files (usually one of platform.c, gbprinter.c, ui.c) so I've added a message that appears if both settings have been modified. I also made the select dropdown show the flag that's being set for clarity.

Screenshot 2024-08-14 at 17 25 48

PR #1503, opened 2024-08-13
Removing the hard reserving tile for the default ui system.

Mico27:
For people who have some scenes that have more than 192 unique tiles (up to 256 in normal mode and up to 512 in color-only mode) and dont use the default ui system. issues fixed:
- The background image compiler would cap at 192 unique tile, not allowing any way to have more tiles even with a plugin fix
- Removed ALLOC_BKG_TILES_TOWARDS_SPR definition in the engine as the new allocation hard caps at 192 tiles
This fixes a breaking change that was introduced in GBS 4 by adding this hard cap.

### DIFF
--- a/appData/src/gb/src/core/data_manager.c
+++ b/appData/src/gb/src/core/data_manager.c
@@ -16,8 +16,6 @@
 #include "data/spritesheet_none.h"
 #include "data/data_bootstrap.h"
 
-#define ALLOC_BKG_TILES_TOWARDS_SPR
-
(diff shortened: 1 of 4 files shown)


Mico27:
Ignore this pull request, I'll do some more work on it to be able to specify for a background which tile allocations style a user wants instead. As I am aware that this will not allows sprite overflowing in the shared vram.

chunkysteveo:
That's a great idea about utilising the extra tiles if the UI elements aren't used in a scene. Keen to see this come to life if possible!

PR #1516, opened 2024-08-16
Add missing localization for Minimize menu

Pomdap:
Add an entry for the minimize menu in the "en" lang file and reference it into the menu.ts code

Bug fix

Minimize menu is not localized

Minimize menu has in own entry in the lang file

No

* **Other information**:


chrismaltby:
Oh, weird. I was pretty sure that menu item text localisation was handled at the operating system level, but just testing now, yes it looks like it needs to be in the language files. Thanks for sorting

PR #1542, opened 2024-08-29
Update ja.json localization for new dialog and channel related events

tomo666:
Update ja.json localization for new dialog and channel related events


tomo666:
@chrismaltby No problems :)
If you prefer me to update the translations at a certain point of time, plz let me know. Otherwise, I will just update and send new PRs as new commits are created.

PR #1547, opened 2024-08-29
Update ja.json localization for camera properties and printer

tomo666:
Update ja.json localization for camera properties and printer


chrismaltby:
Thanks again :-) I added a few more last night with the introduction of Actor Effects and Thread Start/Stop events. I'm going to try to stop there before the next release now!

PR #1590, opened 2024-09-19
Common tileset optimizations on compilation

Mico27:
-Optimizes compilation of background and tilesets to prevent data redundancy and unused generated data.

-currently tilesets compiled with common tilesets duplicates data despite having the exact same data
- currently background with a common tilesets in all scenes will still compile a version without the common tileset even if never used.

- reuses tilesets with the exact same data instead of duplicating it
- dont compile backgrounds without a common tileset if there is no scene that uses the background without a common tileset

No breaking changes or user changes needed.

* **Other information**:
Example project that uses a common tileset among 13 scenes gives me a 30% size reduction of the rom used with the optimization (% size reduction is even greater the more scene I add):
image


maxoakland:
This is great! Improved efficiency is very welcome

PR #1605, opened 2024-10-14
engine.json defined collision layers (flags, masks, colors, names, and icons)

Q-Bert-Reynolds:

feature

Collision tile colors, names, and icons are fixed as is the opacity with which the collision layer is drawn. If your project uses the spare collision layers, your maps are coated with meaningless numbers and letters. If your project doesn't need slopes, you're welcome to do so in code, but the editor will still show slopes.

Collision tile colors, icons, and names can be edited in `engine.json`, per sceneType.
A layer opacity slider is now present on the collision tiles brush toolbar.

no

* **Other information**:

{C6466697-246C-4E64-9DE1-940A3AB5D376}
{464F3DB5-93B2-4304-AA63-AA37AD75C88E}

This last image is no longer relevant now that collision labels are set in engine.json instead of the settings page, but I'm leaving it for posterity:
{191D958C-BECE-4854-B22B-CC51C762E1D8}


maxoakland:
This is cool and very useful

I think it should be something you change in the App settings (on Mac it's GBStudio > Settings…) instead of for each game in the game's settings screen. Unless there's some reason people would want to change it per game, which I don't think I would

Q-Bert-Reynolds:
> I think it should be something you change in the App settings (on Mac it's GBStudio > Settings…) instead of for each game in the game's settings screen. Unless there's some reason people would want to change it per game, which I don't think I would

A default preference could be useful for setting up new projects, but per game settings are the whole point. I can even see a case for tying collision layer settings to scene type (ie. Point and Click vs Platformer settings).  The game I'm working uses tile map layers for things like changing dmg palettes, background priority, and palette switching. The collision icons and colors were basically meaningless to me at that point. If I were to open an older project, though, I would expect to see the tile data the way it was used in that game.

src/shared/lib/resources/types.ts
```diff
@@ -558,6 +558,15 @@ export const ScriptEventPreset = Type.Object({
 
 export type ScriptEventPreset = Static<typeof ScriptEventPreset>;
 
+export const CollisionSetting = Type.Object({
+  key: Type.String(),
+  name: Type.Optional(Type.String()),
+  color: Type.String(),
+  icon: Type.Optional(Type.String()),
```
Q-Bert-Reynolds, line 565: I never ended up using this, but the plan was show a custom image. Instead, I just use the first character of name for the icon.

src/components/forms/CollisionLayerPicker.tsx
```diff
@@ -0,0 +1,188 @@
...
+          case "bottom": 
+            name = l10n("FIELD_COLLISION_BOTTOM");
+            icon = <BrushToolbarTileBottomIcon $color={layer.color} />; 
+            break;
+          case "left": 
+            name = l10n("FIELD_COLLISION_LEFT");
+            icon = <BrushToolbarTileLeftIcon $color={layer.color} />; 
+            break;
+          case "right": 
+            name = l10n("FIELD_COLLISION_RIGHT");
```
Q-Bert-Reynolds, line 87: There are a lot of these massive switch blocks because of how the bits of a tile are used (ie. the upper nibble is used as flags, but the lower nibble is used as an index). Perhaps some more options for how those bits can be used would be useful.

maxoakland:
> > I think it should be something you change in the App settings (on Mac it's GBStudio > Settings…) instead of for each game in the game's settings screen. Unless there's some reason people would want to change it per game, which I don't think I would
> 
> A default preference could be useful for setting up new projects, but per game settings are the whole point. I can even see a case for tying collision layer settings to scene type (ie. Point and Click vs Platformer settings). The game I'm working uses tile map layers for things like changing dmg palettes, background priority, and palette switching. The collision icons and colors were basically meaningless to me at that point. If I were to open an older project, though, I would expect to see the tile data the way it was used in that game.

I like that idea. Personally, I think I would use the universal setting but I can see why those other ideas would be useful

chrismaltby:
Nice feature! I'm going to make a few changes before merging, started a branch at https://github.com/chrismaltby/gb-studio/tree/feat/pr-1605 just got a few UI tweaks I'd still like to make but overall looking good

pau-tomas:
Could this be a type of global plugin instead of a project setting? 🤔

chrismaltby:
If I'm not mistaken the need to relabel and rename the collision tiles is a per project thing. e.g if you're using the additional tiles in a custom engine then you can label/color them differently to indicate what they're being used for in your specific game.

This makes me think maybe this should not expose be exposed in the IDE but instead put the labels/colors in `engine.json` allowing engine plugins to set the labels for additional tiles they use. Any use of these tiles would need a replaced engine anyway.

I like the ability to change the collision layer opacity, which I've moved below the brush toolbar like this
Screenshot 2024-10-18 at 17 42 54

but actually the more I think about it, I think the tile label replacement should be handled by engine plugins and not exposed in the app directly

src/components/forms/CollisionTileLabelsPicker.tsx
```diff
@@ -0,0 +1,280 @@
...
+          onChange({
+            key: value.key,
+            name: e.currentTarget.value,
+            icon: value.icon,
+            color: value.color,
+          });
+        }}
+      />
+      <StyledColorInput
+        // type="color" //truncates alpha value
```
Q-Bert-Reynolds, line 235: `type="color"` gives you the nice color picker, but it truncates the alpha value. Without transparency, the extra tiles look like they're replacing a solid tile instead of being placed in addition to it. It also means I can't make some tiles really faint when they're less important. Killing this and reverting some of the styling was the quick and easy fix, but perhaps there's a way to style the color picker popup and configure it to allow alpha values? If not, maybe a separate opacity slider.

chrismaltby: Hmm, yeah I noticed that switching to the default colour picker also removed the alpha channel for each tile. Personally I didn't mind as I found it confusing to have the two alphas for each tile (the per tile and the global one) and wasn't sure why setting to full alpha on the settings page wasn't showing as I expected in the world view (because I didn't notice the global setting) but if that feature is important to you we can keep it.

I think I might adapt this PR to work as I described above instead though:
- Keep the collision layer opacity setting placed in a panel just below the brush toolbar
- Move the data to be defined in the `engine.json` file rather than exposed in the IDE on the settings page
- I'll can allow colours defined in with alpha channels 

But I think this could be an opportunity to allow the collisions tiles to be much more configurable, so in the engine.json
- allow specifying exactly which bits are set
- allow setting which collision tiles are enabled on a per scene type basis, e.g. ladders only make sense when using platform scenes so don't show them in the toolbar for anything else and that value could be reused for something else on other scene types

I need to think on this a bit more before attempting

chrismaltby: I've extracted the layer opacity portion of this PR and merged (+credited you in changelog) here 0a88cfacd1709b8c64e2e836d0fb62f070138fb8. Will pick this up again next week to have think how best to handle the collision tile replacement portion 👍

Q-Bert-Reynolds: You could use mask and set bits. COLLISION_BOTTOM, for example, sets the lower nibble to 2 (ie.  (tile & 0xF0) | 0x02). It's mask is 0xF0 and its set bits are 0x02. This could also be stored as a single string, XXXX0010. When you draw your brush toolbar, you group buttons by similar mask bits. Mask and flag bits for the current tile layers looks like this:

```
solid              XXXX1111
top                XXXX0001
bottom             XXXX0010
left               XXXX0100
right              XXXX1000
ladder             0001XXXX
slope_45_right     00101000
slope_45_left      00110100
slope_22_right_bot 01110010
slope_22_right_top 01111010
slope_22_left_top  01110110
slope_22_left_bot  01110010
spare_08           1000XXXX
spare_09           1001XXXX
spare_10           1010XXXX
spare_11           1011XXXX
spare_12           1100XXXX
spare_13           1101XXXX
spare_14           1110XXXX
spare_15           1111XXXX
```

It's clean, groups the default tiles the same way, and allows you to ditch the "extra" collision data for slopes. Might even go so far as to name those mask groups and give each of them an opacity slider, removing the need for per-tile opacity.

Q-Bert-Reynolds: I went ahead and implemented this. No more customization in settings. Added to engine.json in sceneTypes like this; 
`{
      "key": "METROIDVANIA",
      "label": "GAMETYPE_METROIDVANIA",
      "files": ["include/states/metroidvania.h", "src/states/metroidvania.c"],
      "collisionTileLabels": [
        { "flag": 15, "mask": 15, "color": "#FF0000FF", "icon": "█", "name": "FIELD_SOLID" },
        { "flag": 1, "mask": 1, "color": "#FF0000FF", "icon": "▘", "name": "FIELD_COLLISION_TOP_LEFT" },
        { "flag": 2, "mask": 2, "color": "#FF0000FF", "icon": "▝", "name": "FIELD_COLLISION_TOP_RIGHT" },
        { "flag": 4, "mask": 4, "color": "#FF0000FF", "icon": "▖", "name": "FIELD_COLLISION_BOTTOM_LEFT" },
        { "flag": 8, "mask": 8, "color": "#FF0000FF", "icon": "▗", "name": "FIELD_COLLISION_BOTTOM_RIGHT" },
        { "flag": 16, "mask": 16, "color": "#00A000FF", "icon": "H", "name": "FIELD_LADDER" },
        { "flag": 32, "mask": 32, "color": "#80808080", "icon": "+", "name": "FIELD_PRIORITY" },
        { "flag": 64, "mask": 64, "color": "#2040F040", "icon": " ", "name": "FIELD_PROP_WATER" },
        { "flag": 128, "mask": 128, "color": "#F080F080", "icon": "s", "name": "FIELD_SCRIPT" }
      ]
    },`

Q-Bert-Reynolds:
It says no conflicts, but I resolved the conflicts in the web editor and haven't tested it yet.

chrismaltby:
This is definitely getting there.

What do you think about this change allowing the icons (both in toolbar and what gets overlaid on the scene view) to be defined in engine.json too?

https://github.com/chrismaltby/gb-studio/commit/578bacbfcee64dfdacdf8380b35045dd25fec589

I've not created icons for all the tiles yet but you get the idea

Screenshot 2024-10-23 at 13 06 22

icons are just defined as a mask stating which pixels should be drawn with the tile colour.

I just made a small script to quickly get the values for a few of the icons
```js
// Create ladder icon
[
"xx    xx",
"xxxxxxxx",
"xxxxxxxx",
"xx    xx",
"xx    xx",
"xxxxxxxx",
"xxxxxxxx",
"xx    xx"
].map((line) => parseInt(line.split("").map((c) => c!==' ' ? 1 : 0).join(""),2).toString(16).padStart(2,"0").toUpperCase()).join("")
// Outputs C3FFFFC3C3FFFFC3
```

I've not had chance to investigate but it seems theres a few issues with this PR though:
- Shift click to select multiple walls doesn't seem to be working anymore. Previously you could shift click a left wall + a right wall to draw both, seems this implementation only currently allows drawing one or the other
- The slope drawing tool is acting weird, you can draw slopes with it and they work in game but the tiles are invisible on the collision tiles layer (but weirdly drawing the slope collision tiles manually works fine)

Also I noticed you've had a few linting warnings in your code some that could have caused re-rendering issues (missing dependencies on some useEffects). I'm not sure what text editor you're using but there's likely an `ESLint` plugin for it which should flag, and potentially suggest fixes for these. You can also run `npm run lint` to list warnings in a terminal too. (speaking of which I just noticed I've got a warning in an unrelated file so I'll fix that now)

chrismaltby:
I fixed the slope tile issue, it looks like you'd changed the behaviour of slope tiles to no longer be one-way collisions 
this change fixed them to display correctly again c04c35392ec2e9d455317070bb55f9b97263d632
I've also added the icons for 22deg tiles here e4b6ec92acd2618eea449c97790850d6183fcf41

chrismaltby:
Screenshot 2024-10-23 at 14 14 23

Also playing around with the idea of allowing the hex value of each tile to be displayed that could make it a bit easier to see what values are being used and to debug issues when multiple collision tile types are overlaid on each other

Q-Bert-Reynolds:
> What do you think about this change allowing the icons (both in toolbar and what gets overlaid on the scene view) to be defined in engine.json too?

The original plan was to just define a path to a texture. I think that's probably still the path of least resistance (pun fully intended).

chrismaltby:
> > What do you think about this change allowing the icons (both in toolbar and what gets overlaid on the scene view) to be defined in engine.json too?
> 
> The original plan was to just define a path to a texture. I think that's probably still the path of least resistance (pun fully intended).

Hmm, so making pngs with an alpha mask and filling in with the chosen colours? Probably would end up working about the same. Think I’d prefer to keep this method rather than bringing in more files for the images 👍

Q-Bert-Reynolds:
Sick. I look forward to the next round of you sifting through my half baked pull requests.

PR #1610, opened 2024-10-18
Add Animation States To Global Header File

Mico27:
Add animation states to the game_globals.h file (not just the game_globals.i file) so that they can be referenced in engine code.

Animation states are currently only in game_globals.i file, so to be able to refer to them in engine code you have to manually create the defines. However using this method, somtimes the animation states values change and you have to recreate the defines by copy pasting from the game_globals.i file everytime that happens.

This will add the animations state includes in the game_globals.h file which can be referenced in engine code without having to define them manually.

Not that I'm aware of

* **Other information**:


chrismaltby:
For some reason I thought these were already there, thanks for adding

PR #1614, opened 2024-10-19
Update ja.json for projectiles, collision opacity, vram debug labels

tomo666:
Update ja.json for projectiles, collision opacity, vram debug labels


chrismaltby:
Thanks! I'm spending a few weeks travelling in Japan very soon, very excited to visit your country :-)

PR #1627, opened 2024-11-02
Add support for value field to Wait event

pau-tomas:

Feature

Wait event only supports numbers

Wait event now supports value field 

No. I think I implemented the migration properly.

* **Other information**:

It might be a bit annoying that switching between seconds and frames will switch the value as they're two different fields?


pau-tomas:
I was pretty sure I wrote the migration well this time (console.log aside, I should add a lint check or something for those). Sorry!

PR #1632, opened 2024-11-06
Add Custom Scripts and Comment with string to Add Event Menu

pau-tomas:

Feature

To add a call to a custom script the user has to add the Call Script event and then select the script. 

Custom Scripts appear in the search results of the add menu event, when selected, they a Call Script event is added with the script field defaulting to the given custom script.

image
image

Additionally, if no event exists for the given search string a `Comment "search string"` option appears in addition to the `Display Dialogue` one that allows to add a comment event.

image

No

* **Other information**:

I added the `displayLabel` field to avoid showing all the custom scripts if `Call Script` (or any query that will contain those words) is typed in the box... but maybe that's not that of a bad idea?


maxoakland:
Thank you so much for doing this. It's been one feature I really wanted

PR #1738, opened 2025-05-27
Feature: Allow setting the horizontal/vertical flip attributes of a tile in color mode

Mico27:
Added in the color toolbar the posibility to mark background tiles as flipped horizontaly and or verticaly in color mode.
Also added the option to automaticaly detect which tiles are flippables.

https://github.com/user-attachments/assets/bd0384c0-db8d-4a65-b4fd-99f7b6fd67a9

### DIFF
--- a/src/components/ui/icons/Icons.tsx
+++ b/src/components/ui/icons/Icons.tsx
@@ -99,6 +99,21 @@ export const AutoColorIcon = () => (
   </svg>
 );
 
+export const AutoFlipIcon = () => (
+  <svg width="24" height="24" viewBox="0 0 16 16">
+    <path fill="#444" d="M0 15l6-5-6-4.9z"></path>
+	<path fill="#444" d="M9 10.1l6 4.9v-10l-6 5.1zM14 12.9l-3.4-2.8 3.4-3v5.8z"></path>
+	<path fill="#444" d="M7 5h1v1h-1v-1z"></path>
+	<path fill="#444" d="M7 3h1v1h-1v-1z"></path>
+	<path fill="#444" d="M7 7h1v1h-1v-1z"></path>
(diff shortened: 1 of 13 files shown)


patrickmollohan:
I think what @maxoakland is getting at (correct me if I'm wrong) is if the IDE can automatically detect this and optimise the tiles without human intervention, would there be a use case for the manual mode?

Mico27:
Well yeah, there can be cases when you dont necesseraly want a scene to have flipped tiles and have them be their one unique tile in the tileset VRAM, tileswapping comes to mind for example and it allows you more control.
Id rather have this being able to be toggable (as is currently implemented) rather than forced on users.

patrickmollohan:
I did think of that after the fact. Would it be possible to skip the optimisation if the engine detects tileswapping/other future tile-specific manipulations?

Mico27:
Personaly Id rather have the auto-detection done per scene (as it is currently implemented) like auto-color is. Useful if you are making a gbc+dmg game and have non-flipped scenes for dmg and flipped ones (and more detailed) for gbc.

nielsderijk:
Would love to use this to optimize my puzzle game's VRAM (I am using a lot of flipped tiles). 
Is there a way to run this on the latest version of the main branch?

nielsderijk:
Would love to use this on the world maps I'm building, this would save tons of VRAM!
@chrismaltby Would be really nice to include this before the level editor is added. Can help testing if needed!

maxoakland:
> Would love to use this on the world maps I'm building, this would save tons of VRAM! @chrismaltby Would be really nice if we had the flip option even before the level editor (if that’s still on the roadmap). Can help testing if needed!

I agree. At best you could double or even quadruple the amount of available tiles using tile flipping

PR #1745, opened 2025-06-01
Feature/add json support to plugins

kinostl:

Feature

Plugins do not support importing any type of file.

Adds the ability for plugins to load json files, allowing for field elements to be shared between plugins. Useful for getter/setter setups for example. It shouldn't be plausible to exploit this, as the json files are loaded using the mock function of NodeVM.

No

* **Other information**:
I'm opening this PR to start the conversation and get feedback on the idea. It currently does not meet what I would consider spec, as the watch code doesn't seem to update events when data is updated.


kinostl:
An example of usage.
image
image

kinostl:
It's probably that, I finished this at like 2 AM, lol.

I'm using it to help with structures, so that actors can have more complex attributes that work across scripts. I have an event that takes an actor and then you can modify an attribute on its attached structure if it's linked to one in the engine. 

Entirely possible with two events, but keeping those events synchronized is as pain.

kinostl:
I think extended js support would be reasonable if it worked similarly. Look for lib*.js and add them to the mock after running them through the VM to scrub them.

kinostl:
It was in fact the fact that json is not js

kinostl:
Rebased onto develop, was originally forked from v4.1.3

ReptiIe:
This is kind a nice. Would it be possible to... hmm.. develop this in a way to make a let we say
GBS -> IMPORT/EXPORT all messages (dialogs, text, etc) into JSON?

It might be helpful to create some multilanguage system in the future.

PR #1752, opened 2025-06-24
Changed the hardcoded way to mask collision tile data upon drawing to scene

Mico27:
Changed the hardcoded way to mask collision tile data upon drawing to scene and instead use the mask values specified in the tile definition in the engine.json for the specified scene type. (useful if you create custom collision tile definition for a scene and specify masks that dont fit the default's COLLISION_ALL (0x0f) and TILE_PROPS (0xf0) masks)

Example, in this case, I made a custom scene type that contains 3 collision types and 16 directions, the 16 directions could not fit in either of the default masks so it couldn't overlap the separate custom mask correctly: 
image
image

### DIFF
--- a/src/components/world/BrushToolbar.tsx
+++ b/src/components/world/BrushToolbar.tsx
@@ -170,7 +170,7 @@ const BrushToolbar = ({ hasFocusForKeyboardShortcuts }: BrushToolbarProps) => {
   const dispatch = useAppDispatch();
 
   const sceneId = useAppSelector((state) => state.editor.scene);
-  const { selectedPalette, selectedTileType, selectedBrush, showLayers } =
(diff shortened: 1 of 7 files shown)


chrismaltby:
Hi @Mico27 while working on improving the Adventure scene type to add support for rounded corner tiles I came across this exact same issue and ended up implementing a fix very similar to this (hadn't noticed you'd already fixed the issue!) After reviewing your implementation I've ended up going with a mix of mine + yours in this version
https://github.com/chrismaltby/gb-studio/commit/4c6bd925c2b22e34fade86b2cdef63e1f24efb38
Do you mind testing if this version still works for your pinball project?

Mico27:
Yep, works with the pinball project with custom collision tiles!

Issue #1779, opened 2025-08-01
Single 8x8 sprites are weirdly placed in the scene


morphoice:
Though looking correct in the editor
sprites with a size 8x8 have their tile weirdly positioned in the scene (approx 4px lower than its pink placeholder box)

pau-tomas:
Could you share screenshots of the scene and the sprite setup?

morphoice:
Ofc! Here's two screenshots of what it looks like in the sprite editor vs the placed sprite in a scene
it only happens when the sprite is set to the newly available 8px height

https://www.dropbox.com/scl/fi/5iyt6fodf6fzp1cx0dr8a/Screenshot-2025-08-02-at-09.58.25.png?rlkey=mreppdlz13sdw5ro4gqifyw8k&dl=0

PR #1782, opened 2025-08-02
Fix: Single 8x8 sprites are weirdly placed in the scene #1779

pau-tomas:

Fix

Sprites with 8px height displayed with a 4px vertical offset (as reported in #1779)

image

Fixes (somehow) that issue by setting the wrapper for the spritesheet canvas to `flex` (CSS 🤷 )

I don't think so

* **Other information**:


chrismaltby:
Oh sorry @pau-tomas didn't spot this, I've implemented almost the same fix, just using display:block on the canvas instead!

PR #1784, opened 2025-08-02
Add setting to decide if the build folder should open after export

pau-tomas:

Feature

The build folder always opens/focus after an export ROM/Web/Pocket build which is annoying for some users (latest report [here](https://discord.com/channels/554713715442712616/570925583421276160/1400029770930065458))

Added an option in the `Build Options` dropdown (true by default) to select if the folder should open or not.

image

The `Build Options` dropdown might be a bit hidden for this as you can't start an Export ROM build from there (only Run). But this option felt similar to `Generat Debugging Files` which only affects the Export ROM builds and is in that menu too.

No

* **Other information**:


chrismaltby:
Screenshot 2025-08-02 at 12 29 28
The options in that dropdown also exist in the Build Options pane in Settings, think it probably can go there too :-)

pau-tomas:
oh, missed that one. will add it now.

PR #1793, opened 2025-08-11
Add toggle for ROM usage view

pau-tomas:

Feature

We always show the ROM usage bar after a build. Some users get unnecessarily stressed by it thinking they'll run out of space soon. 

Implemented a toggle to show/hide the bar (hidden by default) as proposed on Discord [here](https://discord.com/channels/554713715442712616/570925583421276160/1404437430295199835)

No. Well, users will have to enable the toggle once for their project if they want to keep seeing the bar.

* **Other information**:

Another suggestion was to show the bar once you're over a certain threshold but I think that'd be more stressful for some users ("why is it appearing now all of a sudden? Am I going to run out of space soon?!")

### DIFF
--- a/src/components/debugger/DebuggerBuildLog.tsx
+++ b/src/components/debugger/DebuggerBuildLog.tsx
@@ -151,6 +151,9 @@ const DebuggerBuildLog = () => {
   const openBuildFolderOnExport = useAppSelector(
     (state) => getSettings(state).openBuildFolderOnExport,
   );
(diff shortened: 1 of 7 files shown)


maxoakland:
Seems totally reasonable to me. Sometimes it's good to put advanced features in the options

My only concern is I feel like we're starting to get to the point where options are spread out in too many areas. I think the best place for global options to go is the App Title > Settings window. This is also where I think Theme and language options should go. Maybe I should create a separate issue for this

PR #1795, opened 2025-08-13
Feature: Add contextual functions to reorder actors and triggers within a scene

Mico27:
Added contextual function that allows reordering actors and triggers order in a scene
Useful for users that wants to reorder actors/triggers to specific indexes for referencing in scripts.
The workaround before was to move the actors out of the current scene and re-add them in in the desired order.

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add zoom level 1600 when viewing assets
 - Add "If Engine Field Compare With Value/Variable" events
 - Add setting in Build Options to determine if build folder should be opened after exporting game [@pau-tomas](https://github.com/pau-tomas)
+- Add ability to reorder actors and triggers within a scene from context menus [@Mico27](https://github.com/Mico27)
 
 ### Changed
 


pau-tomas:
Nice! I had a very similar [branch](https://github.com/pau-tomas/gb-studio/tree/feat/scene_actor_reorder) lingering on my drafts for a while now (only for actors tho). One thing that is a bit counter intuitive is that, for actors, the ones at the bottom are the ones that, in Color Mode will render on top in the game. 

Other than that, I wonder if we should expose GBVM alias to reference the actors on a scene by their index in the form of `ACTOR_${scene_symbol}_${actor_symbol}_INDEX` (ie: `ACTOR_SCENE_1_ACTOR_3_INDEX` or `ACTOR_BATTLE_SCREEN_ENEMY_1_INDEX`)

maxoakland:
> Nice! I had a very similar [branch](https://github.com/pau-tomas/gb-studio/tree/feat/scene_actor_reorder) lingering on my drafts for a while now (only for actors tho). One thing that is a bit counter intuitive is that, for actors, the ones at the bottom are the ones that, in Color Mode will render on top in the game.
> 
> Other than that, I wonder if we should expose GBVM alias to reference the actors on a scene by their index in the form of `ACTOR_${scene_symbol}_${actor_symbol}_INDEX` (ie: `ACTOR_SCENE_1_ACTOR_3_INDEX` or `ACTOR_BATTLE_SCREEN_ENEMY_1_INDEX`)

I think it would make sense to flip the order to make it more intuitive

Mico27:
> Nice! I had a very similar [branch](https://github.com/pau-tomas/gb-studio/tree/feat/scene_actor_reorder) lingering on my drafts for a while now (only for actors tho). One thing that is a bit counter intuitive is that, for actors, the ones at the bottom are the ones that, in Color Mode will render on top in the game.
> 
> Other than that, I wonder if we should expose GBVM alias to reference the actors on a scene by their index in the form of `ACTOR_${scene_symbol}_${actor_symbol}_INDEX` (ie: `ACTOR_SCENE_1_ACTOR_3_INDEX` or `ACTOR_BATTLE_SCREEN_ENEMY_1_INDEX`)

Regarding the render order in color mode, that's not exactly right. What decides the order is the active list order, not the "real" order of actors. When you have actors offscreen they get disabled and removed from the active list, when they go on screen they are activated and pushed on top of the active actor list (so they are rendered behind every other actors) regardless of their real order in the actor list.
It just so happen that in the case that if all of your scene actors do happens to be on screen and that you never disable/enable an actor, the active actor list order will be the same as the real actor list order.
So its best to see this simply has being able to order the actor list of a scene and nothing else.

PR #1827, opened 2025-09-20
Added IDE feature allowing editing a metasprite origin/anchor point

Mico27:
Instead of the hardcoded origin/anchor point of metasprite being the x = (canvas width / 2 - 8) and y = (canvas height - 8), it is now editable. (currently setting it to 0 will reset it to its previous behavior values - alternative: initialize in migration code?)
(tested it in both 8x16 and 8x8 sprite modes) (also toggling the "show grid" button in the editor will also toggle the visibility of the origin/anchor point)

image


chrismaltby:
Ended up making a few changes but this is a nice update!

While updating I also fixed an inconsistency we had with positioning things on the sprite editor, previously Y coordinates for tiles and bounding boxes used +Y as upwards, -Y as downwards (opposite to everything else in the app where +Y was downwards), now +Y is downwards everywhere and tile/bounds coordinates are shown in the UI as values relative to the canvas origin. This required a small project migration but I've written this also and added some tests to verify this works as expected.

PR #1828, opened 2025-09-20
Added the feature to toggle the positioning of an actor between tiles and pixels

Mico27:
Added a toggle to allow setting an actor in a scene per pixel or per tiles
moving the actor in the scene (either by drag n dropping or by changing the coordinate values) will either move per tile or per pixels depending on the toggle set on the actor.

image

image


chrismaltby:
Nice! Made a few minor changes but seems to work well :-) thanks @Mico27

PR #1843, opened 2025-10-25
[Fix] Inconsistent dialogue close speeds

pau-tomas:

Bugfix

Using the "Close Non-Modal Dialogue" or "Overlay Move To event" moves the overlay at a slower speed than the "Text Close Speed" and "Text Open Speed" options in the Dialogue event.

I added the same conversion to the speed fields in the Close Non-Modal Dialogue event and Overlay Move To event.

I added a migration so old speed values on the events are migrated to the new ones.

* **Other information**:

Should we rename the `cameraSpeed` type and the CameraSpeedSelect attribute to OverlayMovementSpeed or something similar that sets the right values (ie: `-3` for instant) and that way we avoid having to do conversion on different events? 

That selector is only used for overlay related events at the moment.

### DIFF
--- a/src/lib/project/migration/versions/410to420.ts
+++ b/src/lib/project/migration/versions/410to420.ts
@@ -1,4 +1,5 @@
 import { TILE_SIZE } from "consts";
+import { eventHasArg } from "lib/helpers/eventSystem";
 import {
   ScriptEventMigrationFn,
   ProjectResourcesMigration,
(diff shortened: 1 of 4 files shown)


chrismaltby:
Hmmm yeah not quite sure how that happened but you're right, we've got an off by one error on some of the events handling overlay movement.

> Should we rename the cameraSpeed type and the CameraSpeedSelect attribute to OverlayMovementSpeed or something similar that sets the right values (ie: -3 for instant) and that way we avoid having to do conversion on different events?

I've adapted your PR to work like you suggest here, was a little odd that the field type was called cameraSpeed anyway and now all the fields store the values correctly so the `convertSpeed` function isn't needed anymore.
Updated version is here https://github.com/chrismaltby/gb-studio/commit/4254e04379e6ed331e1b5479af918c25f77b4f24 if you want to have a look. This is merged now so I'll close this PR

PR #1899, opened 2025-12-23
Implement _stackPushScriptValue method

Phidias618:
Added a method to push script values onto the stack,
akin to _stackPushEvaluatedExpression.

The new function allows to push the evaluated value of a "ScriptValue". 
Before, to achieve the same goal inside a javascript event file,
 the simplest way to do it was
```
_reserve(1);
variableSetToScriptValue(".ARG0", scriptValue_to_push)
```
Which results in slightly larger code even though the implementation
of _stackPushScriptValue is very simple.
In fact _stackPushScriptValue is a variant of variableSetToScriptValue
that just do not perform the very last memory assignment operation.


Phidias618:
I realised the solution i described above doesn't even work as ".ARG0" cannot be pass to variableSetToScriptValue, which makes this PR even more useful

Issue #1907, opened 2025-12-30
Horizontal + Movement Up and down


goblinheath:
From the 4.2b change log "Added new sprite animation types "Horizontal" and "Horizontal + Movement" for creating sprites which can only face left or right"

**Describe the bug**
When using the new horizontal+movement animation state in adventure mode, pressing just Up defaults to facing right, and pressing down defaults to left. When holding Left or right, and tapping or holding Up/down buttons it retains its direction unless the last button held is Up or down, then defaults back to Up=Right or Down=Left. This sometimes turns your player sprite potently the wrong way.

**To Reproduce**
1. Create Adventure Scene
2. Create sprite and Set Sprite to Horizontal +movement animation 
3. During gameplay press up and down. 

**Expected behavior**
I expected this animation type to retain the direction last pressed. So if I was pressing right, and released the right button, and then press the up or down button, the player should still be facing Right, and vice versa.

I kinda expected this to be similar of the beltscroller plugin from Shin, which was built upon editing the adventure engine and retains right and left directions. https://github.com/shin-gamedev/gbs-belt-scroller

**Platform:**
 - OS: MacOS Sonoma14.4
 - App version GBStudio 4.2b

PR #1909, opened 2025-12-31
Fix two-way directional animation for adventure type scene

Mico27:
Fixes chrismaltby/gb-studio#1907


chrismaltby:
Thanks @Mico27 I ended up adapting this into 5559418a692373e482dc42e887407113a5bdcc8a which is now merged into develop :-)

PR #1967, opened 2026-02-24
Add pullable actor system, collision checks for push/pull

Dan512:
**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines

**What kind of change does this PR introduce?**

Feature

**What is the current behavior?**

Solid actors in Adventure scenes have no built-in push or pull mechanics. The only push that exists is a push animation if a player runs into an actor or wall. Players cannot interact with solid actors to move them around the scene, natively.

Solid actors can now be marked as "Is Pushable" and/or "Is Pullable" via new collision flags. Each has a corresponding feature toggle in the Adventure engine settings.

**Push:** Player walks into a pushable actor and enters a Push state after a configurable delay. The actor's "On Pushed" script fires each frame, passing direction and hold frame count. A "Push Grace Frames" setting prevents animation flickering during movement. A collision check prevents the script from firing if the actor is blocked.

**Pull:** Player presses interact while facing a pullable actor, then walks away to pull. The actor's "On Pulled" script fires each frame, passing direction and hold frame count. Collision checks verify both the actor and player have room to move. Recommended to use threads in the pull script to move the player and actor simultaneously.

New conditional events "If Pushed Direction" and "If Pulled Direction" allow scripts to branch based on direction and an optional hold frame threshold.

Depends on engine-side changes in Dan512/gbvm#281

**Does this PR introduce a breaking change?**

No. Both features are disabled by default and require opting in via engine settings. Existing projects are unaffected.

**Other information:**

Engine changes (adventure.c, gbs_types.h, engine.json) are in a separate PR on the gbvm repo. This PR covers the compiler, editor UI, events, and localization.


chrismaltby:
Very nice, I've had a little play around with this and it seems to work nicely.

On the implementation side I think I'll need to rework this a little as I don't like how the React components and files like compileData.ts have needed specific workarounds to handle the feature. Ideally I'd like to be more generically defined in engine.json so that the app itself doesn't really know anything about pushable and pullable things, and future scene plugins could use this too.

I'll need to do a full review but from a quick look
- The logic for handling "Pushable/pullable actors must not have a collision group set" etc I'll probably make configurable in engine.json in `extraActorCollisionFlags`
- Not sure how I'd handle it yet but the bit you've got for defining additional script tabs per actor depending on flags set I'd also try to move into engine.json
- I do slightly regret defining "solid" and "platform" (which you extended by adding push/pullable to resources/types) rather than just setting flags 1/2/3/4, I think I had them left over from an earlier more hard coded implementation

Might take me a while to review fully, want to work on getting another small bug fix release out first. Once I'm done though I can start adapting this to work more generically. Thanks @Dan512 :-)

Dan512:
Thanks! Yeah I wasn't sure the best/preferred way to implement it in the GUI etc., but I think the "IF pushed/pulled for ____ frames" is really solid and solves a lot of issues and opens a lot of gameplay options (although "for ____ seconds" would also be a good alternative option, like the Wait function).

PR #1973, opened 2026-02-26
Add font selection input to events and adds a set font event

Mico27:
This adds font selection input for custom events and adds a "set font" event that makes uses of the VM_SET_FONT gbvm operation. 
( This PR was made is mainly because I want to be able to use the font selection input for my plugin )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add support for creating Notes that can be positioned and resized in the World view with multiple color options
+- Add "Set Font" event to change the current default font from scripts [@Mico27](https://github.com/Mico27)
 
 ## [4.2.2] - 2026-02-24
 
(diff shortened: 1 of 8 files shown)


chrismaltby:
Thanks @Mico27 

Made a few small changes so the generated GBVM will use symbols which are in game_globals.i to make it slightly more readable:
```
        ; Set Font
        VM_SET_FONT             FONT_GBS_VARIABLE_WIDTH
```

and I updated setFont to take in an id string rather than looking up the index and passing it in.

The idea is that 
- *ScriptBuilderBase* functions should deal with low level things generating the raw GBVM and mostly taking symbols/indexes/constant values as inputs. Functions here are subject to change a fair bit as the engine is modified.
- *ScriptBuilder*  functions should be more high level, dealing with IDE constructs like the string ids stored in gbsres files. They should provide a reasonably stable wrapper around the internal functions from *ScriptBuilderBase* (which should mostly be prefixed with `_`). As the internals of the engine change I try to keep these up to date to call the right internal functions to keep logic consistent between GBS versions.
- *compile* function in events, ideally only use functions from *ScriptBuilder* and don't get involved with the internals. That way they shouldn't need updating if the GBVM operations change. I know this isn't the case for many plugins that need to get involved with internals but for the inbuilt events at least I prefer to use the "public" functions from ScriptBuilder instead (though there's a few of the more complex events break this rule like "eventReplaceTileXYSequence" I'd prefer to avoid internal functions in events where possible)

^ I should probably document this somewhere :-)

Thanks,
Chris

PR #1974, opened 2026-02-27
Add feature allowing to modify the font table via the font json

Mico27:
Add feature allowing to modify the font table via the font json. 
Comparatively to the mapping field which replace the actual content of the text, this new section of the json allows for changing the mapping of the table itself. This also fixes issues with variables not rendering correctly when trying to map into a different font arrangement using the mapping field and doesnt scrambles/replace the text in the GBVM.
Example json:
image
font file:
image
working preview that match the table:
image
image

Variables correctly displaying:
image

And this is an example before by trying to use the text replacement via mapping:
image
image
image
image


chrismaltby:
I've made small change so we don't end up storing the derived mapping & table values in `.gbsres` :-)

One thing I want to check though 
```
//modify the table with the tableMapping
    Object.entries(tableMapping).forEach(([key, value]) => {
        const tableIndex = key.charCodeAt(0) % 256; //get ascii value of mapped char
        table[tableIndex] = value; //assign mapped value to table
});
```
because the tableMapping is a unicode character in the JSON and then you modulo 256 it means there's some potentially confusing cases if characters outside of ascii range are used as input

e.g a font json
```
{
  "name": "Test Mono",
  "mapping": {},
  "table": {
    "あ": 63
  }
}
```
because
```
("あ".charCodeAt(0) % 256) === ("B".charCodeAt(0) % 256)
```
Will cause "B" to replaced too, so:
Screenshot 2026-02-27 at 14 17 58

Shows up as:
 
Screenshot 2026-02-27 at 14 28 50

Maybe this isn't a problem, but feels potentially confusing to me. What do you think?

Mico27:
mmh Yeah, I could change the code to ignore charcodes that are greater than 256 instead of doing a modulo.

Mico27:
Here I made modifications so that instead of doing a modulo I'll just filter out the character code that arent within the 256 range.
So effectively, 
The mapping field maps (Char/string) -> (Ascii code/Array of Ascii codes)
The table field maps (Ascii code) -> (Tile index)

chrismaltby:
Thanks @Mico27 , I'll do another review pass but that's probably good now :-) Might not get chance to test until Monday

mxashlynn:
So if I understand correctly, with this feature we are unable to use character values greater than 255 when entering text in the IDE.  Is that correct?

If so, is it possible to keep using the old text-replacement-based system with GBS 4.3.0?

Mico27:
The old text replacement system is still there in 4.3.0, they are two separate functionality.
the old text replacement literally replace a character for another, while this new feature allows to map a character to a tile index in the font tileset.
This allows you to create a custom font tileset that dont have to match the default pattern.

mxashlynn:
Very cool!  It sounds great but my current project relies heavily on emoji entered in the text boxes in the IDE so it sounds like I may not be able to use this system at this time.  Thanks for your contributions all the same!

Mico27:
I mean, its used like this:
the "mapping" field is what you use to map non-ascii character that are out of the 255 range into an ascii character in the 255 range(which is how it always worked, in the engine each character is a byte).
So in your case, if your emoji as character out of that 255 range, you use the "mapping" field to map it to a valid ascii character which the IDE will replace every instance of the emoji into the character on compile.
the new feature which adds the "table" field is to specify which ascii character will be mapped to which tile index in the font, by default if that table isnt filled, it'll use the old/default behavior which is ascii value = font tile index + 32 (index of the space character) then go up from there.
So in your case, you'd map the emoji to some ascii character "Ç" (128) for example, then you put that emoji at the (128 - 32)'s tile index in the font. With the "table" field, you could change that to not be dependant on "Ç" being the 96th tile in the font and be anything else you want, thus being able to create your own font structure.

tldr: "mapping" field works just like before, "table" field just allows you more control on the font structure.

mxashlynn:
Thank you for the detailed explanation!  I'm excited to use it.

PR #1987, opened 2026-03-06
Added Ukrainian Language Support

AmakerGame:
**Added Ukrainian Language Support**
This update adds a full Ukrainian (uk) localization to GB Studio.
All interface elements have been translated into Ukrainian, including menus, toolbars, editors, scripting events, settings, warnings, and dialogs — covering 900+ strings in total.


chrismaltby:
That's great :-) thanks for the hard work!  ❤️ 🇺🇦

PR #2073, opened 2026-05-22
Fix: Scene palette overrides with incorrect defaults

pau-tomas:

Fix

When overriding a scene palette, if, for example, palette 4 is the overrode with "new-palette-id" the `paletteIds` field in the scene `.gbsres` will be just ["new-palette-id"]. This means that the next time the project is loaded the replaced palette will be palette 0 and not palette 4.

image

`paletteIds` (and `spritePaletteIds`) are always stored as an 8 elements array, using empty strings for the palettes that aren't overrode.

For the same example as above `paletteIds` will be:
```
[
    "",
    "",
    "",
    "",
    "99af0c5b-2f2d-4c6d-8883-795ad9cb46fa",
    "",
    "",
    ""
  ]
```

I don't think so. 

* **Other information**:

Seems like the error is happening when the scene state is serialized to write to the gbsres file (maybe due to the `pruneMissingEntities`?). There might be a more elegant solution, but I'm not familiar with that part of the codebase.

### DIFF
--- a/src/components/editors/SceneEditor.tsx
+++ b/src/components/editors/SceneEditor.tsx
@@ -658,15 +658,19 @@ export const SceneEditor = ({ id }: SceneEditorProps) => {
     (background?.autoTileFlipOverride !== undefined ||
       autoTileFlipOverrideOpen);
   const onEditPaletteId = (index: number) => (paletteId: string) => {
-    const paletteIds = scene.paletteIds ? [...scene.paletteIds] : [];
+    const paletteIds = Array.from(
+      { length: 8 },
+      (_, i) => scene.paletteIds?.[i] ?? "",
+    );
     paletteIds[index] = paletteId;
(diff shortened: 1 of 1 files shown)


chrismaltby:
Ah, thanks for spotting this @pau-tomas the `pruneMissingEntities` was added recently to fix some issues seen with the compileSprite function when the sprite could have references to deleted spriteStates (though still never managed to track down how this could happen) which cause undefined data in the array of spriteStates after being denomalised. So to help with this I added a pass of filtering out undefined values after denormalising which then caused this issue instead.

I've used this PR as I think it's better to have empty strings and a full array for this anyway but I've also fixed in the `pruneMissingEntities` function too so it actually takes the data schema into account when deciding if values should be filtered out or not. Was worried if I didn't fix there too that a similar issue might occur in another part of the app

PR #2076, opened 2026-05-24
Feature: Super GB palette management and events

pau-tomas:

Feature update

Super GB mode has only one default palette. Other palettes can be set with the "Set Background Palettes" event since they're shared with the CGB ones.

Screen colorization can be done with `VM_SGB_TRANSFER` or [a plugin](https://github.com/Y0UR-U5ERNAME/gbs-plugin-collection/tree/main/plugins/transferSGBPacketPlugin/events).

Got a bit carried away when trying to build a couple of new plugins for the above and ended up implementing all of them (+ some other stuff) into GBS.

**Allow to set four default palettes from Settings**

image

SGB palettes act different than the CGB ones. They all share the same color 0 and they're always set in pairs (it could be any set of pairs but our engine is limited to sending palette 0-1 and palette 2-3 together). The first palette of the pair has 4 colors and the second only 3 (color 0 gets dropped). Because of that, palette 2 is the one that sets the global color 0 (as it's set after the 0-1 pair). 

This could led to confusing behavior like:
image

I tried to explain that with a Note in settings and also by making all the other palettes to only show with 3 colors. Not the most elegant solution, a possible alternative would be to modify our engine to always send color 0 for the first palette when transfering palettes to SGB, this will limit the flexibility but also reduce the possible confusion.  

The new "Set Super GB Palettes" event also has messages that try to explain the behaviour. Here it can get a bit more confusing because setting the second palettes of the pair (Palette 2 and 4) will send the whole pair, so the global color 0 might be overrode too even if Palette 1 or 3 are set to "Don't modify". In this case Palette 1 is represented with 4 colors because it can set color 0 when set. 

image

I also added the palette selector for SGB at Scene level (only if Super GB Mode is enabled and Color Mode isn't)

image

**Set Super GB Color Area event**

This event uses the `VM_SGB_TRANSFER` command to set rectangle areas using any of the 4 Super GB palettes. It allows to set the rectangle area and the palette to fill the inside, the border and the outside of the area. (more details about this transfer command [here](https://gbdev.io/pandocs/SGB_Command_Attribute.html#sgb-command-04--attr_blk))

It includes a scene helper to display the different areas (as a reference I used the red, green, blue and yellow colors used in the [Super GB interface](https://www.spriters-resource.com/snes/supergameboy/asset/60608/)).

image

No

* **Other information**:

- A nice addition would be to be able to preview the game with or without SGB. Right now if color mode is on you can't test without SGB. This will require to modify binjgb and expose a flag to force non-color mode.

- I found an issue with binjgb when rendering border under some circumstances. tl;dr if both fill and outside are selected but not border the border palette shouldn't be set. Sameboy (on the right) does the correct thing. 
image [Update: PR with the fix [here](https://github.com/binji/binjgb/pull/79)]


chrismaltby:
Hi @pau-tomas nice one! Just reviewing this now

Where you've added the palettes to the monochrome+sgb sidebar



does that actually work for you? I've tried setting colors there and it always defaults back to the palettes defined in settings

Screenshot 2026-05-26 at 16 22 22

I think it's fine if the palettes need to be configured using this "Set Super GB Palettes" event but we can just remove that section if it's not working.

and then on this line in the settings
"FIELD_UI_SGB_PALETTE_NOTE": "Note: Super GB palette #4 is also used for UI elements like dialogue windows and menus.",


This is specifically the Window palette for GBC games isn't it, Super GB doesn't use it for dialogue/menus does it?

Maybe that should be something like 

"FIELD_UI_SGB_PALETTE_NOTE": "Note: Super GB palette #4 is also used for UI elements like dialogue windows and menus on GB Color.",

and this warning could be hidden for Monochrome + SGB games

pau-tomas:
> does that actually work for you? I've tried setting colors there and it always defaults back to the palettes defined in settings

Could it be because #2073? The SGB palettes are updating the last 4 palettes of CGB and without that fix they'll always be writing to the first 4 instead.

pau-tomas:
> This is specifically the Window palette for GBC games isn't it, Super GB doesn't use it for dialogue/menus does it?


That's right!

chrismaltby:
> > does that actually work for you? I've tried setting colors there and it always defaults back to the palettes defined in settings
> 
> Could it be because #2073? The SGB palettes are updating the last 4 palettes of CGB and without that fix they'll always be writing to the first 4 instead.

ah, sorry yes, you're right! It was me that broke it 😅 I've merged develop back in and that's working perfectly now!

PR #2083, opened 2026-05-29
Update to latest binjgb emulator

pau-tomas:

Update

We are using a custom fork of binjgb that wasn't updated for almost 2 years 

We the latest binjgb commit (up to https://github.com/binji/binjgb/commit/df28a7d74d3fba725b58437b1ab06e6128380547). We also don't use a fork anymore as all the methods we needed to be exposed are now available. 

I hope not.

I tested the music tracker, music and sound FX preview from events and the preview emulator and debugger and all seem to work fine.

* **Other information**:

Now that we use upstream binjgb should we have it as a submodule for the project like https://github.com/gbdev/rgbds-live does?


chrismaltby:
Nice! Works great from what I can tell.

On the topic of setting it as a submodule, that would be a good idea but it would mean adding a dependency on having `emscripten` installed and setup (don't know how awkward that is on Windows either) for anyone who wants to contribute and/or run from source.

The other option would be to add it to https://github.com/chrismaltby/gb-studio/blob/develop/src/scripts/fetchDependencies.ts if we have a URL to the latest release zip. This is already set up to cause the CI server to fail if the file URL ever changes unexpectedly (it generates a `dependencies.lock` file with expected hashes) so it should be safe to use for fetching binjgb too

PR #2099, opened 2026-06-12
Security: Potential prototype pollution in sortByKey helper

tuanaiseo:
## Problem

The sortByKey function uses Object.keys() and reduces into an object, but doesn't protect against prototype pollution via __proto__, constructor, or prototype keys. If user-controlled data is passed to this function, it could pollute the prototype chain.

**Severity**: `medium`
**File**: `src/shared/lib/helpers/sortByKey.ts`

## Solution

Use Object.create(null) as the initial reduce value, or explicitly check and reject dangerous keys like '__proto__', 'constructor', and 'prototype'.

## Changes

- `src/shared/lib/helpers/sortByKey.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/shared/lib/helpers/sortByKey.ts
+++ b/src/shared/lib/helpers/sortByKey.ts
@@ -6,6 +6,6 @@ export const sortByKey = <T extends Record<string, unknown>>(input: T): T => {
         obj[key] = input[key];
         return obj;
       },
-      {} as Record<string, unknown>,
+      Object.create(null) as Record<string, unknown>,
     ) as T;
 };


chrismaltby:
Thanks, I checked and only one place in the application even used this helper, so instead I've inlined the logic to that location in a way that doesn't allow prototype pollution, added a test to confirm the logic still works and deleted the helper file in 33e82566006816ff7ebcd76a5c4600634f0c1bd0

PR #2108, opened 2026-06-20
Add tooltip to ConstantSelect dropdown options

FlyinWhee:
Constants can have long names that get truncated with no way of seeing the full name even after selecting it. The matter is made worse when using folders (with "/" separators).

Adds a `formatOptionLabel` render to `ConstantSelect` that wraps each option in a `<span>` with a `title` attribute, showing the full constant name on mouse over.

image

### DIFF
--- a/src/components/forms/ConstantSelect.tsx
+++ b/src/components/forms/ConstantSelect.tsx
@@ -271,6 +271,9 @@ export const ConstantSelect: FC<ConstantSelectProps> = ({
               onChange(newValue.value);
             }
           }}
+          formatOptionLabel={(option: Option) => (
+            <span title={option.label}>{option.label}</span>
+          )}
           {...selectProps}
         />
       )}
--- a/src/components/ui/scripting/style.ts
+++ b/src/components/ui/scripting/style.ts
@@ -575,8 +575,8 @@ export const StyledScriptEventBranchHeader = styled.div<StyledScriptEventBranchH
       margin-right: 10px;
     }
     ${StyledFormFieldInput} {
-      width: 100%;
-      max-width: 200px;
+      flex: 1;
+      min-width: 0;
 
       .CustomSelect__control {
         height: 22px;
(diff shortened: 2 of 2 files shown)


maxoakland:
This is good. Is it possible to make the menu wider? That way it would be easier to read from the start. And there's a *lot* of space to the right that has nothing in it

FlyinWhee:
> This is good. Is it possible to make the menu wider? That way it would be easier to read from the start. And there's a _lot_ of space to the right that has nothing in it

Yeah, that was my intent, but couldn't find a way that didn't require changing a lot of stuff. I'm not much of a frontend dev, so there might be a way... I just didn't find one.

Might give it another look later!

maxoakland:
> > This is good. Is it possible to make the menu wider? That way it would be easier to read from the start. And there's a _lot_ of space to the right that has nothing in it
> 
> Yeah, that was my intent, but couldn't find a way that didn't require changing a lot of stuff. I'm not much of a frontend dev, so there might be a way... I just didn't find one.
> 
> Might give it another look later!

That makes sense! I wish I had more skills in this area

FlyinWhee:
I have commited new changes that remove the forced max-width on the "when" selectors and replace them with "flex" to allow using the whole area. Both changes together should alleviate most usability problems from that component.

Current state :
switch_when_current_state

Suggested change :
switch_when_suggested_change

chrismaltby:
Hi @FlyinWhee thanks for pointing out this issue, unfortunately this fix isn't quite right but I've made some changes to improve the UX around the problem you were facing.

I was sure the menus used to be wider so I scanned through old versions of the repo with `git bisect` and found the exact commit where menus were becoming so small that this menu was cropping and used that to make a version where menus will now have the width set to at least fit the content without cropping. 

The "when" dropdowns bit was almost correct but added some additional CSS that wasn't needed and missed that there was a `<FlexGrow />` div in `ScriptEventBranchHeader` that needed removing to allow those inputs to fill the width available

With these changes the select inputs and menu will now look like this

Screenshot 2026-06-21 at 10 27 45

FlyinWhee:
Thanks for the explanation and taking the time to fix it!

Issue #1551, opened 2024-09-02
When using arrow keys to navigate Sprite frames, scroll to keep active frame in view


lkcampbell:
**Describe the bug**
When you have a large enough number of Sprite frames in the bottom Frames view, the view provides a horizontal scroll bar and the excess frames are moved off the left or right side of the screen. One way to navigate these frames quickly are the left and right arrow keys. When you navigate to a frame that is outside of the frames view, the view does not scroll to show the current frame.

**To Reproduce**
1. Make an animation in Sprites that has enough frames that the horizontal scroll appears, say, 10-20.
2. Select a frame as your current frame and use the arrow keys to navigate to a frame outside the left and/or right edge of the horizontal scrolling view.

**Observed behavior**

Before hitting left arrow:
![Screenshot 2024-09-02 at 3 39 47 AM](https://github.com/user-attachments/assets/54519b22-a0e9-478c-abb3-4f1899cd0354)

After hitting left arrow:
![Screenshot 2024-09-02 at 3 40 04 AM](https://github.com/user-attachments/assets/77426171-0588-4b09-86fd-dbe216af2320)

**Expected behavior**

Before hitting left arrow:
![Screenshot 2024-09-02 at 3 39 47 AM](https://github.com/user-attachments/assets/54519b22-a0e9-478c-abb3-4f1899cd0354)

After hitting left arrow:
![Screenshot 2024-09-02 at 3 51 16 AM](https://github.com/user-attachments/assets/6e3975d8-501a-4eb9-83f4-a28bf15a995e)

**Platform (please complete the following information):**
 - OS: macOS 14.6.1 (23G93)
 - App version 4.0.2

lkcampbell:
Confirmed that this is still a bug as of version 4.3.2.

lkcampbell:
FYI, I am going to try to fix this bug so I can learn a bit of the code base. Relatively speaking it seems like a good first issue.

PR #2119, opened 2026-07-10
Fix issue where active sprite frame cannot be seen after arrow navigation

lkcampbell:
- [X] Tests for the changes have been added (for bug fixes / features)
I don't think there are any tests readily available for React components although I did run my change in Storybook and it appeared to work correctly.
- [X] Docs have been added / updated (for bug fixes / features)
No document changes needed.

It is my first PR, so just let me know if I made errors on filling out the requirements.

Bug fix for issue #1551

See issue #1551

The SortableList component now properly scrolls to show its active item after using the keyboard to navigate to a item that is off of the edge of the layout. I also put a small amount of padding into the scroll-margin style so the edge of the item isn't flush with the edge of the layout after the scroll happens. It looks better that way.

No.

* **Other information**:
Since the SortableList component is used both in the Sprite Animation Frames and in the Pattern Order frame in the Music Editor, this fix works in both of those editor frames.

### DIFF
--- a/src/components/ui/lists/SortableList.tsx
+++ b/src/components/ui/lists/SortableList.tsx
@@ -1,4 +1,11 @@
-import React, { JSX, useCallback, useEffect, useRef, useState } from "react";
+import React, {
+  JSX,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from "react";
 import { SortableItem } from "ui/lists/SortableItem";
 import { StyledSortableList } from "./style";
 import throttle from "lodash/throttle";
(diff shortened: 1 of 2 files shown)


chrismaltby:
Hi @lkcampbell nice yeah seems to work great, happy to merge.

> I don't think there are any tests readily available for React components although I did run my change in Storybook and it appeared to work correctly.

There's only a small handful of tests for React components right now, there's an example here https://github.com/chrismaltby/gb-studio/blob/45bdd029b327a2686f50969f7119c7cbdf745ebb/test/components/forms/VariableSelect.test.tsx

but for the moment I've been putting most of the testing on
- anything related to data migration
- shared, mostly pure function, helpers
- the redux store (though I've still got a way to go with that)

You can see the code coverage here if you haven't seen already
https://chrismaltby.github.io/gb-studio/coverage/

Thanks for contributing!

PR #2123, opened 2026-07-13
[dev] Fix react hooks eslint dependency

pau-tomas:

Fix

With the new dependency updates, the react hooks eslint stopped working. So, for example, linter wasn't wanrning if the useEffect dependencies were wrong.

Updated the dependency to make it work.

No

* **Other information**:


chrismaltby:
Ah nice, didn't spot that! I'm going through the list of security advisories that Github provides to remove all the warnings, still got a way to go and some of the updated modules have breaking changes,`glob` looks like it might be a bit of a problem, as it handles paths differently on Windows now. Looks like I might have broke a few things on Windows already actually with some of these updates so I'm getting a dev environment set up to figure out what's going on.

Might be a bit painful for a bit but it will be nice to be on up to date modules again :-)

PR #2145, opened 2026-07-23
Add deduping identical compiled entity scripts feature

Mico27:
This feature adds the option in build settings (enabled by default) to dedupe identical entity scripts to reduce redundant ROM data and save on ROM space. Ideal for people who use actor/trigger prefabs with large scripts and use those prefabs across multiple scenes.

### DIFF
--- a/src/components/settings/section/SettingsSectionBuild.tsx
+++ b/src/components/settings/section/SettingsSectionBuild.tsx
@@ -45,6 +45,9 @@ export const SettingsSectionBuild = ({
   const showRomUsageAfterBuild = useAppSelector(
     (state) => state.project.present.settings.showRomUsageAfterBuild,
   );
+  const dedupeScriptsEnabled = useAppSelector(
+    (state) => state.project.present.settings.dedupeScriptsEnabled,
+  );
   const compilerPreset = useAppSelector(
     (state) => state.project.present.settings.compilerPreset,
   );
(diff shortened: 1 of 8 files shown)


chrismaltby:
Hi @Mico27 nice! Is there a reason this has been made a user configurable setting though? I think this probably should always happen, similar to how `gbvmScriptChecksum` is used to dedupe custom scripts automatically already.

Mico27:
@chrismaltby I wasnt sure we always would want to dedupe scripts in case someone would like to debug and find the corresponding compiled script of an entity more easily when exporting source, but if its a non-issue and people can just trace it by looking at the compiled entity to find the related script we can make it consistent without the option.

chrismaltby:
Hi @Mico27 I've extracted the dedupe functionality from this PR and applied it in d1eb1900937720d81d0dd211aa5c2a028a69048a
This now will dedupe automatically with no user setting needed

I had to make it so that deduping wouldn't happen when the debugger was enabled though otherwise the deduped script was always the one being displayed in the debugger, so it would look like the wrong entity triggered a script to run and if you made changes to the script it wouldn't apply to the one that was currently running, it would instead edit the copy that all the other entities were sharing.

If you "Export ROM", "Export for Web" etc I've made it always do the dedupe in 112215cbefa5f19f4e6a585ddce220be934f16be so the final build will be optimised

While looking at this I found an issue where the debugger wouldn't ever show prefab scripts, this might have been an issue even since prefabs were added and I've fixed that here 239b9d948cd82b877168197a7d5dd396cbe470bd

Thanks

PR #2146, opened 2026-07-23
Global variable rework + Array implementation

Mico27:
Variables reworked
Previously variables were 512 anonymous slots. A variable existed only once named, and a variable only compiled if a script referenced it.

- Explicit add/delete: created via "Add Variable" (pane + menu, folder context menu, or inline "+ Add Variable" in any variable dropdown). Clearing a name no longer deletes. Deletion is an explicit action.
- Folders: group variables with drag-and-drop reparenting, folder rename, and delete, exactly like constants.
- Reordering: Move Up/Down among siblings. Folder members stay contiguous.
- Always compiled, stable indices: Every defined variable compiles even when unreferenced. Runtime index = navigator position (shown next to each variable) and is stable across builds.
- Load migration: Old projects auto-create entities for every script-referenced variable, so nothing changes behavior.
- Defined-first dropdowns: Selects list defined variables grouped by folder. Legacy undefined references still render.
- Folder inspector: Working folder rename (typing a path moves it) plus an Array Uses pane.

Arrays implementation
An array = a folder of variables with contiguous runtime indices, compiled via GBVM indirection (.R_REF_IND / VM_SET_INDIRECT), base index resolved symbolically from the folder's first member.

- Creation: "Add Array" opens a popup for name + variable count.
- Expression syntax: Arr[0], Arr[$Idx], Arr[$Idx] = 5, $result = Arr[$Idx], nested indexing like Outer[Inner[0] + 1], with live editor validation.
- Value fields: an Array type option (shortcut a) in every value field's type menu, with a nested value as the index. Array reads also work in value-field expression mode. Constant indexes fold at compile time.
- Events: Array Set Value (array/index/value) and Array Get Value (array/index/destination), built on a new variableArray field type that plugins can use too.
- Tooling coherence: uses tracking sees variables inside indexes and arrays everywhere they appear. renaming/moving a folder rewrites all script references (event dropdowns, value nodes, expression text with formatting preserved).

Limitations: same-basename folders are ambiguous in expressions (lowest index wins). no runtime bounds checking.


chrismaltby:
Hi @Mico27 very interesting, I'm not sure I can use this as-is but this is a feature I had a few ideas about how I'd like to approach myself. I'm going to start looking at that very soon and might be able to pull parts from this PR in the implementation :-)

Mico27:
@chrismaltby Whatever solution you're planning to take, one of the thing I'd really like to be retained from this would be the "creating variables in bulk" and "always have them compiled even if not referenced in script" and "reordering of variables" part of this rework. One of the issues with the way variable are currently managed makes it very difficult to refer to variable by their index to use with the indirection functions, having to manually create each variables one by one and add a reference for each of them in a custom gbvm script to force them to be compiled is tedious, not to mention reordering them is not possible in the IDE (which currently you have to edit the variable asset file manually).

chrismaltby:
Just to update, I've been working on taking an old branch I had for arrays and getting it into a better place. Still lots of work to do and lots of tests but I've got a version where 

similar to yours Variables are now created rather than having a preset list. I've also taken this opportunity to make variables use UUIDs rather than incrementing integer ids which was a common complaint making it difficult to work in teams without getting merge conflicts.

Screenshot 2026-08-01 at 16 28 14

When adding a variable the inspector on the right allows you to choose the type of the variable, setting array then lets you set the size which will reserve N consecutive addresses for each array element at compile time

Screenshot 2026-08-01 at 16 26 50

Then pretty much everywhere that allowed variables to be chosen will, when an array variable is selected, allow setting the index as both the input and output value, no new events have been added, array support just works in the existing events instead

Screenshot 2026-08-01 at 16 34 42

This also works in scripts, though you need to set the parameter type explicitly as an `Array Ref` for the index field to be displayed
Screenshot 2026-08-01 at 16 37 32

In the debugger variables panel you get an entry for every array item 
Screenshot 2026-08-01 at 16 28 42

Still a bit to go but I think this will be a good base to work from adding any improvements needed later

chrismaltby:
@Mico27 I've merged the array + uuid variables branch into develop now and you can find a dev build at https://github.com/chrismaltby/gb-studio/actions/runs/31000858525

This required a migration to make sure any referenced globals had entries in the variables list and given this is going to be quite a breaking change with the UUID variables preventing easy migration back to previous versions I've set the new file version as 5.0.0

There's one known breaking change when migrating from a previous build, I've made it so the link cable transfer event requires the variables to be arrays if the packet size is greater than 1 and the array must be large enough to fit the packet data. I don't think it's possible to auto migrate that case as depending on how those variables are used in the rest of the application it could break things to automatically replace them. It should be a quick manual change to anyone who did use link cable transfer events for transferring multiple variables at once though.

PR #2155, opened 2026-07-24
Add support for testing projects with external emulators

8bitKittyKat:
A new setting in preferences allows the play button to launch a user specified emulator, the OS default handler, or the default built-in emulator. Debugging is only available with binjgb. This is an opt-in setting, so the default is binjgb unless set otherwise.

This currently will only work with emulators that don't need arguments passed for launching.


8bitKittyKat:
@pau-tomas  Since we were discussing this yesterday on discord, when you have a chance can you review this PR and give any feedback you might have?

pau-tomas:
Nice! I'm away from my computer for the weekend, but will take a look on Monday 🙏

maxoakland:
This is so welcome. I've had to work around this lately by opening the temporary ROM in Sameboy

8bitKittyKat:
> This is so welcome. I've had to work around this lately by opening the temporary ROM in Sameboy

A lot of my own work in GB Studio involves SGB features that binjgb doesn't support so I've needed this feature myself for a while..

8bitKittyKat:
I forgot to say and should have pointed out that I've only tested on Windows. I don't have an environment setup currently for testing Linux or Mac OS.

chrismaltby:
Nice @8bitKittyKat 

Testing on a Mac though this breaks if you pick a specific application to use

Screenshot 2026-07-27 at 14 32 41

Looks like it's the spawn causes a permissions error
```
          spawn(emulatorPath, [romPath], { detached: true });
```

I replaced it with a call to `open` instead matching how we do the same for opening images and MOD files here matching how it works in https://github.com/chrismaltby/gb-studio/blob/3de3771dc5f02b08de4e4bdd0514d0db43695024/src/apps/gb-studio/main.ts#L996-L1020

and it worked for me on Mac
```
          const app = emulatorPath;
          open(romPath, { app });
```
If that still works for you on Windows I'd switching to using `open` instead.

The main thing now is what to do about the debugger when an external app is set.

Currently it's actually still appearing but with a broken visual layout (there's a vertical line appearing on the left column, I assume from the VRAM preview that would normally be there) and obviously the data doesn't work

Screenshot 2026-07-27 at 14 38 08

I'm wondering if it should just be blank and show the message you added to en.json
```The debugger will only work if you are using the built-in emulator.```

it would be nicer though if clicking the "Start Debugger" would just use the inbuilt emulator for that one session. That way you always just using the debugger when you need to while still keeping the default to run using the selected emulator. What do you think?

pau-tomas:
> it would be nicer though if clicking the "Start Debugger" would just use the inbuilt emulator for that one session.

I like that. Maybe changing the text to "Start debugger with built-in emulator" or adding a note below the button to be clear what's happening.

8bitKittyKat:
Switching from `spawn` to `open` worked for me in Windows.  Also, in the Game menu, `Run with Debugging` does launch binjgb. The behavior I'm seeing with `Start Debugger` already has it open the built-in emulator too, but maybe it doesn't call the right emulator every time? I'll take a closer look.

8bitKittyKat:
Alright, the external emulator should now launch properly on Mac OS. The debugger pane won't be active with an external emulator now either, but `Start Debugger` will launch and activate binjgb.

8bitKittyKat:
> > it would be nicer though if clicking the "Start Debugger" would just use the inbuilt emulator for that one session.
> 
> I like that. Maybe changing the text to "Start debugger with built-in emulator" or adding a note below the button to be clear what's happening.

It could be clearer if it read as the debugger starting with the built-in emulator, but I'll wait for @chrismaltby to give input on that before proceeding with a change there, since that's an existing string and would affect all the existing translations too. A note under the button may be good enough too.

8bitKittyKat:
I just made a small change to account for if someone was running the debugger and then launched the external emulator. Previously it would have disconnected the debugger but now it won't touch it if it's running already.

chrismaltby:
Tidied up a few things but this is all good now, merged! Thanks @8bitKittyKat

nielsderijk:
Very nice, another long-standing wish fulfilled! 

Is there a way to force the emulator to bypass the boot screen? I am using SameBoy for testing, but unfortunately, there isn't an option to toggle it off in-app.

chrismaltby:
> Is there a way to force the emulator to bypass the boot screen? I am using SameBoy for testing, but unfortunately, there isn't an option to toggle it off in-app.

I think that's completely controlled by the emulator, as far as I'm aware nothing a ROM could do about it, I would suggest asking the developers of SameBoy as a feature request

PR #2168, opened 2026-08-02
Add Joypad selector to "If Button Held" event for Super GB Mode

pau-tomas:

Feature

Users can't select a joypad on the "If Button Held" event. This option is available in GBVM.

Added a new Joypad selector to the event which is only visible is Super GB mode is on. Also added a setting to pick the max amount of joypads that it also sets the `MAX_JOYPADS` constant in the engine.

image

image

Binjgb doesn't support multiple joypads so it can only be tested with other emulators (like BGB) or with a real device. 

No. I think I covered the cases to default to joypad 0 in all the cases.

* **Other information**:

Toxa mentioned in Discord that it should be ok to just set `MAX_JOYPADS` constant to `4` in the engine. In that case the setting won't be needed and it'd simplify the code a lot. I don't think there's a use case where somebody would like to change that other than to have the event UI less cluttered showing less joypads to pick. @chrismaltby WDYT? The setting can be easily reverted as its a separate commit.


chrismaltby:
Nice :-) Yeah I think it's probably fine to just set MAX_JOYPADS as 4 rather than needing to have an extra option in the settings page. I've pushed a small change to the event schema to make the fields wrap on to a new line when there's no much room but if you want to try with just a set MAX_JOYPADS=4 I think it would be simpler.

What I would say though is it seems odd to only have "If Button Held" supporting multiple joypads, while "Attach Script To Button" and "Pause Script Until Button Pressed" don't support it, I know GBVM doesn't support joypads for those events but maybe it should?

Not sure how we could add it easily though, if this is useful on it's own then maybe it's fine, but it feels potentially confusing to me.

8bitKittyKat:
My understanding is that because of the way the input polling works, when using the multiplayer mode of the SGB, it loops through input for each controller so when that number is set to 4, it's going to loop 4 times. Might not be a big deal but not many games use more than one gamepad, so setting it higher than needed for every SGB enabled game is going to waste cycles in the long run.

chrismaltby:
After some thought I think SGB multi controller is out of scope for now for the core application for now, it should be possible to add any of this through plugins though and if not I'm happy for the plugin API to be extended to allow it

PR #2180, opened 2026-08-06
Added WRAM usage, BANK0 usage and Logging plugin memory usage build options

Mico27:
This add New build options for WRAM and bank 0 usage bars similar to the ROM usage bar, plus an option to generate a plugin memory breakdown report in build/rom/plugin_usage.txt. Also fixes WRAM busts ignoring the 672 bytes of engine-reserved WRAM (shadow OAM, 0xDF00 block, stack).

This will be useful for budgeting bank 0 and WRAM, the scarcest resources in a plugin-heavy projects. It will help troubleshooting if users a busting the bank 0 and/or WRAM limits by using too many plugins/plugins features.
It'll help them deciding which plugin to drop or which plugin features to uncheck when memory runs short.


chrismaltby:
nice @Mico27 I'll start reviewing this one next, related to this I've got a branch that displays memory usage inline at the top of the script editor

Screenshot 2026-08-26 at 14 40 14

I might use this as an opportunity to introduce that change too

chrismaltby:
Continuing to work through this, having to change quite a lot but it's starting to come together. My version adds "ROM Usage" as a new section in the debugger

Screenshot 2026-09-03 at 17 00 35

This also shows when scripts are being reused
Screenshot 2026-09-03 at 17 05 58
and if you click "+6 more"
Screenshot 2026-09-03 at 17 06 04
it expands to show all uses of the script and clicking on any of the entries selects that actor/scene/trigger script in the editor

You can also output this information similar to what you had in plugin_usage.txt using the GB Studio CLI

still lots of edge cases to test but getting there.

chrismaltby:
Hi @Mico27 I've adapted this PR into the version that's now in the `develop` branch.

I had to strip some parts out, especially around how failed builds were handled, as it seemed to be doing things like reading stale .map data from the previous successful build and combining it with data extracted from human readable errors using regexes from `romusage` which would be way too fragile to use. (for one thing wording of error messages changing would break failure reporting in a way that would be difficult to pick up with automated testing). Instead in failure cases I've switched to only parsing the available object files to determine memory usage which shows a reduced report but still highlights when overflows were happening

Screenshot 2026-09-05 at 14 29 58

it's worth testing this version with any of the test projects you were using for this PR to make sure it's still catching all the issues you needed it to. If there's anything I've removed that was ONLY available with that error message regex method I'd suggest it maybe needs, `romusage` to be updated to expose the data directly in the JSON output rather than doing anything hacky with the error messages.

Think people will find this feature very useful, thanks for putting this together :-)

Mico27:
Yeah, romusage would need to be modified to expose the overflows if we want to display it. I did the json parsing on the output message as a workaround but I understand thats too hacky of way to do it.

Thanks for implementing this!

Issue #2181, opened 2026-08-07
Flag names disappeared in latest builds


nielsderijk:
**Describe the bug**
All custom names on Flags have disappears in the latest 3 dev builds.

**To Reproduce**
1. Set flag names in earlier version of GB Studio
2. Clear or add a flag, see how the custom names have disappeared

**Expected behavior**
Flag names are persistent

**Screenshots**
Latest builds:
<img width="923" height="366" alt="Image" src="https://github.com/user-attachments/assets/bd258ebb-6c54-49c6-bf10-815e1a572dc2" />

Before:
<img width="897" height="389" alt="Image" src="https://github.com/user-attachments/assets/9d8fd8ab-648b-4d11-8447-d8f1f3f8769f" />

**Platform (please complete the following information):**
Running the latest CI/CD dev build: https://github.com/chrismaltby/gb-studio/actions/runs/31024349491

chrismaltby:
Ah, oops! I always forget the renaming flags feature exists :-( I’m away on holiday at the moment but I’ll have a look next week when I’m back, should be a simple fix to the migration script

chrismaltby:
This is fixed in the latest dev builds @nielsderijk :-)

PR #2184, opened 2026-08-10
fix: migrate raw variable data before schema cast to preserve flag names

JimScope:

Bug fix

https://github.com/chrismaltby/gb-studio/issues/2181

  Adds a small "raw resource migration" step that runs on raw resource data
  before `Value.Cast`, gated on the project version read from the metadata:
  for pre-5.0.0 projects, variables missing a `type` get `type: "number"`
  (is safe because those projects cannot contain array variables). With `type` present,
  the cast preserves `flags`.

  Applied to both load paths (resource-format projects and legacy single-file
  projects).

 Nope

* **Other information**:


JimScope:
closed fixed in https://github.com/chrismaltby/gb-studio/commit/72124f44c09058dc4f630159e89fa4cc2fee32f1

PR #2190, opened 2026-08-19
Update pl.json 2377 feat: add "len" ...  from [70f4f94]

ReptiIe:
Update for the pl.json.
Added missing lines from the:
- feat: add "len" function to script values and expressions to get the... (70f4f94)

Update

Missing few lines.

Added missing lines.

No.

* **Other information**:
Merge asap (((_^)


chrismaltby:
Thanks @ReptiIe made a small change to this feature today to be more consistent with naming, I'd been switching between using array `size` and `length`, I've settled on using `length` everywhere which means some of the localisations needed editing too but I'd made the change to pl.json in 2a15ec18ca02f2fd6b8b15529b4a3d0949603af5

it also means if anyone has been using this build early their arrays will have reset to length=1 as it's changed in the gbsres data too but I think it was important to make this change to be consistent before an actual release of this build :-)

ReptiIe:
OK, got it. I mean, before making ANYCHANGES, I am comparing ENGLISH in my ENGLISH (en.json)
and then POLISH in my POLISH (pl.json) - so before any PR, I will make sure that PL.JSON (that I am sending) will be the same as EN.JSON. Good to know!

chrismaltby:
I did actually make the changes AFTER you made this PR, so you did everything right. You're just too quick at keeping the localisation up to date 😄

PR #2196, opened 2026-08-21
Add Array Init, Array Shuffle and If Value in Array events

pau-tomas:

Feature

The events don't exist

Added 3 events to manage arrays:

- **Array Init / Array Set Values**: Allows to set the values all the elements of an array in one event
- **Array shuffle**: shuffles the array using the rpn random. Requires a call to the random seed event to ensure proper randomization.
- **If Value In Array**: conditionally runs a script if the value exists as part of the array

No

* **Other information**:

N/A


nielsderijk:
I've been testing Arrays over the past few days, and wow, this opens up so many possibilities. Thanks for adding this!

I do have a request. Would it be possible to have an event that stores the index number of the retrieved value?

Example:
I have an Array of 5, the values are 0,0,9,2,9
Then, for every 9 in this array I want to store the index number of the requested value in a variable (i.e. 2 for the first occurence, 4 for the second).
So basically "If value in array" = true -> "Store iDX value in variable"?

pau-tomas:
Will think about it. For now you could do that with the For each value in array event and a variable that you increase until you find the value in the array. You could also wrap it in a custom script if you're using it in several places.

chrismaltby:
Thanks @pau-tomas I've been working through reviewing this today.

I found some issues with ArraySet, for example if you created an array, then made the event to set the values, then changed the size of the array it would throw an error saying that the sizes didn't match.

Instead I've made it so it will loop the full array length, if a value is provided it will use it otherwise it will set the value to `0` and if the values array is longer than the array you want to set the additional values will be ignored. I've also switched it to use a single RPN call rather than a call to variableSetToScriptValue for each item

I'll work through reviewing the others over the next few days

PR #2197, opened 2026-08-23
Update ja.json localization added for foreach array fields/events

tomo666:
Update ja.json localization added for foreach array fields/events


chrismaltby:
Thanks @tomo666 just to let you know, a few more keys were added today too in https://github.com/chrismaltby/gb-studio/pull/2196 :-)

PR #2213, opened 2026-08-31
Use values for Draw Text event coordinates

pau-tomas:

New feature

The Draw Text event only supports numbers for the coordinates

Draw Text uses a value field for the coordinates

No (if I did the migration correctly)

* **Other information**:

One nice thing you can do now is to write background text at screen coordinates using the camera properties. It'll work much better with the new gbvm version that includes https://github.com/chrismaltby/gbvm/pull/314

Before
Screenshot 2026-08-31 at 23 15 27

After
Screenshot 2026-08-31 at 23 15 50

### DIFF
--- a/src/lib/events/eventTextDraw.js
+++ b/src/lib/events/eventTextDraw.js
@@ -36,21 +36,27 @@ const fields = [
     key: `x`,
     label: l10n("FIELD_X"),
     description: l10n("FIELD_X_DESC"),
-    type: "number",
+    type: "value",
     min: 0,
     max: 19,
+    defaultValue: {
+      type: "number",
+      value: 1,
+    },
     width: "50%",
-    defaultValue: 1,
   },
   {
     key: `y`,
     label: l10n("FIELD_Y"),
     description: l10n("FIELD_Y_DESC"),
-    type: "number",
+    type: "value",
     min: 0,
     max: 17,
+    defaultValue: {
+      type: "number",
+      value: 1,
+    },
     width: "50%",
-    defaultValue: 1,
   },
   {
     key: `location`,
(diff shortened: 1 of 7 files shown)


src/lib/events/eventTextDraw.js
```diff
@@ -67,8 +73,20 @@ const fields = [
 ];
 
 const compile = (input, helpers) => {
-  const { textDraw } = helpers;
-  textDraw(input.text, input.x, input.y, input.location);
+  const { textDraw, textDrawScriptValue, getConstantValue } = helpers;
+
+  if (input.x.type === "number" && input.x.type === "number") {
```
pau-tomas, line 78: I considered moving this optimization logic to inside the `textDrawScriptValue` function but it felt a bit more clear here since that function is already very large. But I'm not 100% sure what's the best approach.

chrismaltby: Thanks @pau-tomas I'm going to have a look at this, I think probably I'd prefer the optimisation happen inside `textDrawScriptValue` so any other plugins which use it will get the optimisation for free, though I agree the function is getting large. This pattern of having checks for the type of script values for different optimisations paths happens quite a lot in script builder, I wonder if it could be wrapped in some helper functions to simplify, I'll have a think

By the way this was broken anyway, you test `input.x.type` twice for both number and constant paths leading to NaN being using in generated GBVM if you used a number for X and a constant for Y :-)

Leave this with me I'll take this PR and build on it, thanks.

chrismaltby:
All good now :-) made a few changes:
- moved the fast path checks for compile time fixed x/y coordinates to inside `textDrawScriptValue()` 
- moved the shared logic from between `textDraw()` and `textDrawScriptValue()` into `_drawText()`
- fixed a bug in `_loadAndDisplayText` where prependTokens were used multiple times if you added a `wait` instruction in the text causing it to jump to the draw x/y position again after the wait
- made it so `precompileScriptValue` can optionally `resolveConstants` to turn constants into their number values so only one fast path check is needed. Now just need to check both values are numbers rather than also checking if they were constants
- not part of this but while looking at this also realised a related issue where codegen wasn't optimal if you did simple maths with constants  
e.g. a script value like this  
  Screenshot 2026-09-18 at 17 48 02  
would give generated GBVM as
```
        VM_RPN
            .R_INT16    CONST_MY_CONST
            .R_INT16    6
            .R_OPERATOR .MUL
            .R_INT16    32
            .R_OPERATOR .B_AND
            .R_REF_SET  VAR_S0_LOCAL_0
            .R_STOP
```
but in 3ed68b883b43d1741e4322e28815cfe743e39e90 it will now recognise parts of the script value that can be generated as a compile time expression to instead give

```
        VM_RPN
            .R_INT16    ^/((CONST_MY_CONST * 6) & 32)/
            .R_REF_SET  VAR_S0_LOCAL_0
            .R_STOP
```

